PR #5, opened 2022-12-08
Add dotnet/icu mapping

directhex:
ICU is needed for mobile/browser


premun (approved):

premun:
@directhex can you please verify that there are no directories we'd like to cloak? We don't have binary scanning in place yet so we won't know what goes into the VMR.

directhex:
@premun what kinds of things would we want to cloak?

premun:
Anything that is not a text-based file. So binaries mostly. But also if you know about something that is definitely not needed, or you just need a small bit from the repo, you can alleviate from the total size

directhex:
OK, well, we can mask the `icu/icu4c/source/test` folder, which has a few non-text test files, e.g. fonts. That's a low-hanging fruit.

Some of the ICU source data is in a binary format I don't have a great understanding of. These live in `icu/icu4c/source/data/in` - I doubt the build works without these files.

And finally, the ugly: right now we have a problem with our ICU build where the results are not always deterministic between runs - so we have prebuilt copies of the various .data files in `eng/prebuilts`, which we update manually as needed with files built from the source tree which are produced with matching checksums from a series of full-tree builds

premun:
@directhex let's discuss this at the sync.

PR #6, opened 2022-12-09
Remove obsolete text-only packages

MichaelSimons:
The checked in text-only packages are useless because they are not being updated via the vmr synchronization process.  The package versions get updated regularly and these versions are not up-to-date.  Additionally the infrastructure to build them is getting removed with https://github.com/dotnet/source-build/issues/3138.  https://github.com/dotnet/source-build/issues/3160 is tracking the work to source-build these packages.


premun:
@MichaelSimons I expect to have the PR build for the VMR in place this week. Should we wait with merging this for that or this change doesn't affect anything?

MichaelSimons:
This change doesn't affect the tarball build.  One reason I would like to merge it now is to reduce the number of modified files after building the VMR.  Building the text-only packages introduced a large number of modified files (about 1/3 of the total).

PR #9, opened 2022-12-12
Update archiveArtifacts to correct version

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/3167.

I would like to update this manually to workaround https://github.com/dotnet/source-build/issues/3167 until a long term solution is implemented.  This issue is causing a few prebuilts which I would like to eliminate.


premun (approved):

premun:
@MichaelSimons do we want this file to be mirrored from installer instead?

MichaelSimons:
@MichaelSimons do we want this file to be mirrored from installer?

I am leaning that way but I would like to get @crummel's input on my comments https://github.com/dotnet/source-build/issues/3167#issuecomment-1346787703 before making the final decision.

PR #13, opened 2023-01-03
Add .gitattributes file

mthalman:
Adds a `.gitattributes` file to ensure consistent attributes for file types. It is an exact duplicate of the same file from the dotnet/installer repo: https://github.com/dotnet/installer/blob/b779a08108728457fafe44a70294ecdb7f14634e/.gitattributes

I originally ran into issues with the .sh scripts in the repo as a result of cloning on Windows and attempting to run the scripts in WSL. The .sh scripts had CRLF line endings instead of LF. This `.gitattributes` file ensures that all .sh scripts will have LF line endings no matter platform is used.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,55 @@
+*.doc  diff=astextplain
+*.DOC	diff=astextplain
+*.docx	diff=astextplain
+*.DOCX	diff=astextplain
+*.dot	diff=astextplain
+*.DOT	diff=astextplain
+*.pdf	diff=astextplain
+*.PDF	diff=astextplain
+*.rtf	diff=astextplain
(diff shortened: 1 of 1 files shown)


mthalman:
cc @premun

premun:
@mthalman please make this change into `dotnet/installer` under `src/SourceBuild/content`. It will be synchronized from there

premun:
Please also note that this might have performance implications on git when added to a repo of VMR's size.

All the repos have their `.gitattributes` synchronized, so the rules should already be set for the folders under `src/`. Is it possible to target more specific paths in the VMR? Maybe you hit a problem with scripts that are in some other top-level directory and can be targeted directly?

PR #19, opened 2023-01-24
tests: ExecuteRunWeb: update expected SIGTERM exit code on mono.

tmds:
Adjust the expected exit code of the WebScenarioTests smoke test so it passes on Mono.

With this change, all smoke tests pass on mono.

https://github.com/dotnet/runtime/issues/81093 tracks fixing Mono so the exit code is the same value as CoreCLR.

cc @crummel @MichaelSimons @omajid @uweigand

### DIFF
--- a/test/Microsoft.DotNet.SourceBuild.SmokeTests/DotNetHelper.cs
+++ b/test/Microsoft.DotNet.SourceBuild.SmokeTests/DotNetHelper.cs
@@ -14,6 +14,7 @@ internal class DotNetHelper
 {
     private static readonly object s_lockObj = new();
 
+    private static bool IsMonoRuntime { get; } = DetermineIsMonoRuntime();
     public static string DotNetPath { get; } = Path.Combine(Config.DotNetDirectory, "dotnet");
     public static string LogsDirectory { get; } = Path.Combine(Directory.GetCurrentDirectory(), "logs");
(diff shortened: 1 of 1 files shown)


test/Microsoft.DotNet.SourceBuild.SmokeTests/DotNetHelper.cs
```diff
@@ -230,5 +234,27 @@ private static string GetBinLogOption(string projectName, string command, string
         return $"/bl:{Path.Combine(LogsDirectory, $"{fileName}.binlog")}";
     }
 
+    private static bool DetermineIsMonoRuntime()
```
tmds, line 237: I'm open to suggestions on improving this method.

My first go was to check if the running .NET was using a Mono runtime, but that didn't work because my test host was actually CoreCLR (and .NET under test was mono).

Instead we need to peek in the .NET folder and find something to key off.
I'm currently using the presence of the `mono-gc.h` file.

MichaelSimons:
@tmds - The VMR is readonly in the .NET 8 timeframe.  It is not until .NET 9.0 that we will support edits and mirroring them back to the product repos.  This is documented [here](https://github.com/dotnet/dotnet#contribution).  This PR needs to be opened against installer.  The source-build infra exists at https://github.com/dotnet/installer/tree/main/src/SourceBuild/content.

tmds:
Replaced by https://github.com/dotnet/installer/pull/15341.

PR #28, opened 2023-05-12
Remove NuGet configs of several submodules

premun:
https://github.com/dotnet/installer/pull/16386
https://github.com/dotnet/source-build/issues/3152
https://github.com/dotnet/source-build/issues/3230


premun:
@MichaelSimons this one is needed as well

MichaelSimons:
I see this auto-merged.  Isn't that a potential issue given https://github.com/dotnet/installer/pull/16386 is not yet merged?  If another installer PR is merged before https://github.com/dotnet/installer/pull/16386 is merged then the NuGet.configs will get mirrored again.

premun:
Only problem could happen if there was a change to one of the `NuGet.config` files in between that. And I don't think that's likely.

premun:
As in, we'd have to merge a change from SB externals that would also be changing a submodule that just changed a config

PR #39, opened 2023-08-23
Remove excluded vstest license

mthalman:
Port of https://github.com/dotnet/dotnet/pull/37


mthalman:
I can't add reviewers so pinging @MichaelSimons and @premun.

premun (approved):
We should get you the permissions

PR #40, opened 2023-09-07
Update README.md

danmoseley:
Is this a reasonable clarification? I don't see any place where we explain why we need the "V" and it would be nice to make clear how most contributors will be unaffected.


premun:
@danmoseley I like the new clarification but the VMR is still readonly. Please open the PR against this file instead: https://github.com/dotnet/installer/blob/main/src/VirtualMonoRepo/README.template.md

PR #45, opened 2023-10-31
Add a placeholder `xliff-tasks` file to fix scanning

premun:
Fixes https://dev.azure.com/dnceng/internal/_build/results?buildId=2305189&view=logs&j=fa5c2909-0e29-5671-d5eb-16d6c40e0c4a&t=2a0bd486-3f67-5c59-da0c-b84d88848a4b&l=15 where the repo was removed artificially and a bit prematurely


mthalman (approved):
What is the long-term fix for this? What is causing this to still be scanned?

ViktorHofer (approved):

premun:
@mthalman the proper way would be to remove the mapping and the synchronization would kill the repo and the git-info files and everything would be taken care of. However, we were not able to do that because most of the repositories still reference xliff-tasks in their `Version.Details.xml`.
So to remove the code, we decide to cloak the repo fully and keep the record in. But delete the code.

Unfortunately, the code scan doesn't account for this. It will get fixed once we clean the Version.Details.xml and the mapping will get flagged as extraneous. Then we remove it with this file.

PR #55, opened 2024-01-25
Delete cloaked file from wpf repo

mthalman:
Don't merge until https://github.com/dotnet/installer/pull/18387 is merged.


jkotas:
Would it be better to cloak the whole WPF repo from the Linux distro drop instead of dealing with individual files? I assume that it is where we want to land.

mthalman:
> Would it be better to cloak the whole WPF repo from the Linux distro drop instead of dealing with individual files? I assume that it is where we want to land.

Yes, something to consider moving forward.

ViktorHofer:
> I assume that it is where we want to land.

That's not clear yet. The aspirational goal is still a unified code base for both MSFT and source-only builds. As discussed on Teams exhaustively, it's clear if that will be possible. I think we shouldn't cloak more files out than necessary until we figured out and decided if source-only builds will have a separate branch long-term.

That said I totally understand the angle that you are coming from here.

MichaelSimons (approved):

PR #62, opened 2024-02-07
Move target that copies to final location to non source-only location

ViktorHofer:


ViktorHofer:
damn, again wrong repo

PR #63, opened 2024-02-07
Remove cloaked aspire samples

ellahathaway:
Related to https://github.com/dotnet/installer/pull/18556

This PR removes the cloaked aspire samples from the VMR.


ellahathaway:
cc @dotnet/source-build-internal

MichaelSimons (approved):

mthalman:
This change didn't flow to the internal/release/8.0.1xx branch. @ellahathaway - Can you make a PR for the same change there?

@premun - Should dotnet-dotnet be hooked up to the internal-code-merge pipeline? I don't know the implications of that.

premun:
I think an internal merge is definitely not possible because it would overwrite the `source-manifest.json` file and also the contents of the VMR would no longer match a SHA in the source repository - it might end up having a change from each branch which would result in contents that do not match any SHA from the original repo.

Since this is a frequent operation, I am not sure whether we should not consider something like auto-removal of cloaked files. But it sounds dangerous. I don't see any other way than that or just manually removing files in all affected branches.

Or maybe we could also have a pipeline which we would manually run as we can see the list of cloaked files in the scans. So maybe we could run a pipeline that would delete those files and commit to a branch. Maybe with an approval in between?

ellahathaway:
> This change didn't flow to the internal/release/8.0.1xx branch. @ellahathaway - Can you make a PR for the same change there?

Done!

PR #65, opened 2024-02-12
Re-enable disable repositories [TEST]

ViktorHofer:


src/winforms/eng/SourceBuild.props
```diff
@@ -4,6 +4,5 @@
   <PropertyGroup>
     <GitHubRepositoryName>winforms</GitHubRepositoryName>
     <SourceBuildManagedOnly>true</SourceBuildManagedOnly>
-    <InnerBuildArgs>$(InnerBuildArgs) -warnNotAsError:MSB3026</InnerBuildArgs>
```
ViktorHofer, line 7: @jkoritzinsky I noticed that while the -warnNotAsError setting is respected in the repo's inner build, it's not respected in the repo's outer build which caused the subsequent build steps to not be entered. This might be an msbuild bug:

image

We run more targets after the `RunInnerSourceBuildCommand` target, i.e. to create the publish manifest.

For more details, see the AzDO build from the "try fixes" commit.

ViktorHofer:
CI is green.

Submitted PRs from this branch:
1. https://github.com/dotnet/arcade/pull/14475
2. https://github.com/dotnet/installer/pull/18618
3. https://github.com/dotnet/winforms/pull/10864

src/winforms/eng/SourceBuild.props
jkoritzinsky: I guess when I tested it, I didn't get the race condition. Can we update the outer build invocation to also suppress warnAsError for this code?

ViktorHofer: I think that I fixed the overall race via https://github.com/dotnet/winforms/pull/10864

PR #71, opened 2024-02-26
Test exit code fix

ViktorHofer:


eng/common/tools.ps1
```diff
@@ -850,7 +850,8 @@ function MSBuild-Core() {
     }
 
     # When running on Azure Pipelines, override the returned exit code to avoid double logging.
-    if ($ci -and $env:SYSTEM_TEAMPROJECT -ne $null) {
+    # Skip this when the build is nested as part of the VMR.
+    if ($ci -and $env:SYSTEM_TEAMPROJECT -ne $null -and $productBuild -eq $null -and $properties -notlike "*DotNetBuildRepo=true*") {
```
ViktorHofer: >  The variable '$productBuild' cannot be retrieved because it has not been set.

Looking for the right syntax...

PR #94, opened 2024-05-20
Add version to newest VS tools description 

ilonatommy:
Fixes https://github.com/dotnet/runtime/issues/93314.

### Questions:
It's the same as https://github.com/dotnet/runtime/pull/102446. Do we need to make the change in both repos?


ilonatommy:
Should automatically flow in after runtime's PR is merged.

PR #97, opened 2024-06-12
Add --target-rid top level option

am11:
Closes https://github.com/dotnet/source-build/issues/4413

cc @tmds

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -12,7 +12,7 @@ usage()
   echo "Common settings:"
   echo "  --binaryLog                     Create MSBuild binary log (short: -bl)"
   echo "  --configuration <value>         Build configuration: 'Debug' or 'Release' (short: -c)"
-  echo "  -r, --target-rid <value>        Runtime identifier to be used by the build. e.g. alpine.3.18-arm64, fedora.37-x64, freebsd.13-arm64, ubuntu.19.10-x64"
+  echo "  --target-rid, -r <value>        Runtime identifier to be used by the build. e.g. alpine.3.18-arm64, fedora.37-x64, freebsd.13-arm64, ubuntu.19.10-x64"
   echo "  --verbosity <value>             Msbuild verbosity: q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic] (short: -v)"
   echo ""
 
--- a/build.sh
+++ b/build.sh
@@ -287,8 +287,6 @@ source $scriptroot/eng/common/native/init-os-and-arch.sh
 source $scriptroot/eng/common/native/init-distro-rid.sh
 initDistroRidGlobal "$os" "$arch" ""
 if [[ -n "$target_rid" ]]; then
-  # override with user's preferred RID
-  export __DistroRid="$target_rid"
   properties+=( "/p:TargetRid=$target_rid" )
 fi
 
--- a/build.sh
+++ b/build.sh
@@ -287,7 +287,8 @@ source $scriptroot/eng/common/native/init-os-and-arch.sh
 source $scriptroot/eng/common/native/init-distro-rid.sh
 initDistroRidGlobal "$os" "$arch" ""
 if [[ -n "$target_rid" ]]; then
-  __DistroRid="$target_rid"
+  # overwrite with user's preferred RID
+  export __DistroRid="$target_rid"
   properties+=( "/p:TargetRid=$target_rid" )
 fi
 
--- a/build.sh
+++ b/build.sh
@@ -287,7 +287,7 @@ source $scriptroot/eng/common/native/init-os-and-arch.sh
 source $scriptroot/eng/common/native/init-distro-rid.sh
 initDistroRidGlobal "$os" "$arch" ""
 if [[ -n "$target_rid" ]]; then
-  # overwrite with user's preferred RID
+  # override with user's preferred RID
   export __DistroRid="$target_rid"
   properties+=( "/p:TargetRid=$target_rid" )
 fi
--- a/build.sh
+++ b/build.sh
@@ -12,6 +12,7 @@ usage()
   echo "Common settings:"
   echo "  --binaryLog                     Create MSBuild binary log (short: -bl)"
   echo "  --configuration <value>         Build configuration: 'Debug' or 'Release' (short: -c)"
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -12,6 +12,7 @@ usage()
   echo "Common settings:"
   echo "  --binaryLog                     Create MSBuild binary log (short: -bl)"
   echo "  --configuration <value>         Build configuration: 'Debug' or 'Release' (short: -c)"
+  echo "  -r, --target-rid <value>        Runtime identifier to be used by the build. e.g. alpine.3.18-arm64, fedora.37-x64, freebsd.13-arm64, ubuntu.19.10-x64"
```
am11, line 15:
```suggestion
  echo "  --target-rid, -r <value>        Runtime identifier to be used by the build. e.g. alpine.3.18-arm64, fedora.37-x64, freebsd.13-arm64, ubuntu.19.10-x64"
```

build.sh
```diff
@@ -280,6 +286,10 @@ fi
 source $scriptroot/eng/common/native/init-os-and-arch.sh
 source $scriptroot/eng/common/native/init-distro-rid.sh
 initDistroRidGlobal "$os" "$arch" ""
+if [[ -n "$target_rid" ]]; then
+  __DistroRid="$target_rid"
```
am11, line 290:
```suggestion
  # overwrite with user's preferred RID
  export __DistroRid="$target_rid"
```

build.sh
```diff
@@ -12,6 +12,7 @@ usage()
   echo "Common settings:"
   echo "  --binaryLog                     Create MSBuild binary log (short: -bl)"
   echo "  --configuration <value>         Build configuration: 'Debug' or 'Release' (short: -c)"
+  echo "  --target-rid, -r <value>        Runtime identifier to be used by the build. e.g. alpine.3.18-arm64, fedora.37-x64, freebsd.13-arm64, ubuntu.19.10-x64"
```
tmds: Suggestion to make more clear what this rid does, and what it does not do (https://github.com/dotnet/source-build/issues/4419):

```suggestion
  echo "  --rid, -r <value>               Overrides the rid that is produced by the build. e.g. alpine.3.18-arm64, fedora.37-x64, freebsd.13-arm64, ubuntu.19.10-x64"
```

build.sh
```diff
@@ -280,6 +286,11 @@ fi
 source $scriptroot/eng/common/native/init-os-and-arch.sh
 source $scriptroot/eng/common/native/init-distro-rid.sh
 initDistroRidGlobal "$os" "$arch" ""
+if [[ -n "$target_rid" ]]; then
+  # overwrite with user's preferred RID
```
am11, line 290:
```suggestion
  # override with user's preferred RID
```

build.sh
```diff
@@ -280,6 +286,11 @@ fi
 source $scriptroot/eng/common/native/init-os-and-arch.sh
 source $scriptroot/eng/common/native/init-distro-rid.sh
 initDistroRidGlobal "$os" "$arch" ""
+if [[ -n "$target_rid" ]]; then
+  # override with user's preferred RID
+  export __DistroRid="$target_rid"
```
tmds:
```suggestion
```

I think we can leave this out.

build.sh
```diff
@@ -100,6 +102,10 @@ while [[ $# > 0 ]]; do
       configuration=$2
       shift
       ;;
+    -target-rid|-r)
```
tmds, line 105:
```suggestion
    -rid|-r)
```

am11: I thought settled on --target-rid?

build.sh
am11: That affects OutputRID calculation in Directory.Build.props.

tmds: I'm not sure what you mean. For runtime, we're passing the rid down here:

https://github.com/dotnet/runtime/blob/40787436cba1b732a2507aaba2cd2a1e6f809a50/eng/DotNetBuild.props#L58

If there are places where `__DistroRid` overrides `TargetRid` then we should fix them.

build.sh
tmds: I think _target_ adds to the confusion of https://github.com/dotnet/source-build/issues/4419.
For this top-level argument, imo we can just name it `--rid`.

build.sh
am11: I think you are right, once `OutputRid` is overriden, we only use `__DistroRid` for outputOS calculation in `Directory.Build.props`: https://grep.app/search?q=__DistroRid&filter[repo][0]=dotnet/runtime

am11, line 291:
```suggestion
```

build.sh
tmds: @ViktorHofer thoughts?

am11:
@tmds, I think we can get rid of calculation in this repo: https://github.com/dotnet/dotnet/blob/3999c972cc2e8330a922e40a60ca89545756d450/Directory.Build.props#L54 It is only used for runtime where we have this calculation. Besides it's the old logic borrowed from runtime.

build.sh
ViktorHofer: I would prefer target-rid here with a good documentation explaining that this really just determines the name and nothing else. I feel that consistency is more important here.

Btw should we add the switch to the Windows script as well?

And last but not least this PR needs to target the sdk repo.

tmds: fwiw, I know _target_ is meant to be similar as how gcc has a _build_, _host_ and _target_ platform.
In contrast to that, a target rid does not describe a target platform. It is just a name.
I think that difference is what causes the confusion in https://github.com/dotnet/source-build/issues/4419.

am11: > this PR needs to target the sdk repo.

@ViktorHofer how would that work? VMR build.sh is different than the one in SDK.

ViktorHofer: https://github.com/dotnet/sdk/blob/main/src/SourceBuild/content/build.sh

We don't accept any PRs in this repository. This is just a mirror of code from various repos. Eventually the VMR will become its own source-of-truth but not until the beginning of .NET 10. See https://github.com/dotnet/dotnet?tab=readme-ov-file#contribution

am11: Ah, didn't knew it. Will close it.

tmds:
> @tmds, I think we can get rid of calculation in this repo: 

We need this in case the user doesn't specify a rid. This is responsible to make non-portable builds use the rid calculated by `init-distro-rid.sh`.

am11:
Closing in favor of https://github.com/dotnet/sdk/pull/41540.

PR #103, opened 2024-08-22
Update comments

mastermike501:


MichaelSimons:
@mastermike501 - The dotnet repo is read-only.  It is a projection of the individual product repos that make up the .NET Product.  This is [documented in the Readme](https://github.com/dotnet/dotnet?tab=readme-ov-file#code-flow). This PR will need to be opened up against the [dotnet/runtime repo](https://github.com/dotnet/runtime).

mastermike501:
@MichaelSimons ah I see!

PR #118, opened 2024-12-03
remove roslyn-infrastructure-current-swat

arunchndr:


MichaelSimons:
@arunchndr, Can you provide some context on this PR?  The VMR does not currently accept PRs rather they should be made to the contributing repos.

arunchndr:
> @arunchndr, Can you provide some context on this PR? The VMR does not currently accept PRs rather they should be made to the contributing repos.

ok we can ignore this then. The contributing repo PR is also out via https://github.com/dotnet/roslyn-analyzers/pull/7495. We are just trying to remove that team and are getting rid of its usages.

PR #125, opened 2024-12-18
[DO NOT MERGE] Update testing analyzers

sebastienros:
Investigating source build dependencies
c.f. https://github.com/dotnet/aspnetcore/pull/59516#discussion_r1889309469

### DIFF
--- a/src/aspnetcore/eng/Dependencies.props
+++ b/src/aspnetcore/eng/Dependencies.props
@@ -66,8 +66,8 @@ and are generated based on the last package release.
     <LatestPackageReference Include="Microsoft.Extensions.Diagnostics.Testing" />
     <LatestPackageReference Include="Microsoft.Extensions.TimeProvider.Testing" />
     <LatestPackageReference Include="Microsoft.Win32.Registry" />
-    <LatestPackageReference Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit" />
-    <LatestPackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.XUnit" />
+    <LatestPackageReference Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing" />
+    <LatestPackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing" />
     <LatestPackageReference Include="Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing" />
     <LatestPackageReference Include="Microsoft.OpenApi" />
     <LatestPackageReference Include="Microsoft.OpenApi.Readers" />
--- a/src/aspnetcore/eng/SourceBuildPrebuiltBaseline.xml
+++ b/src/aspnetcore/eng/SourceBuildPrebuiltBaseline.xml
@@ -18,6 +18,8 @@
     <UsagePattern IdentityGlob="System.Composition.Runtime/*8.0.0*" />
     <UsagePattern IdentityGlob="System.Composition.TypedParts/*8.0.0*" />
     <UsagePattern IdentityGlob="System.Security.Cryptography.Pkcs/5.0.*" />
+    <UsagePattern IdentityGlob="System.Security.Cryptography.Pkcs/6.0.*" />
+    <UsagePattern IdentityGlob="System.Formats.Asn1/8.0.*" />
 
     <!-- These are what the analyzers are built against. They are overridden in full source build.
          It may be possible to generate SBRPs -->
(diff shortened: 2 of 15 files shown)


ellahathaway:
aspnetcore error:

```
/__w/1/s/src/aspnetcore/eng/tools/RepoTasks/RepoTasks.csproj : error NU1605: Warning As Error: Detected package downgrade: System.Formats.Asn1 from 10.0.0-alpha.1.24617.1 to 8.0.1. Reference the package directly from the project to select a different version.  [/__w/1/s/.dotnet/sdk/10.0.100-alpha.1.24555.54/NuGet.targets]
/__w/1/s/src/aspnetcore/eng/tools/RepoTasks/RepoTasks.csproj : error NU1605:  RepoTasks -> NuGet.Packaging 6.13.0-rc.32767 -> System.Formats.Asn1 (>= 10.0.0-alpha.1.24617.1)  [/__w/1/s/.dotnet/sdk/10.0.100-alpha.1.24555.54/NuGet.targets]
/__w/1/s/src/aspnetcore/eng/tools/RepoTasks/RepoTasks.csproj : error NU1605:  RepoTasks -> System.Formats.Asn1 (>= 8.0.1) [/__w/1/s/.dotnet/sdk/10.0.100-alpha.1.24555.54/NuGet.targets]
  Failed to restore /__w/1/s/src/aspnetcore/eng/tools/RepoTasks/RepoTasks.csproj (in 461 ms).
  Build failed with exit code 1. Check errors above.
```

I suggest trying to add an entry for the package to Version.Details.xml.

sebastienros:
This PR is adding the reference and a Version.Details property set to 8.0.1 explicitly because it was pulling a 8.0.0 or lower, at least in the aspnetcore repos. Confirmed by the fact that this build error doesn't appear in the aspnetcore repos. Now if it detects a version downgrade I can just remove the new explicit reference.

I will merge main to the PR in dotnet/aspnetcore in case this changed very recently.

ellahathaway:
> the reference and a Version.Details

In source-build, a few steps are required to ensure that the version gets lifted properly:

- A version prop in Versions.props
- A package reference that uses the version property
- An entry in Version.Details.xml for the dependency

You had 2/3 of the three. I suspect that because you're missing the 3rd item, the version was being downgraded. I will investigate further with my own local build.

sebastienros:
> An entry in Version.Details.xml for the dependency

Correct, didn't realize this one

ellahathaway:
I think I figured it out.

The version prop should match the dependency name. So the prop should be `SystemFormatsAsn1Version` instead of `RepoTasksSystemFormatsAsn1Version`. Then the entry needs to be added to the Version.Details.xml file.

When I did this in my local copy and built, I did not encounter the version downgrade error.

sebastienros:
The downgrade in this repos is also due to the fact that aspnetcore uses `Nuget.Packaging@6.2.4` while VMR uses `Nuget.Packaging@6.13.0-rc.32767`. So Asn1@8.0.1 is an upgrade for aspnetcore and a downgrade for VMR.

Since SourceBuild was successful with the last build (by removing the Asn1 version altogether hence the downgrade), can I just focus of the original problem which was to understand if the following was ok in aspnetcore:

```
    <UsagePattern IdentityGlob="System.Security.Cryptography.Pkcs/5.0.*" />
    <UsagePattern IdentityGlob="System.Security.Cryptography.Pkcs/6.0.*" />
```

ellahathaway:
It appears that the original problem of

```
    <UsagePattern IdentityGlob="System.Security.Cryptography.Pkcs/5.0.*" />
    <UsagePattern IdentityGlob="System.Security.Cryptography.Pkcs/6.0.*" />
```

is fine. I don't see that as a problem.

That aside, do you intend to remove the Asn1 8.0.1 dependency in you aspnetcore PR? If not, your changes will break source-build in the VMR, so you will have to do what I suggested above before your PR can be merged.

sebastienros:
> do you intend to remove the Asn1 8.0.1 dependency in you aspnetcore PR

Hope it won't be a chicken and egg issue, I added it because the aspnetcore build would fail otherwise (source build verification step), but this is breaking VMR ;) Looking into updating Nuget.Packaging instead now with the hope that it makes this addition unnecessary on aspnetcore, and then will be upgraded in VMR with 6.13.0-rc.32767, since it seems to happen already.

ellahathaway:
If you do https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet/commit/a28d681b019aa14316f42cde010a0a84b1e07932/ and add an entry for asn1 to the version.details file, that will fix the SB failure in the VMR.

Edit: I see you updated your aspnetcore PR with those changes. LGTM, thanks.

sebastienros:
I did the same as you at first, and then it failed with a message telling me it was missing the entry in the `Automated` section. So I moved it there, fingers crossed.

PR #154, opened 2025-02-20
Add ThirdPartyNoticesTemplatePath to source mappings

dkurepa:
In https://github.com/dotnet/arcade-services/pull/4473 we made PCS read ThirdPartyNoticesTemplatePath from the source-mappings, so we're updating them here too. This is only used in the service, so we don't need to backport this change, as older branches use the `darc vmr` commands

### DIFF
--- a/src/source-mappings.json
+++ b/src/source-mappings.json
@@ -14,6 +14,8 @@
 
     "sourceMappingsPath": "src/sdk/src/VirtualMonoRepo/source-mappings.json",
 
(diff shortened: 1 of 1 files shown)


oleksandr-didyk (approved):

michalpavelka (approved):

adamzip (approved):

premun:
This must be done in `dotnet/sdk`

PR #164, opened 2025-03-19
Make devcontainer usable in VS Code

jaredpar:
This does a few things:

- Changes init.sh so that it installs the .NET SDK
- Puts `dotnet` on `$PATH` so it can be used in VS Code

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -18,5 +18,8 @@
             ]
         }
     },
+    "remoteEnv": {
+        "PATH": "${containerEnv:PATH}:/workspaces/dotnet/.dotnet"
+    },
     "onCreateCommand": ".devcontainer/init.sh"
 }
--- a/.devcontainer/init-toolset.sh
+++ b/.devcontainer/init-toolset.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+vmr_dir="${1:-/workspaces/dotnet}"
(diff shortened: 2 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -18,5 +18,8 @@
             ]
         }
     },
+    "remoteEnv": {
+        "PATH": "${containerEnv:PATH}:/workspaces/dotnet/.dotnet"
```
jaredpar: Not thrilled about hard coding the path here but couldn't find a good sub to use. Tried using `${localWorkspaceDir}` but that was giving out paths in `/var`

MichaelSimons:
@jaredpar - This PR will need to be made against the sdk if you intend to have it merged.  Until we switch to the flat flow, PRs for the VMR infra have to be opened against the [sdk](https://github.com/dotnet/sdk/tree/main/src/SourceBuild/content) and will get synced into the VMR.

.devcontainer/devcontainer.json
akoeplinger: in dotnet/runtime we use this:

```suggestion
        "PATH": "${containerWorkspaceFolder}/.dotnet:${containerEnv:PATH}",
```

akoeplinger (approved):

akoeplinger:
Now that the flat flow switch is done we could get this one in

ViktorHofer (approved):

PR #175, opened 2025-04-10
Rename SourceBuild.props to DotNetBuild.props

ViktorHofer:
This is the only repository left (after nuget-client and command-line-api for which there are PRs open) that still uses the SourceBuild.props extension point.

This change can't be made in aspire as there's no branch anymore that flows into the PR.


MichaelSimons:
Can you explain what is the driver for this?  With Aspire eventually going away in 10, why is this needed?

ViktorHofer:
This is so that I can clean-up the extension point prior to Aspire getting removed from the VMR. IOW the VMR is the only place where I can make this change and aspire is the only repo that uses the old naming convention.

This is very similar to https://github.com/dotnet/dotnet/commit/a2b8b7692a936dd557d3d956d5894f65459ccdd1 which was also committed directly into the VMR to remove a patch as the change couldn't be made to the aspire repo (as there's no backflow).

ViktorHofer:
Force merging -the PR failures are known (container path issues).

PR #181, opened 2025-04-17
Yield MSBuild node during repo builds

rainersigwald:
Yielding allows a node to do other work while waiting for a task to complete. Since there is very little work happening inside the overall coordination build, this can dramatically improve MSBuild's ability to schedule the repos.

For the PR `Ubuntu2404_Ubuntu_BuildTests_x64` leg:

[Before](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1017778&view=logs&j=879e79ff-d5e4-564a-9322-bee9d0a651ca&t=ff1d5a7f-5035-532c-d1a9-0c1a0debbf81):
image

[This PR](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1019927&view=logs&j=879e79ff-d5e4-564a-9322-bee9d0a651ca&t=ff1d5a7f-5035-532c-d1a9-0c1a0debbf81): image

Note though that elapsed time in that leg is almost the same, despite what looks like earlier scheduling. Not quite sure why yet.


repo-projects/Directory.Build.targets
```diff
@@ -539,7 +539,8 @@
     <Exec Command="$(FullCommand)"
           WorkingDirectory="$(ProjectDirectory)"
           EnvironmentVariables="@(EnvironmentVariables);@(BuildEnvironmentVariable)"
-          IgnoreStandardErrorWarningFormat="true">
+          IgnoreStandardErrorWarningFormat="true"
+          YieldDuringToolExecution="true">
```
ViktorHofer, line 506: So this solved one yield but there was still another yield because we had an outer + inner build per VMR inner repo. This infra complexity got removed with https://github.com/dotnet/dotnet/commit/2774010bee5a0e2a8a554bf87490d9eacebc39ab.

Let's rerun and see what changes now.

rainersigwald: I wouldn't expect much change since the outermost MSBuild invocation (repo-level projects) is still doing "exec {repo build}", and that outermost invocation is the one we were attacking.

ViktorHofer: Right but didn't the outer->inner repo Exec one (that is now removed) also utilize a node and blocked it?
FWIW I posted times in the UB chat and those now actually look much better (only ~2min after arcade and before command-line-api starts and in total ~8m after arcade until runtime starts) than what we had before.

rainersigwald: That should be from a disconnected universe of nodes, though--we don't share nodes between command-line invocations.

ViktorHofer: Oh interesting, didn't know that.

AR-May: There should be earlier scheduling happening due to the change indeed, which I would expect, although possibly still on the same level as before removing inner build. But question is - does it then translate to smaller build times. It is not guaranteed.

AR-May:
It seems like this change lead to a small perf improvement. Aggregated data on 10 measurements points for each scenario:
| Metric               | Before Fix | After Fix | Percentage Change |
|----------------------|------------|-----------|-------------------|
| Average TotalMinutes | 75.413333  | 74.251667 | -1.54%            |
| Max TotalMinutes     | 90.933333  | 89.716667 | -1.34%            |
| Min TotalMinutes     | 69.216667  | 70.933333 | 2.48%             |
| Median TotalMinutes  | 75.158333  | 72.750000 | -3.20%            |

AR-May:
There were some issues with builds in the CI (failures to access feeds to download packages), but it seems like they are unrelated to the changes. I think we can safely merge this PR.

AR-May (approved):

mmitche:
Yay! Thanks!

PR #160, opened 2025-02-28
MSBuild -graph for repo build graph

rainersigwald:

PR #185, opened 2025-04-22
[WIP] [NO-REVIEW] [NO-MERGE] repo build graph

AR-May:
Duplicate of https://github.com/dotnet/dotnet/pull/160 for testing purposes.


eng/merge-asset-manifests.proj
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <ProjectReference Include="$(TasksDir)Microsoft.DotNet.UnifiedBuild.Tasks\Microsoft.DotNet.UnifiedBuild.Tasks.csproj" />
+    <ProjectReference Include="$(RepoProjectsDir)$(RootRepo).proj" />
```
ViktorHofer: I think this also needs metadata to not actually build the project:

```suggestion
    <ProjectReference Include="$(RepoProjectsDir)$(RootRepo).proj" BuildReference="false" />
```

AR-May: Let's try to add this as well, thank you!

ViktorHofer:
@rainersigwald probably is the right person to help here. It looks like the added P2P items get actually built. An easy workaround/fix would be to not define those when `'$(Test)' == 'true'`.

rainersigwald:
I chatted with @AR-May briefly yesterday; the "right" fix is to not create `ProjectReference` items but add the right graph-description items to understand the repo projects.

An easy workaround/fix would be to not define those when `'$(Test)' == 'true'`.

But I love this idea!

ViktorHofer:
@AR-May I'm going to close this PR given that the sources changed significantly since this experiment was started. We now use P2Ps. It would be interesting to explore whether we can use static graph but that's now a lower priority I guess.

Feel free to reopen if you still wanted to test something in this PR.

PR #226, opened 2025-04-25
Move sdk yml to vmr orchestrator

ViktorHofer:


eng/pipelines/templates/stages/vmr-scan.yml
```diff
@@ -16,17 +16,14 @@ stages:
...
-        cd src/sdk
-        ../../.dotnet/dotnet tool restore
+        .dotnet/dotnet tool restore
       displayName: Initialize tooling
-      workingDirectory: $(Build.SourcesDirectory)
 
     - script: |
         set -e
-        sha=`../../.dotnet/dotnet darc vmr get-version --vmr "$(Build.SourcesDirectory)" sdk`
+        sha=`.dotnet/dotnet darc vmr get-version --vmr "$(Build.SourcesDirectory)" sdk`
```
ViktorHofer: @premun any idea what `sdk` should get changed to?

premun: The whole step should be removed. It was putting SDK's SHAs on VMR builds so that we can correlate SDK and VMR pipeline runs (so that SB releases can compare those).

PR #227, opened 2025-04-25
Remove WarnAsError opt-out from msbuild

ViktorHofer:
Contributes to https://github.com/dotnet/dotnet/issues/1848
Follow-up on https://github.com/dotnet/sdk/commit/7532b5d3f8284ddc857641bfde46568d240759e1
Now possible with https://github.com/dotnet/msbuild/commit/d5ac662019bc8456cc48f3e1b340f55572e6d059


akoeplinger (approved):

ViktorHofer:
> CSC : error CS9057: Analyzer assembly 'D:\a\_work\1\s\.dotnet\sdk\10.0.100-preview.4.25217.29\Sdks\Microsoft.NET.Sdk\codestyle\cs\Microsoft.CodeAnalysis.CodeStyle.dll' cannot be used because it references version '5.0.0.0' of the compiler, which is newer than the currently running version '4.14.0.0'. [D:\a\_work\1\s\src\msbuild\src\Xunit.NetCore.Extensions\Xunit.NetCore.Extensions.csproj::TargetFramework=net472]

@jjonescz @jaredpar Is the 5.0.0.0 roslyn version intentional? The VMR uses a P4 SDK: https://github.com/dotnet/dotnet/blob/e2bf5238a81b548bf5e77a7548cf50d05b562778/global.json#L3

jjonescz:
Yes, 5.0 is the latest version of Roslyn in main and in P4: https://github.com/dotnet/sdk/blob/release/10.0.1xx-preview4/eng/Versions.props#L227

ViktorHofer:
~~Any idea why the 4.14.0 version of the compiler is being used here? I thought that when setting `BuildWithNetFrameworkHostedCompiler`, it uses the latest version?~~

OK the msbuild repo is getting build with core roslyn. So I guess we are just using an too old version.

jjonescz:
It looks like the build is using toolset compiler package:

image

that's in MSBuild where the version is hard-coded to 4.14: https://github.com/dotnet/msbuild/blob/d40ed5884d80b92a5fc13adadb5ab3a893243b52/eng/Versions.props#L85

ViktorHofer:
@jjonescz I'm surprised that the toolset compiler package is even respected when the SDK compiler version is newer. Is that by design? Isn't this also a torn state that ideally would get resolved automatically?

jjonescz:
Ideally they shouldn't use the toolset package at all: https://github.com/dotnet/msbuild/blob/d40ed5884d80b92a5fc13adadb5ab3a893243b52/eng/Versions.props#L23

jjonescz:
> I'm surprised that the toolset compiler package is even respected when the SDK compiler version is newer. Is that by design?

The toolset is a temporary measure and if you leave it on long term, you will be broken (that's literally in its description), that's by design, there are no smart checks around it. In fact, it makes sense to me that it takes precedence over anything else, that's its purpose (maybe you want to try out an older compiler version? why not)

ViktorHofer:
Given that many repos in the core stack use the toolset package by default (by setting the property), I wonder if we could add some smartness to it. This will break every time roslyn revs the assembly version and an SDK is used with that newer version.

This will become a problem with the VMR. We could of course ask teams to not use the toolset package but there are usually reasons for its consumption. I.e. in runtime we reference it to always get the latest compiler independent from the .NET SDK used. This is mainly as we want to depend on latest features.

jjonescz:
Perhaps such smartness can be added to arcade, but the package itself should be simple and always take precedence IMO, it's meant to override everything because it's meant for testing a specific compiler version.

jjonescz:
> This will become a problem with the VMR.

Why not just force set `UsingToolMicrosoftNetCompilers=false` in the VMR?

ViktorHofer:
> Why not just force set UsingToolMicrosoftNetCompilers=false in the VMR?

This would break runtime whenever it depends on newer language features from the toolset compiler than the compiler in the .NET SDK. That's quite often.

ViktorHofer:
> Perhaps such smartness can be added to arcade, but the package itself should be simple and always take precedence IMO

Yes I assume we could check `$(RoslynVersion)`, compare it with the toolset version and not use it when the latter is older?

jjonescz:
> This would break runtime whenever it depends on newer language features from the toolset compiler than the compiler in the .NET SDK. That's quite often.

Right, we don't build with latest Roslyn in the VMR as I thought for a minute, makes sense.

ViktorHofer:
Hmm, ok I don't see a property like `RoslynVersion`. I thought that existed. Is there any other property that gives me the roslyn version (ideally package version) inserted into the .NET SDK?

jaredpar:
> I'm surprised that the toolset compiler package is even respected when the SDK compiler version is newer. Is that by design?

The primary use case of the toolset compiler is that a customer upgrades to a new .NET SDK, hits a compiler bug that blocks them and needs fast relief. In that case the toolset package is used to temporarily downgrade the compiler. Hence the toolset package winning even when it's a lower version is by design. 

> Ideally they shouldn't use the toolset package at all:

This.

ViktorHofer:
@rainersigwald does msbuild need to use the OOB compiler package (instead of the compiler that comes with the .NET SDK) or is it just artifact of old times?

repo-projects/msbuild.proj
```diff
@@ -1,7 +1,8 @@
 <Project Sdk="Microsoft.Build.NoTargets">
 
   <PropertyGroup>
-    <BuildArgs>$(BuildArgs) $(FlagParameterPrefix)warnAsError $(ArcadeFalseBoolBuildArg)</BuildArgs>
+    <!-- MSBuild uses a too old version that creates a torn state when being used with a >= .NET 10 P4 SDK. -->
```
rainersigwald, line 4: that's interesting we should be updating it weekly and be very close to the same version . . .

rainersigwald (commented):
Probably the latter: https://github.com/dotnet/msbuild/pull/11767

PR #230, opened 2025-04-25
Fix source-build poison checks and tests

NikolaMilosavljevic:
Fixes: https://github.com/dotnet/source-build/issues/5052, https://github.com/dotnet/source-build/issues/5057

Also fixes https://github.com/dotnet/source-build/issues/5044 that exposed flakiness in poison tests due to missing SBRP attributes in targeting pack packages. With https://github.com/dotnet/source-build-reference-packages/pull/1239, all targeting pack assemblies now have attributes that allow correct and complete poison scanning.

This PR also fixes the layout infra for `Microsoft.NET.Build.Extensions.Tasks` to exclude .NET FX content from source-only builds.

This PR supersedes https://github.com/dotnet/sdk/pull/48550 and includes those changes.


ViktorHofer:
> This PR also fixes the layout infra for Microsoft.NET.Build.Extensions.Tasks to exclude .NET FX content from source-only builds.

Last time when I discussed this with others, we didn't want to change that as it has product impact. IIRC the package is needed for NETStandard support when targeting NETFramework. While that should be a niche scenario on Linux, it is possible today.

NikolaMilosavljevic:
> > This PR also fixes the layout infra for Microsoft.NET.Build.Extensions.Tasks to exclude .NET FX content from source-only builds.
> 
> Last time when I discussed this with others, we didn't want to change that as it has product impact. IIRC the package is needed for NETStandard support when targeting NETFramework. While that should be a niche scenario on Linux, it is possible today.

I see, that's interesting. While it would be easy to update this PR and remove that change (and update the baseline), maybe we could revisit that decision as part of this PR. We have removed all other instances of .NET FX binaries in source-built SDK. This assembly is the only one.

test/Microsoft.DotNet.SourceBuild.Tests/assets/PoisonTests/PoisonUsage.txt
```diff
@@ -1 +1,377 @@
-<PrebuiltLeakReport />
\ No newline at end of file
+<PrebuiltLeakReport>
+  <File Path="artifacts/assets/Release/dotnet-sdk-x.y.z/packs/NETStandard.Library.Ref/x.y.z/ref/netstandard2.1/Microsoft.Win32.Primitives.dll">
```
MichaelSimons, line 2: Are annotations allowed in this file?  It would be nice to have a direct comment calling out the reason this is allowed.

NikolaMilosavljevic: We don't. We use `BaselineHelper.CompareBaselineContents` for straight up diff after we fix-up versions/etc:
https://github.com/dotnet/dotnet/blob/efbfb137a317fb339a6b7be36af7188cc508dc95/test/Microsoft.DotNet.SourceBuild.Tests/PoisonTests.cs#L34
https://github.com/dotnet/dotnet/blob/main/test/Microsoft.DotNet.SourceBuild.Tests/BaselineHelper.cs#L48

If we added support for comments or additional attributes, we'd need to have a way to migrate those attributes to the new baseline, when we have a need to update it. New baseline gets generated from a new poison report and won't have this knowledge, that's stored in the checked-in baseline, unless we migrate it at the time new baseline is generated.

NikolaMilosavljevic:
> > This PR also fixes the layout infra for Microsoft.NET.Build.Extensions.Tasks to exclude .NET FX content from source-only builds.
> 
> Last time when I discussed this with others, we didn't want to change that as it has product impact. IIRC the package is needed for NETStandard support when targeting NETFramework. While that should be a niche scenario on Linux, it is possible today.

@ViktorHofer who should be included in this PR to help make decision on these specific files?

ViktorHofer:
I think @ericstj and @MichaelSimons are the right people to make a call here. @ericstj is my observation correct that this package is used to support .NET Standard when targeting .NET Framework < 4.7.2?

MichaelSimons:
> I think @ericstj and @MichaelSimons are the right people to make a call here. @ericstj is my observation correct that this package is used to support net461-net471 when targeting netstandard2.0?

If @ViktorHofer's observation is correct, I don't think we should be introducing a functional difference here between the MSFT and SB product.  If we want to support this for Linux, it should be supported in SB.

NikolaMilosavljevic:
> > I think @ericstj and @MichaelSimons are the right people to make a call here. @ericstj is my observation correct that this package is used to support net461-net471 when targeting netstandard2.0?
> 
> If @ViktorHofer's observation is correct, I don't think we should be introducing a functional difference here between the MSFT and SB product. If we want to support this for Linux, it should be supported in SB.

We've already had similar differences in .NET SDKs, produced by MSFT or SB builds. Those were captured in sdk-diff exclusions file when this file introduced as a helper for making sense of the diffs: https://github.com/dotnet/dotnet/blob/efbfb137a317fb339a6b7be36af7188cc508dc95/test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/SdkFileDiffExclusions.txt#L22-L37

However, I see that this would cause a regression. I will add these files back and update the poison baseline.

MichaelSimons:
> We've already had similar differences in .NET SDKs, produced by MSFT or SB builds.

The overarching goal we should be working towards is to reduce/eliminate difference.

NikolaMilosavljevic:
I've reverted the change for `Microsoft.NET.Build.Extensions.Tasks`, with https://github.com/dotnet/dotnet/pull/230/commits/4d7a3821e9b3d1e141e3078c1c8b204cd0ef0f57. We will keep .NET FX related assemblies, to stay in parity with MSFT-built SDK.

Also baseline was updated to reflect this change: https://github.com/dotnet/dotnet/pull/230/commits/027ece0b913657f53b5f51a42e075c813d9a788f

MichaelSimons (approved):

PR #260, opened 2025-04-28
Allow ecl-2.0 license

ellahathaway:
Porting change from preview4: https://github.com/dotnet/sdk/pull/48745


ellahathaway:
Closing this PR because the change that sparked this PR should be treated as a false-positive exclusion. Additionally, because the license scan tests are not currently running on main due to some changes to how we install scancode, I'd rather close this and wait for the automated PR to be created.

PR #264, opened 2025-04-28
VMR vs MSFT signing comparison

ellahathaway:
Closes https://github.com/dotnet/arcade/issues/15192

This PR introduces new tooling for comparing the MSFT and VMR artifacts, naming the signing status of their signed files. Some notable changes in this PR include:
- A new pipeline called [`dotnet-unified-build-comparison`](https://dev.azure.com/dnceng/internal/_build?definitionId=1447&_a=summary) for running the asset and signing comparison. This pipeline runs on every build of main and release branches and can also be run manually. On manual runs, signing asset comparison is run as specified. On automated runs, asset comparison is always run and signing comparision is run when `signEnabled` is true.
- `BuildComparer` has been reworked so that there are two sub-commands: `assets` and `signing`.

[Test pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2697521&view=results)


ellahathaway:
> @ellahathaway any reason why some YMLs are added to src/sdk? Also, GatherDrop.ps1 is already in the VMR under eng/

Ah I didn't realize that we had moved the vmr templates out of src/sdk/eng/pipelines. Same goes for GatherDrop.ps1, it used to be in src/sdk/eng. Will move them - thanks.

ellahathaway:
@dotnet/product-construction - gentle ping. Can I please get a review on this? Thanks!

eng/GatherDrops.ps1
```diff
@@ -9,14 +9,17 @@ param(
...
+  [String]$assetFilter = ".*",
+  [Switch]$nonShipping = $false
 )
 $jsonContent = Get-Content -Path $filePath -Raw | ConvertFrom-Json
 foreach ($repo in $jsonContent.repositories) {
     $remoteUri = $repo.remoteUri
     $commitSha = $repo.commitSha
     $path = "$outputPath$($repo.path)"
-    $darcCommand = "$darcPath gather-drop -c $commitSha -r $remoteUri --non-shipping --skip-existing --continue-on-error --use-azure-credential-for-blobs -o $path --github-pat $githubPat --azdev-pat $azdevPat --verbose --ci"
+    $darcCommand = "$darcPath gather-drop -c $commitSha -r $remoteUri --skip-existing --continue-on-error --use-azure-credential-for-blobs -o $path --github-pat $githubPat --azdev-pat $azdevPat --asset-filter $assetFilter --verbose --ci --include-released $(if ($nonShipping) { '--non-shipping' })"
```
mmitche, line 22: What's the reason for including released?

eng/tools/BuildComparer/SignCheckResults.cs
```diff
@@ -0,0 +1,57 @@
+using System;
```
mmitche, line 4: Nit: missing headers in a few files.

eng/tools/BuildComparer/AssetComparer.cs
```diff
@@ -0,0 +1,660 @@
...
+            })
+    { }
+
+    /// <summary>
+    /// Evaluates a single asset mapping for issues based on its type (Package or Blob).
+    /// </summary>
+    /// <param name="mapping">Asset mapping to evaluate</param>
+    protected override async Task EvaluateAsset(AssetMapping mapping)
+    {
+        if (mapping.AssetType == AssetType.Package)
```
mmitche, line 59: Does this also need ignorable assets like

https://github.com/dotnet/dotnet/pull/264/files#diff-a09aad65478b8f16340f0891ee525126753ccb9ae97dde52b5009d5bcbb4b299R70-R73

eng/tools/BuildComparer/SignCheckResults.cs
ViktorHofer: We should at some point enable a rule to check for this automatically. This needs an editorconfig entry. Can you please file an issue @ellahathaway?

eng/pipelines/templates/stages/vmr-compare.yml
```diff
@@ -0,0 +1,142 @@
...
+      - template: ../steps/vmr-compare.yml
+        parameters:
+          continueOnError: false
+          unifiedBuildRunId: ${{ parameters.unifiedBuildRunId }}
+          sourceManifestCommit: ${{ parameters.sourceManifestCommit }}
+          command: assets
+
+# On manual builds, run of comparisionType is not assets
+# On automated builds, run if eq(variables.signEnabled, 'true')
+  - ${{ if or(and(ne(parameters.comparisonType, 'assets'), eq(variables['Build.Reason'], 'Manual')), and(eq(variables.signEnabled, 'true'), ne(variables['Build.Reason'], 'Manual'))) }}:
```
mmitche, line 87: `eq(parameters.comparisonType, 'signing')`?

mmitche: Why compare manual builds if signing is not enabled?

eng/pipelines/templates/steps/vmr-compare.yml
```diff
@@ -0,0 +1,268 @@
...
+
+# This is used to workaround limitations with the available space
+# on the MacOS agents.
+- name: clean
+  type: boolean
+  default: false
+
+steps:
+- powershell: |
+    $build_id = '${{ parameters.unifiedBuildRunId }}'.Replace(' ', '')
```
mmitche: Is this just `.Trim`?

eng/GatherDrops.ps1
ellahathaway: Released artifacts are needed when someone triggers a manual run of the pipeline to compare artifacts across different branches, and one branch has been released (eg main vs preview3).

For example, I used this infra to compare main vs preview3 when we were signing main builds and we hadn’t yet branched for preview4. Another use case would be if there’s a regression and we want to compare the current artifacts against artifacts from the last release.

eng/pipelines/templates/steps/vmr-compare.yml
```diff
@@ -0,0 +1,268 @@
...
+    Write-Host "##vso[task.setvariable variable=UnifiedBuildRunId;isOutput=true]$build_id"
+    Write-Host "##vso[task.setvariable variable=UnifiedBuildBranch;isOutput=true]$build_branch"
+    Write-Host "##vso[task.setvariable variable=UnifiedBuildCommitSha;isOutput=true]$commit_sha"
+  displayName: Find associated build
+  name: GetBuildInfo
+  env:
+    AZURE_DEVOPS_EXT_PAT: $(System.AccessToken)
+
+- powershell: |
+    $source_manifest_commit = '${{ parameters.sourceManifestCommit }}'.Replace(' ', '')
```
mmitche: `.Trim()`?

mmitche (approved):

mmitche:
@ellahathaway The test pipeline didn't quite do a correct asset comparison. It fails to find the merged manifests and then doesn't compare any assets.

ellahathaway:
> @ellahathaway The test pipeline didn't quite do a correct asset comparison. It fails to find the merged manifests and then doesn't compare any assets.

The gather drop step fails with a 403 error when trying to download the merged manifest (known arcade issue that I’ve contacted FR about - they have a fix in flight), which is why the comparison failed. I will change the comparison to error when a manifest doesn’t exist.

Edit: Addressed in https://github.com/dotnet/dotnet/pull/264/commits/0e5559f01b8af720a37c23c8a864c023ad118901

eng/pipelines/templates/stages/vmr-compare.yml
ellahathaway: > `eq(parameters.comparisonType, 'signing')`?

`all` is an option. I can do  `or(eq(parameters.comparisonType, 'signing'), eq(parameters.comparisonType, 'all'))`, but I personally think that the current implementation is the cleanest.

> Why compare manual builds if signing is not enabled?

That is referring to a manual build of the comparison pipeline. If someone manually triggers it, they should be able to select whatever kind of comparison that they would like. For automatically triggered builds, the comparison type is determined based on the branch name (release vs main).

eng/tools/BuildComparer/AssetComparer.cs
ellahathaway: The ignorable assets are already skipped in the evaluation methods for each asset type: https://github.com/dotnet/dotnet/blob/bee11b0b2e2181f538418c5ae265685e9b0ab445/eng/tools/BuildComparer/AssetComparer.cs#L82-L86 and https://github.com/dotnet/dotnet/blob/bee11b0b2e2181f538418c5ae265685e9b0ab445/eng/tools/BuildComparer/AssetComparer.cs#L355-L359.

eng/tools/BuildComparer/SignCheckResults.cs
ellahathaway: Filed https://github.com/dotnet/source-build/issues/5113

ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/264/commits/a66362c13679d7cc2ad77b665b870d690f55c9d8

mmitche:
> > @ellahathaway The test pipeline didn't quite do a correct asset comparison. It fails to find the merged manifests and then doesn't compare any assets.
> 
> The gather drop step fails with a 403 error when trying to download the merged manifest (known arcade issue that I’ve contacted FR about - they have a fix in flight), which is why the comparison failed. I will change the comparison to error when a manifest doesn’t exist.
> 
> Edit: Addressed in [0e5559f](https://github.com/dotnet/dotnet/commit/0e5559f01b8af720a37c23c8a864c023ad118901)

Ahh that's right. That should be fixed now, in the latest rollout.

mmitche (approved):

eng/pipelines/templates/steps/vmr-compare.yml
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/264/commits/9ba391b302ab3d267069e2b780fa87432fdb7b3c

eng/pipelines/templates/steps/vmr-compare.yml
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/264/commits/9ba391b302ab3d267069e2b780fa87432fdb7b3c

PR #176, opened 2025-04-10
Remove VMR inner-build concept

ViktorHofer:
Fixes https://github.com/dotnet/source-build/issues/4249
Fixes https://github.com/dotnet/source-build/issues/3789
Fixes https://github.com/dotnet/source-build/issues/3807
Fixes https://github.com/dotnet/source-build/issues/4184
Fixes https://github.com/dotnet/source-build/issues/4444
Fixes https://github.com/dotnet/source-build/issues/5135

1. Remove the outer/inner-build concept for SB & UB. A repo's DotNetBuild.props file doesn't get imported anymore when using an Arcade SDK with this change.
2. Move runtime's DotNetBuild.props in runtime.proj in the VMR. Most of the content became unnecessary as we don't need to pass through args anymore with the inner build removal.
3. Remove the DotNetBuildInnerRepo and DotNetBuildPhase properties and replace its usage with the now only two inclusive properties. Update docs.
4. Remove now unnecessary DotNetBuild.props files (runtime, aspnetcore, nuget-client, source-build-externals). The remaining DotNetBuild.props files are for repos that are still on Arcade 9 (tooling repos).
5. Make the PrebuiltUsage an opt-in feature in the Arcade.Sdk (defaults to true for source-build) and untie it from the source-build infra so that it can be enabled for all kinds of builds.

Extra work for which I will submitted separate PRs:
- [x] Move the `WritePrebuiltUsageData` and `ReportPrebuiltUsage` targets that were executed only for dotnet.proj into finish-source-only.proj to better distinguish them from the newly added ArPow targets that got moved into the VMR orchestrator: https://github.com/dotnet/dotnet/commit/3492e4b399001edb2699f61e1ec6225bfd73f128
- [x] Place the "prebuilt-report" logs under `artifacts\log\<config>\` and avoid the now unnecessary copy step in the YML: https://github.com/dotnet/dotnet/commit/3492e4b399001edb2699f61e1ec6225bfd73f128
- [x] Remove unnecessary SourceBuild.props / DotNetBuild.props which only contained properties that were used for repo source-build pre-built intermediate package creation: https://github.com/dotnet/dotnet/commit/6f266c2fe6333766468531540151cb8dc56e8d27
- [x] The latest roslyn commit should be extracted and submitted as a PR into dotnet/roslyn: https://github.com/dotnet/roslyn/pull/78429
- [x] InnerClone equivalent functionality needs to be implemented in source-build-externals: https://github.com/dotnet/source-build-externals/pull/478
[...]

PR #280, opened 2025-04-29
Clean-up prebuilt infra in VMR and remove unnecessary prebuilt infra from ArPow Arcade

ViktorHofer:
Extracted from https://github.com/dotnet/dotnet/pull/176

1. Move the WritePrebuiltUsageData and ReportPrebuiltUsage targets that were executed only for dotnet.proj into finish-source-only.proj to better distinguish them from the ArPow targets which need to run per repo.

2. The prebuilt report files are now placed under the repo log directory for a better UX and to simplify publishing (except for SBRP + Arcade which need a re-bootstrap).

3. Remove the SourceBuild.props import in favor of the DotNetBuild.props one. All VMR repos already migrated to the latter.

4. Remove the `ValidateUsageAgainstBaseline` target from ArPow SB and simplify the code.

5. Only restore the Arcade SB tasks package and import AfterSourceBuild.proj when building source-only. The non-source-only build doesn't need those code paths anymore.


MichaelSimons:
What is the driver behind removing the prebuilt report when doing a repo level source-build?  This does provide some value.

ViktorHofer:
I didn't know that this is desirable to keep. I was under the impression that without intermediate packages, the prebuilt reports aren't useful. Should I bring that back?

All tasks? WritePackageUsageData, WriteUsageReports and ValidateUsageAgainstBaseline

MichaelSimons:
> I didn't know that this is desirable to keep. I was under the impression that without intermediate packages, the prebuilt reports aren't useful. Should I bring that back?

I lean towards keeping it in place.  It is helpful to easily get a list of the external dependencies e.g. prebuilts and where they come from in one report.  We don't need any of the validation or baseline infra, only the usage reports.

ViktorHofer:
@MichaelSimons @NikolaMilosavljevic the PR is now ready :)

NikolaMilosavljevic (approved):

PR #306, opened 2025-04-30
[main] Port "Use Arcade's publishing logic to publish a given VMR vertical"

mmitche:
Forward port https://github.com/dotnet/sdk/pull/47076 to main.

Most of this cherry-picked cleanly. There was one conflict with in the poison infra here https://github.com/dotnet/dotnet/commit/e2ac4269f3a7feed64bffcb42a321727d8f61e35 that I had to manually resolve.


src/sdk/src/SourceBuild/patches/arcade/0001-Add-loose-PDB-relative-blob-path.patch
ViktorHofer, line 1: Remove

mmitche:
> Revert the changes under src/sdk/eng/pipelines. We already moved all pipelines under eng/pipelines in the VMR. The files are still there in sdk because we haven't had a forward flow PR for a few days.

Those changes need to be moved into eng/pipelines, then.

ViktorHofer:
Yes

jkoritzinsky (commented):
LGTM once Viktor's feedback is addressed.

mmitche:
Latest test build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2699420&view=results

jkoritzinsky:
@mmitche we should apply https://github.com/dotnet/sdk/pull/48803 here as well

mmitche:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2702426&view=results

eng/Tools.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <Import Project="$(RepositoryEngineeringDir)VmrLayout.props" />
+
+  <!-- Override RestoreSources in Source-Build builds so the Arcade SDK restores its packages from offline sources. -->
+
+  <PropertyGroup>
+    <RestoreSources Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(IsTestRun)' != 'true'">$(PrebuiltPackagesPath);$(PreviouslySourceBuiltPackagesPath)</RestoreSources>
+  </PropertyGroup>
```
ViktorHofer, line 8: @jkoritzinsky I'm still not exactly sure why this was needed. What changed that now requires these feeds to be used when restoring Arcade's Tools.proj?

eng/finish-source-only.proj
```diff
@@ -160,6 +159,7 @@
                           OutputPath="$(ArtifactsLogDir)" />
   </Target>
 
+
```
ViktorHofer:
```suggestion
```

ViktorHofer:
@mmitche I resolved the merge conflict

eng/Tools.props
jkoritzinsky: We needed to change the restore sources to ensure that during an offline source-build that we restore the tools from the offline feeds even for Arcade's Tools.proj.

mmitche:
Latest build https://dev.azure.com/dnceng/internal/_build/results?buildId=2703214&view=results

eng/PublishSourceBuild.props
```diff
@@ -0,0 +1,119 @@
...
+  <UsingTask TaskName="Microsoft.DotNet.SourceBuild.Tasks.LeakDetection.CheckForPoison" AssemblyFile="$(MicrosoftDotNetSourceBuildTasksLeakDetectionAssembly)" TaskFactory="TaskHostFactory" Condition="'$(EnablePoison)' == 'true'" />
+  <Target Name="ReportPoisonUsage"
+          AfterTargets="DiscoverArtifacts"
+          DependsOnTargets="CreatePrivateSourceBuiltArtifactsArchive"
+          Condition="'$(EnablePoison)' == 'true'"
+          Inputs="$(MSBuildProjectFullPath)"
+          Outputs="$(BaseIntermediateOutputPath)ReportPoisonUsage.complete">
+    <ItemGroup>
+      <!-- Include dotnet-sdk-*.tar.gz -->
+      <SdkArchive Include="$(ArtifactsAssetsDir)**/dotnet-sdk-*$(ArchiveExtension)" />
```
jkoritzinsky, line 12: Alternatively, this could be `@(SdkTarball)` to grab the SDK that we're about to publish (defined in `Publishing.props` in a very similar manner).

eng/Tools.props
ViktorHofer: Why wasn't that necessary before?

eng/PublishSourceBuild.props
mmitche: This is post-publish though, right?

eng/Tools.props
jkoritzinsky: Before this change we weren't using `Tools.proj` in the orchestrator as we weren't using the Arcade build infrastructure (we were directly invoking msbuild on the root `build.proj`, not building using the bootstrapped Arcade's `Build.proj`)

eng/PublishSourceBuild.props
jkoritzinsky: This target is currently sequenced after the discovery of the artifacts that we will publish, but will run before publish. `SdkTarball` is initialized in the `DiscoverArtifacts` target that this runs after.

jkoritzinsky: Sorry, scratch that. This runs post-publish.

 However, the `ArtifactsAssetsDir` is pointing at the pre-publish directory. if you want to pull it from post-publish, you should pull from `SourceBuiltAssetsDir`.

mmitche: Given that they're all pulling from the pre-publish dirs, I think keeping it that way is the right thing to do for now?

ViktorHofer (approved):

eng/PublishSourceBuild.props
jkoritzinsky: Sounds good.

PR #309, opened 2025-04-30
Remove unnecessary DotNetBuild.props files and update codeowners

ViktorHofer:
Extracted from https://github.com/dotnet/dotnet/pull/176

These DotNetBuild.props files were only necessary to set GithubRepositoryName and other metadata that was read when creating the SB intermediate packages.

GithubRepositoryName now gets passed in from the VMR orchestrator as it's also read by Publish.proj in Arcade.

Remove the DotNetBuild.props files and update the codeowner files.


repo-projects/Directory.Build.props
```diff
@@ -129,6 +129,8 @@
     <CommonArgs>$(CommonArgs) /p:SourceBuiltAssetsDir=$(ArtifactsAssetsDir)</CommonArgs>
     <CommonArgs>$(CommonArgs) /p:SourceBuiltAssetManifestsDir=$(RepoAssetManifestsDir)</CommonArgs>
 
+    <CommonArgs>$(CommonArgs) /p:GitHubRepositoryName=$(RepositoryName)</CommonArgs>
```
MichaelSimons, line 132: Given this addition, does it make sense to remove this setting from all DotNEtBuild.props as well in this PR, or is that coming in a follow-up PR?

ViktorHofer: I can't touch the ones that use Arcade 9 and produce SB intermediate packages. The other DotNetBuild.props files (aspnetcore, fsharp, runtime, ...) that are on .NET 10 will go away with the outer/inner removal so there's not much benefit in removing those lines.

MichaelSimons: I should have been more clear, I was referring to the other repos on arcade 10.  Thanks for the explanation.

MichaelSimons (approved):

PR #327, opened 2025-05-01
Build scenario-tests in installers validation job

NikolaMilosavljevic:
Fixes: https://github.com/dotnet/source-build/issues/5053

Installer tests require scenario-tests package, which is not published anymore from any vertical.

This PR modifies installer tests project to build scenario-tests repo, as a dependency (project reference). Building scenario-tests, like any other repo in VMR, requires that UnifiedBuild.Tasks assembly is built first. That project is now also built as a required dependency of installer tests project.

- Separate traversal project was needed for build ordering of the dependencies.
- We don't want and don't need to build dependencies for scenario-tests project (repo), it builds OK without them. This required a new property, `SkipScenarioTestsDependencies`, to skip the build of project's references/dependencies.
- When skipping dependencies, we need to use the bootstrap arcade - same property is used in that condition. I think this is OK and self-explaining, without additional comments in the code.

Verification build was successful: https://dev.azure.com/dnceng/internal/_build/results?buildId=2699510&view=results

### DIFF
--- a/repo-projects/scenario-tests.proj
+++ b/repo-projects/scenario-tests.proj
@@ -13,8 +13,8 @@
     <UseBootstrapArcade Condition="('$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' == 'true') or '$(SkipScenarioTestsDependencies)' == 'true'">true</UseBootstrapArcade>
   </PropertyGroup>
 
-  <!-- In source only test mode, do not depend on other repositories as the scenario tests repo needs to be capable of being built
-       by itself in a standalone validation job. -->
+  <!-- In source only test mode, or if SkipScenarioTestsDependencies is set to 'true', do not depend on other repositories
+       as the scenario tests repo needs to be capable of being built by itself in a standalone validation job. -->
   <ItemGroup Condition="'$(DotNetSourceOnlyTestOnly)' != 'true' and '$(SkipScenarioTestsDependencies)' != 'true'">
     <RepositoryReference Include="arcade" />
     <RepositoryReference Include="command-line-api" />
--- a/repo-projects/scenario-tests.proj
+++ b/repo-projects/scenario-tests.proj
@@ -10,12 +10,15 @@
     <!-- Don't use the updated NuGet.config file that includes the live package feeds when testing source-only as in that configuration
          the Microsoft built packages should be used. -->
     <TestArgs Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' != 'true'">$(TestArgs) /p:TestRestoreConfigFile=$(OriginalNuGetConfigFile)</TestArgs>
-    <UseBootstrapArcade Condition="('$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' == 'true') or '$(SkipScenarioTestsDependencies)' == 'true'">true</UseBootstrapArcade>
+
+    <!-- In source only test mode, or if SkipScenarioTestsDependencies is 'true', do not depend on other repositories
+         as the scenario tests repo needs to be capable of being built by itself in a standalone validation job.
+         This also requires the use of bootstrap arcade. -->
+    <SkipScenarioTestsDependencies Condition="'$(DotNetSourceOnlyTestOnly)' == 'true'">true</SkipScenarioTestsDependencies>
+    <UseBootstrapArcade Condition="'$(SkipScenarioTestsDependencies)' == 'true'">true</UseBootstrapArcade>
   </PropertyGroup>
 
-  <!-- In source only test mode, or if SkipScenarioTestsDependencies is set to 'true', do not depend on other repositories
-       as the scenario tests repo needs to be capable of being built by itself in a standalone validation job. -->
-  <ItemGroup Condition="'$(DotNetSourceOnlyTestOnly)' != 'true' and '$(SkipScenarioTestsDependencies)' != 'true'">
+  <ItemGroup Condition="'$(SkipScenarioTestsDependencies)' != 'true'">
     <RepositoryReference Include="arcade" />
     <RepositoryReference Include="command-line-api" />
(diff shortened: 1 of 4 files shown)


repo-projects/scenario-tests.proj
```diff
@@ -10,12 +10,12 @@
...
          the Microsoft built packages should be used. -->
     <TestArgs Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' != 'true'">$(TestArgs) /p:TestRestoreConfigFile=$(OriginalNuGetConfigFile)</TestArgs>
-    <UseBootstrapArcade Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' == 'true'">true</UseBootstrapArcade>
+    <UseBootstrapArcade Condition="('$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' == 'true') or '$(SkipScenarioTestsDependencies)' == 'true'">true</UseBootstrapArcade>
   </PropertyGroup>
 
   <!-- In source only test mode, do not depend on other repositories as the scenario tests repo needs to be capable of being built
        by itself in a standalone validation job. -->
-  <ItemGroup Condition="'$(DotNetSourceOnlyTestOnly)' != 'true'">
+  <ItemGroup Condition="'$(DotNetSourceOnlyTestOnly)' != 'true' and '$(SkipScenarioTestsDependencies)' != 'true'">
```
mthalman: I think this can be simplified:

```suggestion
  <ItemGroup Condition="'$(UseBootstrapArcade)' != 'true'">
```

repo-projects/scenario-tests.proj
```diff
@@ -10,12 +10,12 @@
     <!-- Don't use the updated NuGet.config file that includes the live package feeds when testing source-only as in that configuration
          the Microsoft built packages should be used. -->
     <TestArgs Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' != 'true'">$(TestArgs) /p:TestRestoreConfigFile=$(OriginalNuGetConfigFile)</TestArgs>
-    <UseBootstrapArcade Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' == 'true'">true</UseBootstrapArcade>
+    <UseBootstrapArcade Condition="('$(DotNetBuildSourceOnly)' == 'true' and '$(DotNetSourceOnlyTestOnly)' == 'true') or '$(SkipScenarioTestsDependencies)' == 'true'">true</UseBootstrapArcade>
   </PropertyGroup>
 
   <!-- In source only test mode, do not depend on other repositories as the scenario tests repo needs to be capable of being built
        by itself in a standalone validation job. -->
```
mthalman: Comment should be updated to reflect all conditions.

repo-projects/scenario-tests.proj
NikolaMilosavljevic: I was thinking of simplifying the conditions for this `ItemGroup` and the condition for setting the `UseBootstrapArcade` property.

Originally `UseBootstrapArcade` property was set if both `DotNetBuildSourceOnly` and `DotNetSourceOnlyTestOnly` are `true`, but dependency build was skipped if only `DotNetSourceOnlyTestOnly` is `true`.

Should those two conditions be updated? Would `DotNetBuildSourceOnly` always be `true` if `DotNetSourceOnlyTestOnly` is? Perhaps we can use just the second property in both conditions.

repo-projects/scenario-tests.proj
NikolaMilosavljevic: Will do - thanks.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/327/commits/bd120ef6744a3c4689ba8644ae5a56578b4f5a2d

repo-projects/scenario-tests.proj
mthalman: > Would `DotNetBuildSourceOnly` always be `true` if `DotNetSourceOnlyTestOnly` is?

Yes, `DotNetSourceOnlyTestOnly` implicitly means a source-only build.

test/Microsoft.DotNet.Installer.Tests/installer-tests-prereqs.proj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.Build.Traversal">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+  </PropertyGroup>
+
+  <!-- Make sure that Microsoft.DotNet.UnifiedBuild.Tasks is built before building the scenario-tests project -->
+  <ItemGroup>
+    <ProjectReference Include="$(TasksDir)Microsoft.DotNet.UnifiedBuild.Tasks/Microsoft.DotNet.UnifiedBuild.Tasks.csproj" BuildInParallel="false" />
+    <ProjectReference Include="$(RepoProjectsDir)scenario-tests.proj" BuildInParallel="false" />
```
ViktorHofer:
```suggestion
    <ProjectReference Include="$(RepoProjectsDir)scenario-tests.proj" BuildInParallel="false" AdditionalProperties="SkipScenarioTestsDependencies=true" />
```

So that this works outside of YML as well. You can remove the property from YML with this.

NikolaMilosavljevic: Nice, thanks!

repo-projects/scenario-tests.proj
NikolaMilosavljevic: Great - will simplify that condition as well.

NikolaMilosavljevic: I've simplified the conditions to make them more straightforward, with https://github.com/dotnet/dotnet/pull/327/commits/e1ffe9a44880c9c3ecbc11ecb1e96e7f9ea16a52

mthalman (approved):

PR #331, opened 2025-05-01
Fix `CreateBaselineUpdatePR` tool deleting files

ellahathaway:
Closes https://github.com/dotnet/source-build/issues/5118

Switching to the dotnet/dotnet repo in [dotnet/dotnet#331](https://github.com/dotnet/dotnet/pull/331) caused the PR baseline update tool to incorrectly delete files when creating PRs. This happened because the the github API can only recursively fetch up to 100,000 files from a tree, and due to the size of the dotnet/dotnet repository, the entire test/ directory was excluded. As a result, the PR baseline update tool mistakenly identified the files as deleted when, in fact, they were not fetched.

This PR resolves the issue by ensuring that only a subset of files in the desired path are fetched from the original tree.

[Test pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2702494&view=results)

### DIFF
--- a/eng/tools/CreateBaselineUpdatePR/PRCreator.cs
+++ b/eng/tools/CreateBaselineUpdatePR/PRCreator.cs
@@ -66,18 +66,22 @@ public class PRCreator
             return treeItems;
         }
 
-        foreach (var item in treeResponse.Tree)
+        await Parallel.ForEachAsync(treeResponse.Tree, async (item, cancellationToken) =>
         {
             string path = Path.Combine(relativePath, item.Path);
             if (!path.StartsWith(desiredPath) && !desiredPath.StartsWith(path))
             {
-                continue;
+                return;
             }
 
             if (item.Type == TreeType.Tree)
             {
                 TreeResponse subTree = await _client.Git.Tree.Get(_repoOwner, _repoName, item.Sha);
-                treeItems.AddRange(await FetchOriginalTreeItemsAsync(subTree, targetBranch, desiredPath, path));
+                List<NewTreeItem> subTreeItems = await FetchOriginalTreeItemsAsync(subTree, targetBranch, desiredPath, path);
+                lock (treeItems)
+                {
+                    treeItems.AddRange(subTreeItems);
+                }
             }
             else
             {
@@ -89,9 +93,12 @@ public class PRCreator
                     Sha = item.Sha
                 };
 
-                treeItems.Add(newItem);
+                lock (treeItems)
+                {
+                    treeItems.Add(newItem);
+                }
             }
-        }
+        });
 
         return treeItems;
     }
--- a/eng/tools/CreateBaselineUpdatePR/PRCreator.cs
+++ b/eng/tools/CreateBaselineUpdatePR/PRCreator.cs
@@ -60,7 +60,7 @@ public class PRCreator
         string desiredPath,
         string relativePath = "")
     {
-        var treeItems = new List<NewTreeItem>();
+        List<NewTreeItem> treeItems = [];
         if (treeResponse == null)
         {
             return treeItems;
@@ -251,7 +251,7 @@ public class PRCreator
 
     private async Task<TreeResponse> CreateTreeFromItemsAsync(List<NewTreeItem> items, string path = "")
     {
-        var newTreeItems = new List<NewTreeItem>();
+        List<NewTreeItem> newTreeItems = [];
 
         var groups = items.GroupBy(item => Path.GetDirectoryName(item.Path));
         foreach (var group in groups)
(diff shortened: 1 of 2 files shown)


eng/tools/CreateBaselineUpdatePR/PRCreator.cs
```diff
@@ -40,29 +40,62 @@ public async Task<int> ExecuteAsync(
...
         return Log.GetExitCode();
     }
 
+    private async Task<List<NewTreeItem>> FetchOriginalTreeItemsAsync(
+        TreeResponse? treeResponse,
+        string targetBranch,
+        string desiredPath,
+        string relativePath = "")
+    {
+        var treeItems = new List<NewTreeItem>();
```
mthalman: I prefer this syntax because it's shorter.

```suggestion
        List<NewTreeItem> treeItems = [];
```

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
```diff
@@ -40,29 +40,62 @@ public async Task<int> ExecuteAsync(
...
+            string path = Path.Combine(relativePath, item.Path);
+            if (!path.StartsWith(desiredPath) && !desiredPath.StartsWith(path))
+            {
+                continue;
+            }
+
+            if (item.Type == TreeType.Tree)
+            {
+                TreeResponse subTree = await _client.Git.Tree.Get(_repoOwner, _repoName, item.Sha);
+                treeItems.AddRange(await FetchOriginalTreeItemsAsync(subTree, targetBranch, desiredPath, path));
```
mthalman: Since a network call is being done here, this could be made more efficient but executing things in parallel. Right now, each iteration of the `foreach` loop would potentially be making this `Tree.Get` call and waiting for the result. Instead, you could change this to run in parallel across the items in `treeResponse.Tree` so that all those network calls are done at once.

ViktorHofer: Not familiar with this code path but also consider adding retries for this network call.

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/331/commits/4d5896a56ed2026cc85abf490c34a60b1a8ae55b

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
ellahathaway: > Since a network call is being done here, this could be made more efficient but executing things in parallel. Right now, each iteration of the `foreach` loop would potentially be making this `Tree.Get` call and waiting for the result. Instead, you could change this to run in parallel across the items in `treeResponse.Tree` so that all those network calls are done at once.

Addressed in https://github.com/dotnet/dotnet/pull/331/commits/d305c5e27a435d4abc0446fcbca467dbfb2198ba

> Not familiar with this code path but also consider adding retries for this network call.

Addressed in https://github.com/dotnet/dotnet/pull/331/commits/d7ba22bc31064bf27b28478e8b48576da2beaf36

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
```diff
@@ -40,29 +43,69 @@ public async Task<int> ExecuteAsync(
...
+            }
+
+            if (item.Type == TreeType.Tree)
+            {
+                TreeResponse subTree = await ApiRequestWithRetries(() => _client.Git.Tree.Get(_repoOwner, _repoName, item.Sha));
+                List<NewTreeItem> subTreeItems = await FetchOriginalTreeItemsAsync(subTree, targetBranch, desiredPath, path);
+                lock (treeItems)
+                {
+                    treeItems.AddRange(subTreeItems);
+                }
```
mthalman: Consider using `ConcurrentBag<T>` instead of `List<T>`. `ConcurrentBag<T>` is thread-safe and wouldn't require you to implement locking here.

In addition, consider passing `ConcurrentBag<T>` as a parameter to `FetchOriginalTreeItemsAsync` that it can use to directly add items to. This frees you from having to enumerate through the result just to add it to another `ConcurrentBag<T>` as you're doing here with the call to `AddRange`.

ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/331/commits/940773fe4ac373d276fada49d5990d837658184e

mthalman (approved):

PR #400, opened 2025-05-06
Support for source only dev/ci builds

mthalman:
Fixes dotnet/source-build#5061

Adds support for source-only dev/ci (non-official) builds which allows source-only jobs to run in PR validation within the UB pipeline.

The changes can be classified in the following groups:
* Roslyn analyzer dependencies on Microsoft.CodeAnalysis: These versions needed to be updated because otherwise it would have caused a package downgrade error as described in https://github.com/dotnet/source-build/issues/5135#issuecomment-2857433837. In this case Microsoft.CodeAnalysis gets lifted to the live version, which has a version of 10.0.0-ci. But due to semantic versioning rules, that value is interpreted as a lower version than 10.0.0-preview.4.25225.6. But we really don't want it to be lifted at all. Instead, we can compile against an SBRP version. This works for most of the projects but can't be done for the tools (GenerateDocumentation) which are relied upon for execution in the build. So a ref assembly won't do there. Instead, these projects are explicitly configured to not use CPM to lift the version for the offending packages.
* There are cases in aspnetcore and sdk where CS9057 had to be suppressed. Search for `CS9057` in the changed files for a description of why that was needed.
* Test updates: Some tests rely on having access to Microsoft-built packages, such as portable packages, which obviously aren't produced by a source-only build. Given that this scenario is for PR validation and not all the build legs are running in that context, the tests won't have these Microsoft-built packages available. Instead, these tests are being disabled in this context.


am11:
Suggestions:
* add `--id, --build-id <value>` argument in top-level build scripts and set `OfficialBuildId` accordingly
* when it's not present, `DisableDevBuildAsDefaultForSourceOnly=true` is implied

Dev builds will look like:
```diff
-  ./build.sh --clean-while-building -sb --os linux --arch riscv64 -p:DisableDevBuildAsDefaultForSourceOnly=true
+  ./build.sh --clean-while-building -sb --os linux --arch riscv64
```

mthalman:
> add --id, --build-id <value> argument in top-level build scripts and set OfficialBuildId accordingly

For official builds, we will either not require a build ID as input at all or address the UX with https://github.com/dotnet/source-build/issues/5109.

mthalman:
Finally ready for review.

eng/pipelines/templates/stages/source-build-and-validate.yml
```diff
@@ -89,7 +89,8 @@ stages:
           pool: ${{ parameters.pool_Linux }}
         buildSourceOnly: true
         ${{ if eq(parameters.useMicrosoftBuildAssetsForTests, false) }}:
-          extraProperties: "/p:DisableDevBuildAsDefaultForSourceOnly=true"
+          ${{ if contains(variables['Build.DefinitionName'], '-source-build') }}:
+            extraProperties: "/p:DisableDevBuildAsDefaultForSourceOnly=true"
```
mmitche: Can you add a comment here as to why this is necessary?

am11: https://github.com/dotnet/source-build/issues/5109#issuecomment-2867158302 suggests it is a temporary workaround. I thought this PR is actually supposed to remove it or would it be removed after this change is consumed?

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/WritePackageVersionsProps.cs
```diff
@@ -89,6 +90,8 @@ public class WritePackageVersionsProps : Microsoft.Build.Utilities.Task
         /// </summary>
         public string VersionDetails { get; set; }
 
+        public bool IsPreviousPropsFile { get; set; }
```
mmitche, line 93: Nit: I would change this to an ITaskItem[] PropertySuffixes with default being VersionPropertySuffix and VersionPropertyAlternateSuffix if not set. Then the caller would set this to 'PreviousVersion' when constructing the previously SB props.

repo-projects/aspnetcore.proj
```diff
@@ -67,4 +67,18 @@
...
 
+  <!-- For non-official builds, use the Roslyn version contained in the N-1 build because ASP.NET Core needs to build and execute
+      a source generator which must be built targeting a version of Roslyn contained in the executing SDK. Otherwise, because of how
+      Arcade works for non-official builds, M.CA will get an assembly version of 42.42.42.42 and the source generator would target that
+      version leading to an error like the following: 
+        CSC error CS9057: Analyzer assembly '/repos/dotnet/src/aspnetcore/artifacts/bin/Microsoft.AspNetCore.Http.RequestDelegateGenerator/Release/netstandard2.0/Microsoft.AspNetCore.Http.RequestDelegateGenerator.dll' cannot be used because it references version '42.42.42.42' of the compiler, which is newer than the currently running version '5.0.0.0'.
+      By not lifting the M.CA version to the current version and instead using the N-1 version, we guarantee that it will match the version
+      containing in the executed SDK and avoiding this error.
+  -->
+  <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(OfficialBuild)' != 'true'">
```
mmitche, line 78: And in an official build, the version referenced will be what is running?

repo-projects/templating.proj
```diff
@@ -5,6 +5,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
+    <RepositoryReference Include="runtime" />
```
mmitche, line 8: Interesting that we hadn't seen the need for this yet.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Tools.proj
```diff
@@ -56,8 +57,10 @@
     <PackageReference Include="Microsoft.DotNet.Build.Tasks.VisualStudio" Version="$(MicrosoftDotNetBuildTasksVisualStudioVersion)" Condition="'$(UsingToolVSSDK)' == 'true'" IsImplicitlyDefined="true" />
   </ItemGroup>
 
-  <!-- Unlike the items above, we should restore M.D.B.T.F in source build scenarios since it's used in the inner build. -->
-  <ItemGroup>
+  <!-- Unlike the items above, we should restore M.D.B.T.F in source build scenarios since it's used in the inner build.
```
mmitche: nit: There's no inner build any longer.

src/roslyn/eng/Versions.props
```diff
@@ -96,7 +96,7 @@
   <!--
     Microsoft.CodeAnalysis.* versions used by the RoslynAnalyzers.
   -->
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' != 'true'">
```
mmitche, line 99: nit: Prime case for a `<Choose><When><Otherwise>`?

src/roslyn/src/RoslynAnalyzers/Microsoft.CodeAnalysis.Analyzers/Core/Microsoft.CodeAnalysis.Analyzers.csproj
```diff
@@ -8,7 +8,7 @@
       Restore would conclude that there is a cyclic dependency between Microsoft.CodeAnalysis and Microsoft.CodeAnalysis.Analyzers.
     -->
     <PackageId>*$(MSBuildProjectFile)*</PackageId>
-    <MicrosoftCodeAnalysisVersion Condition="'$(MicrosoftCodeAnalysisVersion)' == ''">$(MicrosoftCodeAnalysisVersionForCodeAnalysisAnalyzers)</MicrosoftCodeAnalysisVersion>
+    <MicrosoftCodeAnalysisVersion>$(MicrosoftCodeAnalysisVersionForCodeAnalysisAnalyzers)</MicrosoftCodeAnalysisVersion>
```
mmitche, line 11: :)

src/roslyn/src/RoslynAnalyzers/Tools/GenerateDocumentationAndConfigFilesForBrokenRuntime/GenerateDocumentationAndConfigFilesForBrokenRuntime.csproj
```diff
@@ -19,7 +19,18 @@
     <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic" VersionOverride="$(MicrosoftCodeAnalysisVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" VersionOverride="$(MicrosoftCodeAnalysisVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" VersionOverride="$(MicrosoftCodeAnalysisVersion)" />
+  </ItemGroup>
 
-    <PackageReference Include="System.Composition" />
+  <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(OfficialBuild)' == 'false'">
```
mmitche, line 24: Have the roslyn repo-SB legs been removed? If not, has this been tested in that context?

src/sdk/src/BuiltInTools/dotnet-watch/dotnet-watch.csproj
```diff
@@ -15,6 +15,16 @@
...
+
+    <!--
+      Disable CS9057 in source-only dev/ci builds. This is necessary because dotnet-watch distributes some ASP.NET Core
+      analyzers. These analyzers get built with a dependency on the live version of Microsoft.CodeAnalysis. In a dev/ci
+      build, the assembly version associated with Microsoft.CodeAnalysis in that case is 42.42.42.4242, set by Arcade in
+      dev/ci builds. So when building dotnet-watch with its analyzer dependencies, it will cause CS9057 because the 
+      analyzer assembly version is not the same as the compiler assembly version being used by the sdk that the project
+      is being built with. But this is fine because the analyzer is just being used here to distribute it with the tool.
+    -->
+    <NoWarn Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(OfficialBuild)' != 'true'">$(NoWarn);CS9057</NoWarn>
```
mmitche, line 27: Really appreciate the in-depth comments here.

eng/pipelines/templates/stages/source-build-and-validate.yml
mmitche: That kind of what I was wondering. Let's see what @mthalman has to say.

repo-projects/templating.proj
ViktorHofer: What requires this? These days runtime is only a toolset dependency in templating.

ViktorHofer:
> In this case Microsoft.CodeAnalysis gets lifted to the live version, which has a version of 10.0.0-ci. But due to semantic versioning rules, that value is interpreted as a lower version than 10.0.0-preview.4.25225.6.

Is that a fundamental problem that needs attention regardless of the changes in this PR?

> But due to semantic versioning rules, that value is interpreted as a lower version than 10.0.0-preview.4.25225.6. But we really don't want it to be lifted at all. Instead, we can compile against an SBRP version.

Why don't we want it to be lifted and prefer SBRP? Because of the above semver issue or because of a design decision?

> Instead, these projects are explicitly configured to not use CPM to lift the version for the offending packages.

Is that a sustainable pattern and if not, do we have ideas how to counter future issues around that?

repo-projects/aspnetcore.proj
```diff
@@ -67,4 +67,18 @@
     <ExtraPackageVersionPropsPackageInfo Include="MicrosoftCodeAnalysisVersion_LatestVS" Version="%24(MicrosoftCodeAnalysisCommonVersion)" />
   </ItemGroup>
 
+  <!-- For non-official builds, use the Roslyn version contained in the N-1 build because ASP.NET Core needs to build and execute
+      a source generator which must be built targeting a version of Roslyn contained in the executing SDK. Otherwise, because of how
+      Arcade works for non-official builds, M.CA will get an assembly version of 42.42.42.42 and the source generator would target that
+      version leading to an error like the following: 
+        CSC error CS9057: Analyzer assembly '/repos/dotnet/src/aspnetcore/artifacts/bin/Microsoft.AspNetCore.Http.RequestDelegateGenerator/Release/netstandard2.0/Microsoft.AspNetCore.Http.RequestDelegateGenerator.dll' cannot be used because it references version '42.42.42.42' of the compiler, which is newer than the currently running version '5.0.0.0'.
```
MichaelSimons, line 74:
```suggestion
        CSC error CS9057: Analyzer assembly '/repos/dotnet/src/aspnetcore/artifacts/bin/Microsoft.AspNetCore.Http.RequestDelegateGenerator/Release/netstandard2.0/Microsoft.AspNetCore.Http.RequestDelegateGenerator.dll'
        cannot be used because it references version '42.42.42.42' of the compiler, which is newer than the currently running version '5.0.0.0'.
```

eng/pipelines/templates/stages/source-build-and-validate.yml
mthalman: It was meant to support the standalone source build pipeline. But it's no longer necessary now that https://github.com/dotnet/dotnet/pull/515 is merged.

mthalman:
Hey folks, I'm going to get this prioritized to be merged and will address feedback in a follow-up. Dev builds need to be enabled to unblock things (see https://github.com/dotnet/source-build/issues/5135#issuecomment-2885555390).

src/roslyn/src/RoslynAnalyzers/Tools/GenerateDocumentationAndConfigFilesForBrokenRuntime/GenerateDocumentationAndConfigFilesForBrokenRuntime.csproj
mthalman: Testing these changes here: https://github.com/dotnet/roslyn/pull/78608

mthalman:
Getting a prebuilt for Microsoft.CodeAnalysis.Common.5.0.0-1.25230.108 coming from aspnetcore. This is due to now being dependent on the PSB and the PSB has some incorrect dependencies. The PSB contains Microsoft.CodeAnalysis.AnalyzerUtilities.5.0.0-1.25265.106 which has a dependency on Microsoft.CodeAnalysis.Common.5.0.0-1.25230.108. That Microsoft.CodeAnalysis.Common version isn't right, it should have the same version as AnalyzerUtilities.

mthalman:
> Getting a prebuilt for Microsoft.CodeAnalysis.Common.5.0.0-1.25230.108 coming from aspnetcore. This is due to now being dependent on the PSB and the PSB has some incorrect dependencies. The PSB contains Microsoft.CodeAnalysis.AnalyzerUtilities.5.0.0-1.25265.106 which has a dependency on Microsoft.CodeAnalysis.Common.5.0.0-1.25230.108. That Microsoft.CodeAnalysis.Common version isn't right, it should have the same version as AnalyzerUtilities.

It's because Microsoft.CodeAnalysis.AnalyzerUtilities has a package reference to Microsoft.CodeAnalysis.Common instead of a project reference: https://github.com/dotnet/dotnet/blob/c7f1d4f51e13d0d05687d7b22c8f4e3d5d12fd63/src/roslyn/src/RoslynAnalyzers/Microsoft.CodeAnalysis.AnalyzerUtilities/Microsoft.CodeAnalysis.AnalyzerUtilities.csproj#L21

This causes it to have a dependency on the PSB version of Microsoft.CodeAnalysis.Common (5.0.0-1.25230.108) instead of the live version.

mthalman:
> > Getting a prebuilt for Microsoft.CodeAnalysis.Common.5.0.0-1.25230.108 coming from aspnetcore. This is due to now being dependent on the PSB and the PSB has some incorrect dependencies. The PSB contains Microsoft.CodeAnalysis.AnalyzerUtilities.5.0.0-1.25265.106 which has a dependency on Microsoft.CodeAnalysis.Common.5.0.0-1.25230.108. That Microsoft.CodeAnalysis.Common version isn't right, it should have the same version as AnalyzerUtilities.
> 
> It's because Microsoft.CodeAnalysis.AnalyzerUtilities has a package reference to Microsoft.CodeAnalysis.Common instead of a project reference:
> 
> https://github.com/dotnet/dotnet/blob/c7f1d4f51e13d0d05687d7b22c8f4e3d5d12fd63/src/roslyn/src/RoslynAnalyzers/Microsoft.CodeAnalysis.AnalyzerUtilities/Microsoft.CodeAnalysis.AnalyzerUtilities.csproj#L21
> 
> This causes it to have a dependency on the PSB version of Microsoft.CodeAnalysis.Common (5.0.0-1.25230.108) instead of the live version.

This seems to be temporarily resolved by using the output of the stage 2 build as rebootstrap source: https://github.com/dotnet/dotnet/pull/598. This makes sense because the issue is that it's using the wrong version from the PSB artifacts. But in stage 2 the correct version is being output from stage 1 which is then selected as the PSB artifact in the stage 2 build.

mthalman:
Merging this now. I'll address the comments in a follow-up.

repo-projects/aspnetcore.proj
mthalman: In an official build, it uses the live version of Roslyn. That version seems to be compatible enough with the running SDK to not produce `CS9057`.

repo-projects/templating.proj
mthalman: It's because sdk has a dependency on `Microsoft.TemplateEngine.*` packages from templating. And templating has a dependency on `Microsoft.Extensions.Logging.Abstractions` from runtime. Without this runtime repo dependency, it gets the N-1 version of Abstractions instead of the live version. Meanwhile, sdk has a direct dependency on Abstractions which gets lifted to the live version. This causes a package downgrade error:

```
/repos/dotnet/src/sdk/src/Cli/Microsoft.TemplateEngine.Cli/Microsoft.TemplateEngine.Cli.csproj error NU1109: Detected package downgrade: Microsoft.Extensions.Logging.Abstractions from 10.0.0-preview.4.25211.17 to centrally defined 10.0.0-dev. Update the centrally managed package version to a higher version. 
 Microsoft.TemplateEngine.Cli -> Microsoft.TemplateEngine.Edge 10.0.100-dev -> Microsoft.Extensions.Logging.Abstractions (>= 10.0.0-preview.4.25211.17) 
 Microsoft.TemplateEngine.Cli -> Microsoft.Extensions.Logging.Abstractions (>= 10.0.0-dev) [/repos/dotnet/src/sdk/source-build.slnf]
```

PR #401, opened 2025-05-06
Move dac signing tooling into the VMR's eng folder

jkoritzinsky:
As we ship from the VMR now, we don't need to have the logic to sign the DAC in the runtime repo's official build any more. Move to the VMR's eng folder for simplicity.

Depends on https://github.com/dotnet/runtime/pull/115311 (removes the usage in runtime)


ViktorHofer (approved):

ViktorHofer:
@jkoritzinsky ready to merge?

jkoritzinsky:
@hoyosjs has a signing change he wants to make while the tooling is still in runtime. Once he's done that, I'll update this PR and merge it.

hoyosjs:
@jkoritzinsky since the runtime side no longer exists, I pushed here. I can add the auth.json changs to runtime separately and wait for them to flow if the backflow isn't easy right now - but this should be in a testable state.

jkoritzinsky:
Yeah backflow won't work super cleanly here (in fact the backflow will delete the files in the runtime repo).

So I'd PR manually to the various runtime release branches. I'll backport this for P5.

PR #457, opened 2025-05-08
Update CODEOWNERS

ViktorHofer:


akoeplinger (approved):

CODEOWNERS
```diff
@@ -9,3 +9,7 @@
 
 # Exclude the src folder which is team specific
 /src/
+
+# Exclude forward flow file changes to common files to reduce noise.
+/src/source-manifest.json
```
mmitche, line 14: I think this is redundant with the previous line?

akoeplinger: yeah but I'd keep it in case we change that to add codeowners for each `src/<repo>`

ViktorHofer: Yes, that was the intention here.

PR #461, opened 2025-05-08
Pass -productBuild and -sourceBuild switches

ellahathaway:
Closes https://github.com/dotnet/source-build/issues/4337


repo-projects/aspnetcore.proj
```diff
@@ -6,6 +6,8 @@
 
     <BuildActions Condition="'$(DotNetBuildSourceOnly)' != 'true'">$(FlagParameterPrefix)restore $(FlagParameterPrefix)all $(FlagParameterPrefix)pack $(FlagParameterPrefix)publish</BuildActions>
     <BuildActions Condition="'$(DotNetBuildSign)' == 'true'">$(BuildActions) $(FlagParameterPrefix)sign</BuildActions>
+    <BuildActions Condition="'$(DotNetBuildSourceOnly)' != 'true'">$(BuildActions) $(FlagParameterPrefix)productBuild</BuildActions>
+    <BuildActions Condition="'$(DotNetBuildSourceOnly)' == 'true'">$(BuildActions) $(FlagParameterPrefix)sourceBuild</BuildActions>
```
ViktorHofer: You already set these in D.B.props. This means that they are passed in twice.

```suggestion
```

repo-projects/nuget-client.proj
```diff
@@ -8,6 +8,8 @@
     <BuildActions></BuildActions>
     <!-- Call into a custom build script as nuget-client doesn't use Arcade infra. -->
     <BuildScript>$([MSBuild]::NormalizePath('$(ProjectDirectory)', 'eng', 'dotnet-build', 'build$(ShellExtension)'))</BuildScript>
+    <BuildArgs Condition="'$(DotNetBuildSourceOnly)' != 'true'">$(BuildArgs) $(FlagParameterPrefix)productBuild</BuildArgs>
+    <BuildArgs Condition="'$(DotNetBuildSourceOnly)' == 'true'">$(BuildArgs) $(FlagParameterPrefix)sourceBuild</BuildArgs>
```
ViktorHofer:
```suggestion
```

repo-projects/roslyn.proj
```diff
@@ -9,6 +9,7 @@
     <BuildActions Condition="'$(BuildOS)' == 'windows'">$(BuildActions) $(FlagParameterPrefix)pack</BuildActions>
     <BuildActions Condition="'$(BuildOS)' == 'windows'">$(BuildActions) $(FlagParameterPrefix)publish</BuildActions>
     <BuildActions Condition="'$(BuildOS)' == 'windows' and '$(DotNetBuildSign)' == 'true'">$(BuildActions) $(FlagParameterPrefix)sign</BuildActions>
+    <BuildActions Condition="'$(BuildOS)' == 'windows'">$(BuildActions) $(FlagParameterPrefix)productBuild</BuildActions>
```
ViktorHofer: same here

src/arcade/eng/common/build.sh
```diff
@@ -136,7 +136,7 @@ while [[ $# > 0 ]]; do
       restore=true
       pack=true
       ;;
-    -productBuild|-pb)
+    -productbuild|-pb)
```
ViktorHofer: Should we support `--product-build|--productbuild|-pb` everywhere consistently?

ViktorHofer: Same for the source-build switch above

repo-projects/aspnetcore.proj
ellahathaway: Aspnetcore, rosyn (on windows), and nuget.client don't not use the defaults, so don't we have to include them here?

ViktorHofer: Add this to BuildArgs instead of BuildActions, then it will work for all repos.

src/arcade/eng/common/build.sh
ellahathaway: Opened https://github.com/dotnet/arcade/pull/15824

repo-projects/aspnetcore.proj
ellahathaway: Fixed in https://github.com/dotnet/dotnet/pull/461/commits/a7f9d5ccc0d3815a09ecd41f2809b4359125d5b7

repo-projects/Directory.Build.props
```diff
@@ -155,6 +153,8 @@
   <!-- Build specific args -->
   <PropertyGroup>
     <BuildArgs>$(BuildArgs) -bl</BuildArgs>
+    <BuildArgs Condition="'$(DotNetBuildSourceOnly)' != 'true'">$(BuildArgs) $(FlagParameterPrefix)productBuild</BuildArgs>
+    <BuildArgs Condition="'$(DotNetBuildSourceOnly)' == 'true'">$(BuildArgs) $(FlagParameterPrefix)sourceBuild</BuildArgs>
```
ViktorHofer: Please pass these in as a `CommonArgs` (in the above property group). That makes them apply to both the build and test repo invocation (mainly for scenario-tests).

ViktorHofer (commented):
Looks good so far :) Please also pass the source-build switch in here instead of these properties: https://github.com/dotnet/dotnet/blob/f5d6cb4f2a429bed3999b119b1beff82047a8939/src/arcade/eng/common/core-templates/steps/source-build.yml#L59-L60

(The src/arcade eng/common files are the ones that backflow back into Arcade)

repo-projects/Directory.Build.props
ellahathaway: Good to know. I addressed this in https://github.com/dotnet/dotnet/pull/461/commits/70fda438e061d9b458e0c6d7531c64608b337c6f

repo-projects/Directory.Build.props
```diff
@@ -106,6 +106,8 @@
 
   <!-- Common args used for building and testing. -->
   <PropertyGroup>
+    <CommonArgs Condition="'$(DotNetBuildSourceOnly)' != 'true'">$(CommonArgs) $(FlagParameterPrefix)productBuild</CommonArgs>
```
ViktorHofer, line 109: @mmitche does this look right to you? We could also always unconditionally pass this switch in as source-build implies product build. We are following this infra: https://github.com/dotnet/arcade/blob/6b63184d8ff276115846b232816ece935a26a22c/eng/common/build.sh#L132-L144

mmitche: Yeah, I think you would only need one switch. SB implies PB

src/arcade/eng/common/core-templates/steps/source-build.yml
```diff
@@ -51,13 +51,12 @@ steps:
     ${{ coalesce(parameters.platform.buildScript, './build.sh') }} --ci \
       --configuration $buildConfig \
       --restore --build --pack -bl \
+      --source-build \
```
ViktorHofer, line 54: In the arcade build.sh, the `--source-build` switch already implies all these build actions: https://github.com/dotnet/arcade/blob/6b63184d8ff276115846b232816ece935a26a22c/eng/common/build.sh#L133-L137

I would double check whether that's also true for fsharp, aspnetcore, roslyn and nuget-client's custom scripts and if so, remove the actions here.

ViktorHofer: The intention here is to make this command match the local dev workflow as closely as possible. My feedback can also be addressed in a follow-up if it would cause additional work.

ellahathaway: I'd rather address that in a follow-up PR if that's alright. But I will file an issue for that work since it should be completed eventually.

ViktorHofer:
The prebuilts are coming from aspnetcore when it builds the repotasks separately.

image

You can see that in the repotasks binlog the DotNetBuildSourceOnly=true property isn't part of the MSBuild command line arguments.

---

As I was already looking at this, I updated the sources and pushed a commit. I hope that's OK. cc @wtgodbe  - yet another reason to not build repotasks differently and just make them be part of the normal msbuild invocation via a P2P.

src/aspnetcore/eng/build.ps1
```diff
@@ -479,6 +484,7 @@ try {
             /p:Configuration=Release `
             /p:Restore=$RunRestore `
             /p:Build=true `
+            /p:DotNetBuildRepo=$ProductBuild `
```
ViktorHofer: This one should now also only be passed in when ProductBuild is explicitly provided. Same for the one in build.sh

ViktorHofer:
OK resolved conflicts with https://github.com/dotnet/dotnet/pull/500. I will now stop touching this PR again ;)

ellahathaway:
Thanks for your help, Viktor :)

Just waiting on the nuget-client changes to flow in, then I'll update this PR and publish it.

ViktorHofer:
If you want to, you can already merge this change. If the content is identical then there won't be any conflicts in the forward-flow PR from nuget and also no backflow into nuget.

ViktorHofer (approved):

ellahathaway:
> If you want to, you can already merge this change. If the content is identical then there won't be any conflicts in the forward-flow PR from nuget and also no backflow into nuget.

Oh sweet, this is great news. Yes, the changes are identical, so I'll go ahead and merge

src/arcade/eng/common/core-templates/steps/source-build.yml
ellahathaway: Filed https://github.com/dotnet/arcade/issues/15831

ViktorHofer:
🎉🎉🎉

PR #465, opened 2025-05-09
Fix license scan timeout by adding container options

Winniexu01:
Fixes: https://github.com/dotnet/source-build/issues/5124

The license scanning step timeouts when processing large files in the Azure Linux 3.0 container. Azure Linux 3.0 container has slower performance in I/O and memory management compared to Ubuntu container.

Add container options and increase the timeout to ensure that the license scan completes successfully.


ViktorHofer (approved):

eng/pipelines/vmr-license-scan.yml
```diff
@@ -53,6 +53,7 @@ extends:
     containers:
       azurelinuxSourceBuildTestContainer:
         image: mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net10.0-source-build-test-amd64
+        options: '--memory=6g --privileged'
```
mthalman: What is the need for `privileged`?

Winniexu01: This was added to help me check the performance of Azure Linux 3.0 containers. Now delete it: https://github.com/dotnet/dotnet/pull/465/commits/be42ca35bccfd3463358f581d4524acebc81b543

mthalman (approved):

PR #531, opened 2025-05-13
Don't include license file in sfx packages with license expressions

ViktorHofer:
Very similar to https://github.com/dotnet/runtime/pull/115518

Don't include the LICENSE file in the shared framework produced packages if those set a PackageLicenseExpression.

As mentioned in the other issue, Arcade already validates that either the license expression or the license file is set.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
+++ b/src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
@@ -169,7 +169,8 @@
     </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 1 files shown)


ViktorHofer:
> artifacts/source-built-sdks/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets(611,5): error MSB3030: (NETCORE_ENGINEERING_TELEMETRY=Build) Could not copy the file "/__w/1/s/src/aspnetcore/LICENSE.TXT" because it was not found.

Interesting.

ViktorHofer:
@jkoritzinsky I'm inclined to close this as I don't have much resources to spend on this. Do you want to take this over or should we burry it for now?

jkoritzinsky:
I don't have cycles to spend on this right now either. We can close and return to it later.

PR #532, opened 2025-05-13
Don't publish loose Windows PDBs in PGO legs or from runtime

jkoritzinsky:
This should markedly reduce the number of loose PDBs we try to publish, saving time in publishing

### DIFF
--- a/repo-projects/Directory.Build.props
+++ b/repo-projects/Directory.Build.props
@@ -144,6 +144,9 @@
     <DefaultArtifactVisibility Condition="'$(PgoInstrument)' == 'true'">Vertical</DefaultArtifactVisibility>
     <!-- ShortStack builds only publish new assets from the root repository. All other assets are duplicates. -->
     <DefaultArtifactVisibility Condition="'$(ShortStack)' == 'true' and '$(MSBuildProjectName)' != '$(RootRepo)'">Vertical</DefaultArtifactVisibility>
+
+    <!-- Don't publish loose Windows PDBs for PGO assets. They're never used and they slow down publishing. -->
+    <PublishWindowsPdb>false</PublishWindowsPdb>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
--- a/repo-projects/Directory.Build.props
+++ b/repo-projects/Directory.Build.props
@@ -146,7 +146,7 @@
     <DefaultArtifactVisibility Condition="'$(ShortStack)' == 'true' and '$(MSBuildProjectName)' != '$(RootRepo)'">Vertical</DefaultArtifactVisibility>
 
     <!-- Don't publish loose Windows PDBs for PGO assets. They're never used and they slow down publishing. -->
-    <BuildArgs Condition="'$(PgoInstrument)' == 'true'">$(BuildArgs) /p:PublishWindowsPdb=$(PublishWindowsPdb)</BuildArgs>
+    <BuildArgs Condition="'$(PgoInstrument)' == 'true'">$(BuildArgs) /p:PublishWindowsPdb=false</BuildArgs>
   </PropertyGroup>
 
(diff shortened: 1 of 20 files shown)


repo-projects/Directory.Build.props
```diff
@@ -144,6 +144,9 @@
     <DefaultArtifactVisibility Condition="'$(PgoInstrument)' == 'true'">Vertical</DefaultArtifactVisibility>
     <!-- ShortStack builds only publish new assets from the root repository. All other assets are duplicates. -->
     <DefaultArtifactVisibility Condition="'$(ShortStack)' == 'true' and '$(MSBuildProjectName)' != '$(RootRepo)'">Vertical</DefaultArtifactVisibility>
+
+    <!-- Don't publish loose Windows PDBs for PGO assets. They're never used and they slow down publishing. -->
+    <PublishWindowsPdb>false</PublishWindowsPdb>
```
ViktorHofer: Missing condition?

repo-projects/runtime.proj
```diff
@@ -15,6 +15,9 @@
     <SkipSetTargetOS>true</SkipSetTargetOS>
     <BuildArgs>$(BuildArgs) $(FlagParameterPrefix)os $(TargetOS)</BuildArgs>
 
+    <!-- Don't publish Windows PDBs from the runtime repo. They're not used and they slow down publishing. -->
+    <PublishWindowsPdb>false</PublishWindowsPdb>
```
ViktorHofer: Can't we have this directly in the runtime repo, i.e. in its Publishing.props file?

jkoritzinsky: I was just following consistency with how we currently do it for sdk and deployment tools. I'll move those into the repos as well.

ViktorHofer: Yes I think that would be better to reduce differences and keep the VMR specialness minimal.

repo-projects/Directory.Build.props
```diff
@@ -144,6 +144,9 @@
     <DefaultArtifactVisibility Condition="'$(PgoInstrument)' == 'true'">Vertical</DefaultArtifactVisibility>
     <!-- ShortStack builds only publish new assets from the root repository. All other assets are duplicates. -->
     <DefaultArtifactVisibility Condition="'$(ShortStack)' == 'true' and '$(MSBuildProjectName)' != '$(RootRepo)'">Vertical</DefaultArtifactVisibility>
+
+    <!-- Don't publish loose Windows PDBs for PGO assets. They're never used and they slow down publishing. -->
+    <BuildArgs Condition="'$(PgoInstrument)' == 'true'">$(BuildArgs) /p:PublishWindowsPdb=$(PublishWindowsPdb)</BuildArgs>
```
ViktorHofer:
```suggestion
    <BuildArgs Condition="'$(PgoInstrument)' == 'true'">$(BuildArgs) /p:PublishWindowsPdb=false</BuildArgs>
```

repo-projects/deployment-tools.proj
```diff
@@ -1,5 +1,4 @@
 <Project Sdk="Microsoft.Build.NoTargets">
-
```
ViktorHofer: Please let's not work against each other. I like to add these new lines after project opening tags and before project closing tags (and our dotnet new templates have them as well) and I noticed you tend to remove them. I really think that the new lines here improves readability of msbuild files.

just a nit observation

src/msbuild/Directory.Build.props
```diff
@@ -41,6 +41,7 @@
     <PackageTags>MSBuild</PackageTags>
 
     <TargetMSBuildToolsVersion>Current</TargetMSBuildToolsVersion>
+    <PublishWindowsPdb Condition="'$(DotNetBuildOrchestrator)' == 'true'">false</PublishWindowsPdb>
```
mmitche, line 45: Can you add a comment to these indicating why the Orchestrator context is important?

mmitche (approved):

PR #535, opened 2025-05-13
Add official build triggers to main CI pipeline

mmitche:


mmitche:
Resolves https://github.com/dotnet/source-build/issues/5048

NikolaMilosavljevic (approved):

ViktorHofer:
Aren't we still using Azdo UI trigger which override these?

akoeplinger:
yeah for most of the pipelines mentioned in the comment at the top of the file we still override in the AzDO UI.
I guess we can clean that up once the source-build pipelines are integrated into the unified-build one... though I guess that won't work with the servicing branches

PR #545, opened 2025-05-14
Consolidate DotNetBuildRepo with DotNetBuild switch -> DotNetBuild

ViktorHofer:
Since the removal of the outer/inner build complexity in the VMR, DotNetBuildRepo is now always true for all kinds of .NET product builds: --source-build and --product-build.

The DotNetBuildRepo switch isn't necessary anymore now that the same can be achieved with the DotNetBuild switch. For anything that applies to both repo product builds and the VMR orchestrator, continue using the DotNetBuildOrchestrator prop.

---

Also apply some small follow-up improvements from the outer/inner removal.


repo-projects/Directory.Build.props
```diff
@@ -107,6 +107,8 @@
   <!-- Common args used for building and testing. -->
   <PropertyGroup>
     <CommonArgs Condition="'$(DotNetBuildSourceOnly)' != 'true'">$(CommonArgs) $(FlagParameterPrefix)productBuild</CommonArgs>
+    <!-- TODO: Rename to source-build when the new switch is available everywhere. -->
+    <CommonArgs Condition="'$(DotNetBuildSourceOnly)' == 'true'">$(CommonArgs) --sourceBuild</CommonArgs>
```
mmitche:
```suggestion
    <CommonArgs Condition="'$(DotNetBuildSourceOnly)' == 'true'">$(CommonArgs) $(FlagParameterPrefix)sourceBuild</CommonArgs>
```

src/deployment-tools/src/clickonce/Directory.Build.props
```diff
@@ -10,13 +10,13 @@
   </PropertyGroup>
 
   <!-- Platform is different during restore than during build. Therefore, restore everything when not building the product. -->
-  <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.proj' and ('$(DotNetBuild)' != 'true' or '$(DotNetBuildPass)' == '2') ">
+  <ItemGroup Condition="'$(MSBuildProjectExtension)' == '.proj' and '$(DotNetBuild)' != 'true'">
```
mmitche, line 13: This is changing the meaning of the condition here. Intentional?

src/aspire/eng/Build.props
```diff
@@ -1,18 +1,16 @@
 <Project>
-  <ItemGroup Condition="'$(DotNetBuildFromSource)' != 'true' and '$(DotNetBuild)' != 'true'">
+
+  <ItemGroup Condition="'$(DotNetBuild)' != 'true'">
```
mmitche, line 3: Aspire within the VMR is on 8.0 arcade...I don't think we can change this?

src/arcade/eng/common/build.ps1
```diff
@@ -127,7 +127,7 @@ function Build {
     /p:Deploy=$deploy `
     /p:Test=$test `
     /p:Pack=$pack `
-    /p:DotNetBuildRepo=$productBuild `
+    /p:DotNetBuild=$productBuild `
```
mmitche, line 130: Based on your recent change to copy the eng/common* files except when building arcade, will this have any effect on SBRP's build? It won't have the updated arcade.

src/deployment-tools/src/clickonce/Directory.Build.props
ViktorHofer: Yes. This was copied incorrectly from a repo that participates in BP2. Deployment-tools does not.

src/aspire/eng/Build.props
ViktorHofer: Aspire inside the VMR builds with live arcade :)

src/arcade/eng/common/build.ps1
ViktorHofer: Yes, that's by design. SBRP builds with the bootstrap Arcade SDK and therefore shouldn't get these script changes.

mmitche (approved):

PR #546, opened 2025-05-14
Resolve NuGet Audit errors in VMR builds

ViktorHofer:
Fixes https://github.com/dotnet/source-build/issues/5187


ViktorHofer:
```
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1103: Unable to find a stable package Microsoft.Build.Tasks.Core with version (>= 17.14.7)
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1103:   - Found 5 version(s) in /__w/1/s/prereqs/packages/previously-source-built/ [ Nearest version: 17.15.0-preview-25225-08 ]
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1103:   - Found 0 version(s) in /__w/1/s/prereqs/packages/prebuilt/
```

I think this means that https://github.com/dotnet/dotnet/blob/b60c10ba96f8af424454bb2c9ec3add6633ba763/eng/Versions.props#L48-L50 doesn't work for the sdk resolver project which is part of the init-source-build phase. @MichaelSimons any ideas?

MichaelSimons:
> @MichaelSimons any ideas?

The imports happen in the init-source-build phase before the PSB tarball is extracted therefore the PSB.versions.props doesn't get applied.  Possible solutions:

1. Is it feasible/possible to do re-import the sdk.props after the PSB initialization before BuildMSBuildSdkResolver?
1. Define the MSBuildSdkResolver proj to explicitly depend on the version from psb for SB.
1. Add the new version to SBRP and rebootstrap.

ViktorHofer:
Thanks. I just pushed a commit that fixes that. The sequence was already correct (the msbuildsdkresolver evaluates, restores and builds after the PSB archive is extracted) but there was an import mistake. The `PreviouslySourceBuiltPackageVersionsPropsFile` pointed to the wrong location because VmrLayout.props was imported too late.

eng/Versions.props
```diff
@@ -30,7 +30,7 @@
     <!-- command-line-api dependencies -->
     <SystemCommandLineVersion>2.0.0-beta5.25208.1</SystemCommandLineVersion>
     <!-- msbuild dependencies -->
-    <MicrosoftBuildVersion>17.12.6</MicrosoftBuildVersion>
+    <MicrosoftBuildVersion>17.14.8</MicrosoftBuildVersion>
```
rainersigwald: If it's useful to have a smaller change you can use 17.12.36.

ViktorHofer: Hah same thought :) Thanks

Directory.Build.props
```diff
@@ -45,6 +45,10 @@
 
   <!-- Manually import the Versions.props file when the Arcade SDK isn't used. -->
   <Import Project="$(RepositoryEngineeringDir)Versions.props" Condition="'$(SkipArcadeSdkImport)' == 'true'" />
+  <Import Project="$(RepositoryEngineeringDir)VmrLayout.props" />
+  <!-- For source only builds, override the version props with previously built artifacts.
+       If a version should be pinned, declare it after this import. -->
```
MichaelSimons: I feel like this sentence could be removed as it provides less value with this refactor and there is very little chance this should ever be used.

ViktorHofer:
The non-source-build jobs should be unblocked now but source-build ones need an update to use PSB to avoid prebuilts.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
```diff
@@ -6,8 +6,8 @@
 
   <ItemGroup>
     <!-- IncludeAssets=compile to treat these packages as targeting-packs only. The assemblies are available in the SDK. -->
-    <PackageReference Include="Microsoft.Build.Tasks.Core" IncludeAssets="compile" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" IncludeAssets="compile" />
+    <PackageReference Include="Microsoft.Build.Tasks.Core" IncludeAssets="compile" NoWarn="NU1901;NU1902;NU1903;NU1904" />
```
mthalman, line 9: nit: It'd be nice to have a shared ItemGroup to reference here for the NuGet warnings named something like `@(NuGetAuditWarning)`.

mthalman (approved):

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
rainersigwald: I think to be usable here it'd need to be a property.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Microsoft.DotNet.UnifiedBuild.Tasks.csproj
```diff
@@ -6,11 +6,11 @@
 
   <ItemGroup>
     <!-- IncludeAssets=compile to treat these packages as targeting-packs only. The assemblies are available in the SDK. -->
-    <PackageReference Include="Microsoft.Build.Tasks.Core" IncludeAssets="compile" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" IncludeAssets="compile" />
-    <PackageReference Include="NuGet.Protocol" IncludeAssets="compile" />
-    <PackageReference Include="NuGet.ProjectModel" IncludeAssets="compile" />
-    <PackageReference Include="Newtonsoft.Json" IncludeAssets="compile" />
+    <PackageReference Include="Microsoft.Build.Tasks.Core" IncludeAssets="compile" NoWarn="NU1901;NU1902;NU1903;NU1904" />
```
rainersigwald, line 9: Do these in-the-vmr projects actually need to reference Tasks.Core? Usually you don't and I think I got it building locally without (but I rebooted and lost my context there, sorry).

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
ViktorHofer: Agreed. This best belongs directly into nuget's NuGet.props so that customers benefit from the list as well: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/NuGet.props

@zivkan @nkolev92 any objections on that?

nkolev92: There's a lot of movement around audit, so I'd like for it to be considered in the scope of that work as a potential enhancement.

fwiw, suppressing warnings, I'd almost rather this be more difficult than easier for people to do.

src/arcade/eng/Versions.props
```diff
@@ -35,10 +35,10 @@
...
     <MicrosoftSymbolUploaderVersion>2.0.0-preview.1.23470.14</MicrosoftSymbolUploaderVersion>
     <!-- msbuild -->
-    <MicrosoftBuildFrameworkVersion>17.8.3</MicrosoftBuildFrameworkVersion>
-    <MicrosoftBuildTasksCoreVersion>17.8.3</MicrosoftBuildTasksCoreVersion>
-    <MicrosoftBuildUtilitiesCoreVersion>17.8.3</MicrosoftBuildUtilitiesCoreVersion>
-    <MicrosoftBuildVersion>17.8.3</MicrosoftBuildVersion>
+    <MicrosoftBuildFrameworkVersion>17.12.36</MicrosoftBuildFrameworkVersion>
+    <MicrosoftBuildTasksCoreVersion>17.12.36</MicrosoftBuildTasksCoreVersion>
+    <MicrosoftBuildUtilitiesCoreVersion>17.12.36</MicrosoftBuildUtilitiesCoreVersion>
+    <MicrosoftBuildVersion>17.12.36</MicrosoftBuildVersion>
```
rainersigwald, line 41: If you want to minimize (but I don't think there's any scenario where Arcade 10 needs to work on MSBuild 17.8):

```suggestion
    <MicrosoftBuildFrameworkVersion>17.8.29</MicrosoftBuildFrameworkVersion>
    <MicrosoftBuildTasksCoreVersion>17.8.29</MicrosoftBuildTasksCoreVersion>
    <MicrosoftBuildUtilitiesCoreVersion>17.8.29</MicrosoftBuildUtilitiesCoreVersion>
    <MicrosoftBuildVersion>17.8.29</MicrosoftBuildVersion>
```

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
nkolev92: We also have other features in mind that should eventually address the auditing of packages that we don't deploy like Microsoft.Build.Tasks.Core. 

https://github.com/nuget/home/issues/13860 might be related as well.

src/sourcelink/eng/Versions.props
```diff
@@ -12,9 +12,8 @@
     <!-- commandline -->
     <SystemCommandLineVersion>2.0.0-beta5.25208.1</SystemCommandLineVersion>
     <!-- msbuild -->
-    <MicrosoftBuildVersion>17.8.3</MicrosoftBuildVersion>
-    <MicrosoftBuildTasksCoreVersion>17.8.3</MicrosoftBuildTasksCoreVersion>
-    <MicrosoftIORedistVersion>6.0.1</MicrosoftIORedistVersion>
+    <MicrosoftBuildVersion>17.11.31</MicrosoftBuildVersion>
```
rainersigwald, line 15: As a standalone package sourcelink might still want to target down to 17.8? Not sure.

ViktorHofer: The msbuild packages are consumed at compile time only in sourcelink, not at run time. It shouldn't matter which version is referenced given that msbuild started pinning the assembly version years years ago. That's at least my understanding, please correct me if I'm wrong here.

src/arcade/eng/Versions.props
ViktorHofer: Right, Arcade 10 requires a .NET 10 SDK and therefore a newer MSBuild and VS version (probably 17.12?). The assemblies also all target net10.0. And last but not least, 17.8.29 has vulnerable dependencies (STJ/7.0.3).

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Microsoft.DotNet.UnifiedBuild.Tasks.csproj
ViktorHofer: Good one to look at in a follow-up

src/sourcelink/eng/Versions.props
rainersigwald: If you reference a higher package version you can reference APIs that don't exist at runtime when running on a lower MSBuild. This is not likely to be a problem for this version range and usage though, so almost certainly not a big deal.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Microsoft.DotNet.UnifiedBuild.Tasks.csproj
rainersigwald: Yeah absolutely not blocking!

ViktorHofer:
@dotnet/source-build can you please help with the SB test failures? TIA

MichaelSimons:
> @dotnet/source-build can you please help with the SB test failures? TIA

Yes - @mthalman is already investigating.

mthalman:
The SB leg is getting failures in the test step like the following:

```
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Microsoft.DotNet.UnifiedBuild.Tasks.csproj error NU1102: Unable to find package Microsoft.Build.Tasks.Core with version (>= 17.15.0-preview-25230-108)
  - Found 67 version(s) in dotnet-public [ Nearest version: 17.14.8 ]
  - Found 1 version(s) in dotnet-eng [ Nearest version: 16.5.0-preview-19606-01 ]
  - Found 0 version(s) in dotnet-libraries [/__w/1/s/.dotnet/sdk/10.0.100-preview.5.25230.108/NuGet.targets]
```

This is because it's picking up the PSB version but there's no PSB feed defined.

eng/tools/Directory.Build.props
```diff
@@ -8,4 +8,11 @@
     <RestoreSources Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(IsTestRun)' != 'true'">$(PrebuiltPackagesPath);$(PreviouslySourceBuiltPackagesPath)</RestoreSources>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
```
MichaelSimons: @mthalman - this prevents the PSB artifacts from being used?  Viktor and I were trying to use PSB to reduce maintenance cost.  w/o this, these versions would need to be added to SBRP and then rebootstrap to pick them up.

mthalman: Then the problem is this:

https://github.com/dotnet/dotnet/blob/3a4520155cd43fe9bf469961e91159822ba67405/eng/tools/Directory.Build.props#L5-L9

That specifically prevents PSB feed from being used for the tests. That was done here: https://github.com/dotnet/sdk/pull/46442

ViktorHofer: PSB is only needed during the build and must not be used during testing. But the PSB Versions import was happening unconditionally. I fixed that with the latest commit and the failures are gone now.

ViktorHofer:
Just to summarize, this change also discovered two regression that are getting fixed:
- The path to the PSB versions file was incorrect due to recent publishing refactoring of properties
- The PSB versions import wasn't conditioned out when running tests while the PSB feed was.

PR #547, opened 2025-05-14
Fix UpdateEngCommonFiles target not applying to SBRP

ViktorHofer:
Otherwise SBRP builds with bootstrap Arcade but with _a_ version of eng/common (the one that got last synced) which makes it really hard to make changes to eng/common in the VMR.

### DIFF
--- a/repo-projects/Directory.Build.targets
+++ b/repo-projects/Directory.Build.targets
@@ -265,10 +265,11 @@
     </Touch>
   </Target>
 
-  <!-- TODO: Remove when all repos use a consistent set of eng/common files: https://github.com/dotnet/source-build/issues/3710. -->
-  <Target Name="UpdateEngCommonFiles" Condition="'$(UseBootstrapArcade)' != 'true'">
+  <!-- Don't update arcade's eng/common layout as that is considered live even though it uses bootstrap Arcade. -->
+  <Target Name="UpdateEngCommonFiles" Condition="'$(RepositoryName)' != 'arcade'">
     <ItemGroup>
-      <InputEngCommonFile Include="$(RepoRoot)src\arcade\eng\common\**\*" />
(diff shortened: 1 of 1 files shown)


NikolaMilosavljevic (approved):

repo-projects/Directory.Build.targets
```diff
@@ -265,10 +265,11 @@
     </Touch>
   </Target>
 
-  <!-- TODO: Remove when all repos use a consistent set of eng/common files: https://github.com/dotnet/source-build/issues/3710. -->
-  <Target Name="UpdateEngCommonFiles" Condition="'$(UseBootstrapArcade)' != 'true'">
+  <!-- Don't update arcade's eng/common layout as that is considered live even though it uses bootstrap Arcade. -->
+  <Target Name="UpdateEngCommonFiles" Condition="'$(RepositoryName)' != 'arcade'">
     <ItemGroup>
-      <InputEngCommonFile Include="$(RepoRoot)src\arcade\eng\common\**\*" />
+      <InputEngCommonFile Include="$(RepoRoot)eng\common\**\*" Condition="'$(UseBootstrapArcade)' == 'true'" />
```
MichaelSimons, line 271: This is really not the right version either.  The only correct way to solve this is to capture this within PSB and reference it from there.

ViktorHofer: The eng/common layout from the VMR root must be in sync with bootstrap Arcade. We need to guarantee that that's the case. I assume that means updating the re-bootstrap pipeline to fetch the right eng/common files (probably based on commit).

MichaelSimons (approved):

repo-projects/Directory.Build.targets
MichaelSimons: I logged https://github.com/dotnet/source-build/issues/5183 and https://github.com/dotnet/source-build/issues/5185.

ViktorHofer: Thanks a lot

PR #548, opened 2025-05-14
Reintroduce support for building VMR from repo PR/CI

NikolaMilosavljevic:
Contributes to: https://github.com/dotnet/arcade-services/issues/4539

This PR is basically bringing back support for VMR builds that are launched from individual repos, and not from VMR. It essentially reverts the removal of `isBuiltFromVmr` property and associated logic, that was done with: https://github.com/dotnet/dotnet/pull/226

Additionally there are few tweaks:
- We only use a single template now for vmr-sync steps: `vmr-pull-updates.yml`. The template is referenced using `@self` and resolved from repo sources.
- I've removed 2 usages of `@self`, in template references, as they were unnecessary

Vmr syncing template and scripts will live in repo sources, under `eng/common` and will be provided to all repos via `arcade` updates.

Full verification build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2708573&view=results There is a known failure in source-build poison tests, that exists in the baseline as well.


eng/common/darc-init.sh
```diff
@@ -68,7 +68,7 @@ function InstallDarcCli {
     fi
   fi
 
-  local arcadeServicesSource="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json"
+  local arcadeServicesSource="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json"
```
ViktorHofer: Changes to eng/common must not be made in the VMR. You probably want to make this change in src/arcade/eng/common/darc-init.sh (which probably is already done). That's the location that flows to repos in and outside of the VMR.

NikolaMilosavljevic: OK. I'll remove this. Yeah, the change is already in arcade.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/548/commits/0734954118e50dace0b818f383264f8dcd39ccd7

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -311,6 +322,13 @@ jobs:
         sbomEnabled: false
 
   steps:
+  - ${{ if not(parameters.isBuiltFromVmr) }}:
+    # Synchronize new content in the VMR during PRs (we expect this to come
```
MichaelSimons: This comment doesn't seem complete.

NikolaMilosavljevic: Yeah, it was the original comment I brought back along with the surrounding code. I'll remove the content in parenthesis as it doesn't add any value.

Here's the same comment in P4 branch: https://github.com/dotnet/dotnet/blob/c22dcd0c7a78d095a94d20e59ec0271b9924c82c/src/sdk/eng/pipelines/templates/jobs/vmr-build.yml#L339

NikolaMilosavljevic: Fixes with https://github.com/dotnet/dotnet/pull/548/commits/9981a4f32fd0601604d18d166fa325a2bfaba45a

NikolaMilosavljevic:
PR checks are hitting known NuGet Audit errors - will likely need to wait until https://github.com/dotnet/dotnet/pull/546 is merged before rerunning checks.

mthalman (approved):

PR #549, opened 2025-05-14
Remove source-build pipelines

ellahathaway:
Removes the references to SB and SB-lite pipelines

[Internal pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2709183&view=results)


eng/pipelines/templates/stages/source-build-stages.yml
```diff
@@ -59,8 +63,8 @@ stages:
         reuseBuildArtifactsFrom: '{0}_Online_MsftSdk_x64'
         officialBuildId: ${{ parameters.officialBuildId }}
       
-      ### Additional legs for lite/full builds ###
-      - ${{ if in(parameters.scope, 'lite', 'full') }}:
+      ### Additional legs for full build ###
+      - ${{ if in(parameters.scope, 'full') }}:
```
mthalman, line 73: I'm confused. Based on the indentation further below, it looks like most of the jobs will always be built in either lite or full mode. And then Ubuntu_Offline_MsftSdk and CentOS_Offline_PreviousSourceBuiltSdk will be included in full mode. That doesn't seem right unless I'm parsing this wrong.

ellahathaway: We're getting rid of the lite pipeline, so I moved the lite legs to be included in the full build. This way, lite only applies to PR validation.

In other words, "ultralite" became our "lite".

Lite legs:
image

Full legs:
image

ViktorHofer (approved):

mthalman (commented):
There's an issue in the full pipeline where SB legs aren't using the fake official build ID.

ellahathaway:
> There's an issue in the full pipeline where SB legs aren't using the fake official build ID.

Fixed in https://github.com/dotnet/dotnet/pull/549/commits/7bdf3403138e0af5e5284cbe9c9b57767c1ab495

mthalman (approved):

ellahathaway:
Filed https://github.com/dotnet/aspnetcore/issues/61954 for the UB failures, which are unrelated to my changes. Once all of the source-build legs are passing (waiting on FR to fix a pool issue with the ubuntu arm64 leg), this PR will have to be force merged.

ViktorHofer:
Looks like merging is possible because the failures are all in the optional pipeline :)

PR #550, opened 2025-05-14
Fix malformed NuGet.config file

NikolaMilosavljevic:
Fixes: https://github.com/dotnet/source-build/issues/5122

`NuGet.config` file in `msbuild` repo is malformed. `packageSourceMapping` element is incorrectly placed under `packageSources` element. Package source mappings in this, and the file modified by our tasks, are never used by Nuget, causing intermittent failures with `NETStandard.Library` being picked up from `previously-source-built` source, instead of `reference` source.

`packageSourceMapping` section was originally added with https://github.com/dotnet/msbuild/pull/11255

Fixing in VMR instead of `msbuild` repo, to get this fixed asap.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -45,6 +45,9 @@
 
   <!-- Manually import the Versions.props file when the Arcade SDK isn't used. -->
   <Import Project="$(RepositoryEngineeringDir)Versions.props" Condition="'$(SkipArcadeSdkImport)' == 'true'" />
+  <Import Project="$(RepositoryEngineeringDir)VmrLayout.props" />
+  <!-- For source only builds non test builds, override Versions.props entries with previously sourcebuilt artifacts. -->
+  <Import Project="$(PreviouslySourceBuiltPackageVersionsPropsFile)" Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(IsTestRun)' != 'true' and Exists('$(PreviouslySourceBuiltPackageVersionsPropsFile)')" />
 
   <PropertyGroup>
     <!-- TODO: Remove when Arcade updated NetCurrent to net10.0. -->
@@ -73,5 +76,4 @@
     <ShellExtension Condition="'$(BuildOS)' != 'windows'">.sh</ShellExtension>
   </PropertyGroup>
 
-  <Import Project="$(RepositoryEngineeringDir)VmrLayout.props" />
 </Project>
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -2,13 +2,13 @@
   <ProductDependencies>
   </ProductDependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.25221.2">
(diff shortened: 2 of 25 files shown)


MichaelSimons:
> Fixing in VMR instead of msbuild repo, to get this fixed asap.

Please loop in the appropriate msbuild folks to review.

src/msbuild/NuGet.config
```diff
@@ -15,13 +15,13 @@
...
     <add key="vs-impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vs-impl/nuget/v3/index.json" />
-    <!-- We don't want other than VS OpenTelemetry libraries from vs-impl -->
-    <packageSourceMapping>
-      <packageSource key="vs-impl">
-        <package pattern="Microsoft.VisualStudio.OpenTelemetry*" />
-      </packageSource>
-    </packageSourceMapping>
   </packageSources>
+  <!-- We don't want other than VS OpenTelemetry libraries from vs-impl -->
+  <packageSourceMapping>
```
MichaelSimons, line 20: Isn't this going to fail the build because there are no wild cards for the other packages?

NikolaMilosavljevic: Good point - I will add `*` mappings for all other sources.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/550/commits/d55a2e0d17c2b56243c60f2a23f389e4ecdb51fe

ViktorHofer (approved):

MichaelSimons (approved):

NikolaMilosavljevic:
PR checks are hitting known NuGet Audit errors - will likely need to wait until https://github.com/dotnet/dotnet/pull/546 is merged before rerunning checks.

NikolaMilosavljevic:
@rainersigwald @baronfel it would be great to have a review from someone on the `msbuild` repo side.

baronfel (approved):
Wow, the fact that this isn't a validation error of some kind is kind of huge. Good find.

rainersigwald (approved):

baronfel:
Logged a bug on the NuGet side asking for some more validation in NuGet.config to prevent these kinds of errors.

JanProvaznik:
Does the dotnet/dotnet -> dotnet/msbuild flow work already? or should I copy the commit to MSBuild manually?

ViktorHofer:
Already works. Should come in tomorrow morning with the backflow subscription: dotnet -> msbuild (main)

rainersigwald:
It looks like this is breaking our official build on toolset restore; am I correct that @NikolaMilosavljevic's https://github.com/dotnet/dotnet/pull/554 would let us revert this change (and then fix forward shortly)?

mthalman:
Yes, I believe so.

mthalman:
@NikolaMilosavljevic is out this week, btw

MichaelSimons:
> It looks like this is breaking our official build on toolset restore; am I correct that @NikolaMilosavljevic's https://github.com/dotnet/dotnet/pull/554 would let us revert this change (and then fix forward shortly)?

Confirming @mthalman's response of _Yes, I believe so,_

PR #553, opened 2025-05-14
Make Scan for binaries a required VMR PR validation check

ellahathaway:
Closes https://github.com/dotnet/source-build/issues/5040

Relocated the binary scanning from the `Tag & Scan` stage to a new unit test called `ScanForBinaries`.

Some additional changes in this PR:
- Moved ExecuteHelper from source build and unified build tests to the shared TestUtilities class. This was done to prevent code duplication across test classes.
- Prevented the execute helper from removing PATH environment variable. This was causing the binary scan to fail as it was removing path to the Linux utility functions.
- Added the option to include an additional output comment when a process fails with an unexpected exit code.


ViktorHofer (commented):
Adding a new binary into the VMR sources wasn't failing the build / the scan step before, right? Would it fail the build (test step) now?

test/Microsoft.DotNet.Tests/Config.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+using System.IO;
+using System.Runtime.Versioning;
+
+namespace Microsoft.DotNet.Tests;
+
+internal static class Config
+{
+    const string ConfigSwitchPrefix = "Microsoft.DotNet.Tests.";
```
mthalman:
```suggestion
    private const string ConfigSwitchPrefix = "Microsoft.DotNet.Tests.";
```

MichaelSimons:
> Adding a new binary into the VMR sources wasn't failing the build / the scan step before, right? Would it fail the build (test step) now?

New binaries would not fail the build before and always required a follow-up checkin to fix.  The reason this was not enforced before is because of the indirect flow into the VMR.  If a repo transitively flowed into the SDK, we didn't want to block the flow on new binaries.  With direct flows, this is something we would like to enforce up front.

ViktorHofer:
> With direct flows, this is something we would like to enforce up front.

Got it. The plan is to move the test execution out of the VMR vertical builds. At that point, a new binary would fail the test but not fail the build (similar to how the source-build tests are now running after build but not failing the build).

ViktorHofer:
Looking at this change makes me wonder if scanning for binaries after the build is the right thing to do. There are repos that write into the working directory as part of the build.

ellahathaway:
> Looking at this change makes me wonder if scanning for binaries after the build is the right thing to do. There are repos that write into the working directory as part of the build.

Unfortunately, this is a drawback to running this test immediately post-build. I'm assuming that when we move the test execution out of the VMR vertical builds that this will no longer be an issue because we'll be running the tests on a clean clone of the repo.

mthalman (approved):

MichaelSimons:
> Unfortunately, this is a drawback to running this test immediately post-build. I'm assuming that when we move the test execution out of the VMR vertical builds that this will no longer be an issue because we'll be running the tests on a clean clone of the repo.

Even if we more the test execution out of the vertical builds, there is possibility that tests can mutate the working directory.

Should this be treated more like how we do poisoning?  Should we scan for binaries before the build.  Write the results to a report (log file), and then the test "analyze" the report and fails accordingly.

ellahathaway:
> > Unfortunately, this is a drawback to running this test immediately post-build. I'm assuming that when we move the test execution out of the VMR vertical builds that this will no longer be an issue because we'll be running the tests on a clean clone of the repo.
> 
> Even if we more the test execution out of the vertical builds, there is possibility that tests can mutate the working directory.
> 
> Should this be treated more like how we do poisoning? Should we scan for binaries before the build. Write the results to a report (log file), and then the test "analyze" the report and fails accordingly.

I okay with doing either approach.

So far, the mutations to the working directory that affect the binary scanning are minimal. If we think that more binaries are going to continuously pop up due to build mutations, then I think it'd be appropriate to scan before we build and compare the baselines.

Edit: We discussed this offline, and we decided to run the binary detection before building and then check the baseline after building. https://github.com/dotnet/dotnet/pull/553/commits/f26a634a5cb22a0030ac085ea1f9f33adf2ace6b implements this logic.

eng/init-detect-binaries.proj
```diff
@@ -0,0 +1,24 @@
...
+    
+    <PropertyGroup>
+      <DetectBinariesLogLevel Condition="'$(DetectBinariesLogLevel)' == ''">None</DetectBinariesLogLevel>
+      <DetectBinariesAdditionalArgs>--with-sdk $(DotNetRoot)</DetectBinariesAdditionalArgs>
+      <DetectBinariesAdditionalArgs Condition="'$(DotNetBuildSourceOnly)' == 'true'">$(DetectBinariesAdditionalArgs) --with-packages $(PreviouslySourceBuiltPackagesPath)</DetectBinariesAdditionalArgs>
+    </PropertyGroup>
+
+    <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Detecting binaries." />
+
+    <Exec Command="$(RepositoryEngineeringDir)/detect-binaries.sh --output-dir &quot;$(BinariesReportDir)&quot; --log-level $(DetectBinariesLogLevel) $(DetectBinariesAdditionalArgs)"
```
MichaelSimons, line 18: With this integrated into the build, Is there still a need for a stand-alone script?  Eliminating the script and integrating it into this target would simplify some things as well as make it faster.  Additionally it would DetectPlatform cross platform.

MichaelSimons, line 18: We should be capturing a binlog for this in case anything goes wrong.  This would be solved by my previous comment.

ellahathaway: We wrote the script so that the removal of binaries could be easily done by distro-maintainers. If we no longer see a need for that support, I'm fine with removing the script & the removal of binaries in prep.

ellahathaway:
I forgot that PR was on auto-merge, hence why it was merged yesterday at EOD before my new changes could be reviewed. I will apply additional feedback in a follow-up PR.

eng/init-detect-binaries.proj
ellahathaway: Filed https://github.com/dotnet/source-build/issues/5203 to continue this discussion there

PR #575, opened 2025-05-15
Fixes required for stable preview builds + some improvements

mmitche:
Sets of fixes required for creating stable preview builds.
- Use the ref pack versions in place of the x64 runtime version
- Use the platform version for places that should use non-stable versions
- Fix some places that were using the wrong version type
- Remove all places that were using arch specific versions
- Remove all the special property setups for the arch specific versions.


src/runtime/src/libraries/Microsoft.Extensions.Configuration.UserSecrets/tests/Microsoft.Extensions.Configuration.UserSecrets.Tests.csproj
```diff
@@ -15,8 +15,8 @@
       <_Parameter2>$(TargetFramework)</_Parameter2>
     </AssemblyAttribute>
     <AssemblyAttribute Include="System.Reflection.AssemblyMetadataAttribute">
-      <_Parameter1>MicrosoftNETCoreAppRuntimeVersion</_Parameter1>
-      <_Parameter2>$(MicrosoftNETCoreAppRuntimeVersion)</_Parameter2>
+      <_Parameter1>MicrosoftNETCoreAppRefVersion</_Parameter1>
```
ericstj: Why did you need to change runtime? Seems this will break tests https://github.com/dotnet/runtime/blob/91195a7948a16c769ccaf7fd8ca84b1d210f6841/src/libraries/Microsoft.Extensions.Configuration.UserSecrets/tests/MsBuildTargetTest.cs#L54

mmitche: That change should be reverted.

src/diagnostics/eng/Versions.props
```diff
@@ -114,8 +114,8 @@
   </PropertyGroup>
   <ItemGroup Condition="!$(PrivateBuildTesting) AND !$(InternalReleaseTesting)">
     <RuntimeTestVersions Include="Latest">
-      <RuntimeDownload>$(VSRedistCommonNetCoreSharedFrameworkx64100Version)</RuntimeDownload>
-      <Runtime>$(MicrosoftNETCoreAppRuntimewinx64Version)</Runtime>
+      <RuntimeDownload>$(MicrosoftNETCorePlatformsVersion)</RuntimeDownload>
```
ericstj: Why the platforms package? Just arbitrary?  Will It's version be different than ref pack version?

jkoritzinsky:
Do you have a validation run of the full unified build pipeline (with later build passes)?

src/diagnostics/eng/Versions.props
jkoritzinsky: The platforms package is a non-shipping package (so non-stable version even in stable version scenarios) that's built on all verticals.

jkoritzinsky:
Also, I think it would be useful to have a parameter to the official pipeline to select building with stabilized versions so we can easily validate (in the same way we can turn signing and ibc on and off).

src/diagnostics/eng/Versions.props
mmitche: It doesn't stabilize, and is a package that I don't expect to go away, become stable, etc. in the future. We could add a formal sentinel package to runtime, but I think it's unnecessary.

mmitche:
> Also, I think it would be useful to have a parameter to the official pipeline to select building with stabilized versions so we can easily validate (in the same way we can turn signing and ibc on and off).

I think the way we'll probably ship is with the "stable" switch checked in. But a parameter that sets the value if not specific seems like a good idea.

> Do you have a validation run of the full unified build pipeline (with later build passes)?

Good point. I'll get this done after I resolve the SB issue.

ViktorHofer (changes requested):
Versions.props changes don't sync back so blocking this until the individual repos changed flowed in.

mmitche:
> Versions.props changes don't sync back so blocking this until the individual repos changes flowed in.

Fine by me. I'll iterate on the stable switch and source build errors.

ViktorHofer:
Looks like the windowsdesktop, winforms and wpf ones are already in. Do you plan to submit a PR to dotnet/sdk for the sdk ones?

mmitche:
SDK: https://github.com/dotnet/sdk/pull/49017

mmitche:
Diagnostics: https://github.com/dotnet/diagnostics/pull/5486

ViktorHofer (approved):

mmitche:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2715341&view=results

mmitche:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2717926&view=results

mmitche:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2718058&view=results

mmitche:
Rebased this back onto main

build.sh
```diff
@@ -122,6 +123,20 @@ while [[ $# > 0 ]]; do
       properties+=( "/p:TargetArchitecture=$2" )
       shift
       ;;
+    -branding)
+      if [[ "$2" == "preview" ]]; then
+        properties+=( "/p:DotNetFinalVersionKind=prerelease" )
+      elif [[ "$2" == "rtm" ]]; then
+        properties+=( "/p:DotNetFinalVersionKind=release" )
```
ViktorHofer, line 130: Do we need this rtm vs release difference? IOW can rtm be renamed to release?

mmitche: I don't like the naming we have in arcade for prerelease vs. release tbh. It doesn't really same much about branding. since pre-release branding is ambiguous between a build with a date suffix and without. Changing the Arcade logic is 
a pretty impactful change.
 
I'm not totally tied to `preview/rtm` vs. `prerelease/release`. Thoughts from the distro maintainers SB folks? @tmds @omajid @MichaelSimons

tmds: I don't have a preference for the names of these labels.

For our daily CI we follow what the official daily builds from Microsoft are producing in order to have something that behaves like a "daily SDK".

ViktorHofer:
I think it would be good to extract the pgo clean-up work into a separate PR.

mmitche:
> I think it would be good to extract the pgo clean-up work into a separate PR.

https://github.com/dotnet/dotnet/pull/1134

jkoritzinsky (commented):
runtime changes LGTM

mmitche:
https://github.com/dotnet/sdk/pull/49463

mmitche:
Waiting on any last feedback here: https://github.com/dotnet/fsharp/pull/18701

mmitche:
FSharp is in. Now to get that forward flow and then merge in here. Then this PR should be good to go.

ViktorHofer (commented):
Does command-line-api.proj need `AllowDotNetFinalVersionKindOverride` as well?

repo-projects/windowsdesktop.proj
```diff
@@ -1,5 +1,9 @@
 <Project Sdk="Microsoft.Build.NoTargets">
 
+  <PropertyGroup>
+    <!-- May stabilize. -->
+    <AllowDotNetFinalVersionKindOverride>true</AllowDotNetFinalVersionKindOverride>
+  </PropertyGroup>
```
ViktorHofer, line 6: Just in case you do another push to this PR:

```suggestion
  </PropertyGroup>

```

PR #592, opened 2025-05-16
Expose -fromVMR CLI switch & rename DotNetBuildOrchestrator -> DotNetBuildFromVMR

ViktorHofer:
See commit messages for details and reasoning. Commit 1 exposes the new CLI args and commit 2 renames the msbuild property.


ViktorHofer:
I'm not sold on this approach. I still wonder if an environment variable would make more sense

akoeplinger:
I still think approximately 0 people outside the VMR team know what the "orchestrator" is, I'd prefer getting rid of that term, DotNetBuild + DotNetBuildRepo should be enough :)

ViktorHofer:
`DotNetBuildRepo` doesn't exist anymore. That got replaced by `DotNetBuild`. Repositories that condition on product build use that switch. The `Repo` suffix was necessary back when there was an inner and outer build that we had to differentiate.

---

`DotNetBuildOrchestrator` (now `DotNetBuildFromOrchestrator`) signals that the product build is executed within the VMR orchestrator. That's necessary to know in a few cases in both msbuild and scripts.

We have the following scenarios:

| Name | Build type | Command | Passed-in msbulid properties |
| --- | --- | --- | --- |
| Repo build | | build.sh |
| Repo source-build | implies product build | build.sh --source-build | /p:DotNetBuild=true /p:DotNetBuildSourceOnly=true |
| Repo product-build | | build.sh --product-build | /p:DotNetBuild=true |
| VMR source-build | implies product build | build.sh --source-build --from-orchestrator | /p:DotNetBuild=true /p:DotNetBuildSourceOnly=true /p:DotNetBuildFromOrchestrator=true |
| VMR product-build | | build.sh --product-build --from-orchestrator | /p:DotNetBuild=true /p:DotNetBuildFromOrchestrator=true |

> I still think approximately 0 people outside the VMR team know what the "orchestrator" is, I'd prefer getting rid of that term, DotNetBuild + DotNetBuildRepo should be enough :)

With those given scenarios it would be very confusing if `DotNetBuild` would imply a VMR build but `DotNetBuildRepo` (read: `DotNetBuild` + `Repo` suffix) implying a VMR or repo product build. That's why we chose the opposite: `DotNetBuild` meaning product build. And if that build runs inside the VMR, the `DotNetBuildFromOrchestrator` property signals that.

ViktorHofer:
These input controls should be as clear as possible for individual repos. In ~90% of cases you want to use `DotNetBuild` to affect  the product build code path, regardless of inside our outside the VMR. In those 10% where it matters you would use `DotNetBuildFromOrchestrator`.

If you dislike the orchestrator term, we can for sure rename that to VMR, i.e. `DotNetBuildFromVMR`. I'm not sure if that's better.

akoeplinger:
> If you dislike the orchestrator term, we can for sure rename that to VMR, i.e. DotNetBuildFromVMR. I'm not sure if that's better.

I think that'd be much better yeah

ViktorHofer:
I renamed the arguments to what we decided on in yesterday's VMR sync.

akoeplinger (approved):

PR #593, opened 2025-05-16
Provide a way to skip binary detection

mthalman:
For whatever reason, binary detection runs really slow in my dev container (7 minutes). I see it runs quickly in the pipelines though. But I'd like a way to skip it.

### DIFF
--- a/eng/init-detect-binaries.proj
+++ b/eng/init-detect-binaries.proj
@@ -5,7 +5,8 @@
   </PropertyGroup>
 
   <Target Name="DetectBinaries"
-          BeforeTargets="Build">
+          BeforeTargets="Build"
+          Condition="'$(SkipDetectBinaries)' != 'true'">
     
     <PropertyGroup>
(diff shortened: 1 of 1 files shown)


ellahathaway (approved):
Do you have a lot of artifacts in the build eg .packages, nested artifact folders, etc? That would slow down the detection for sure.

mthalman:
> Do you have a lot of artifacts in your dotnet directory eg .packages, nested artifact folders, etc? That would slow down the detection for sure.

No, I did a clean of the repo before building.

ellahathaway:
> > Do you have a lot of artifacts in your dotnet directory eg .packages, nested artifact folders, etc? That would slow down the detection for sure.
> 
> No, I did a clean of the repo before building.

Weird. It runs quickly in my local build. If you give me the details of your build environment, I'll be happy to investigate. Also, if this becomes a persistent issue, I'll make some adjustments to the tool to increase the parallelism.

mmitche (approved):

PR #596, opened 2025-05-16
Change AnalyzerUtilities to reference Microsoft.CodeAnalysis.Common by project instead of package

mthalman:
See https://github.com/dotnet/dotnet/pull/400#issuecomment-2886950873

Changes Microsoft.CodeAnalysis.AnalyzerUtilities reference to Microsoft.CodeAnalysis.Common to be a project instead of package reference. This ensures it has a dependency on the current version rather than the previously published version.

This is needed to get a VMR build with a good previously-source-built artifacts tarball that is coherent which will unblock https://github.com/dotnet/dotnet/pull/400.


mthalman:
Changing this to draft as I work through issues uncovered by the roslyn PR: https://github.com/dotnet/roslyn/pull/78614

PR #619, opened 2025-05-19
Remove git-info files

ViktorHofer:
These were still necessary for source-build which now also uses unified versioning.

Contributes to https://github.com/dotnet/source-build/issues/4491


ViktorHofer:
@dkurepa @premun can you please update the sync tool to stop producing the git-info files? Can you please also stop syncing the `packageVersion` into https://github.com/dotnet/dotnet/blob/main/src/source-manifest.json? That's not need anymore either.

premun:
I opened https://github.com/dotnet/arcade-services/issues/4858

premun:
@ViktorHofer we can go ahead with this now

premun:
But maybe we should first merge all FF PRs?

akoeplinger:
I think we still need packageVersion in the .json right? https://github.com/dotnet/arcade-services/issues/2428 wasn't done yet

premun:
Yeah, we still need `packageVersion` and it will keep coming back

PR #652, opened 2025-05-20
Update allowed-sb-binaries.txt to exclude MicroBuild

jkoritzinsky:
Working around dotnet/arcade#15731

Fixes https://github.com/dotnet/dotnet/issues/650

### DIFF
--- a/eng/allowed-sb-binaries.txt
+++ b/eng/allowed-sb-binaries.txt
@@ -9,6 +9,7 @@ artifacts/
 .git/
 .packages/
 prereqs/packages/
+MicroBuild/
 
 **/*.bmp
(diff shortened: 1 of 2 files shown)


eng/allowed-sb-binaries.txt
```diff
@@ -9,6 +9,7 @@ artifacts/
 .git/
 .packages/
 prereqs/packages/
+MicroBuild/
```
ellahathaway: Thinking about this in detail, I actually think that this exclusion better belongs in `allowed-vmr-binaries.txt` since we don't sign source-only builds.

ellahathaway (approved):

PR #677, opened 2025-05-21
Support having spaces in OfficialBuilder

omajid:
Currently, running the following command fails:

    $ ./build.sh --prep --source-build /p:OfficialBuilder='Red Hat'

Lots of scripts and projects assume that there are no spaces in OfficialBuilder (and other properties). Having spaces breaks in all sorts of places. This is a minimal fix to try and resolve that.

The primary change is to use bash arrays to store values (properties and arguments) to make sure they are quoted correctly when passed to other places.

### DIFF
--- a/repo-projects/Directory.Build.props
+++ b/repo-projects/Directory.Build.props
@@ -113,7 +113,7 @@
     <CommonArgs Condition="'$(CrossBuild)' == 'true'">$(CommonArgs) /p:CrossBuild=true</CommonArgs>
     <CommonArgs Condition="'$(DotNetBuildUseMonoRuntime)' == 'true'">$(CommonArgs) /p:DotNetBuildUseMonoRuntime=$(DotNetBuildUseMonoRuntime)</CommonArgs>
     <CommonArgs Condition="'$(OfficialBuildId)' != ''">$(CommonArgs) /p:OfficialBuildId=$(OfficialBuildId)</CommonArgs>
-    <CommonArgs Condition="'$(OfficialBuilder)' != ''">$(CommonArgs) /p:OfficialBuilder=$(OfficialBuilder)</CommonArgs>
+    <CommonArgs Condition="'$(OfficialBuilder)' != ''">$(CommonArgs) /p:OfficialBuilder="$(OfficialBuilder)"</CommonArgs>
     <CommonArgs Condition="'$(ForceDryRunSigning)' != ''">$(CommonArgs) /p:ForceDryRunSigning=$(ForceDryRunSigning)</CommonArgs>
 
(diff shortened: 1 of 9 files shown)


omajid:
I am not sure what the code flow looks like these days. Please let me know if I should open PRs against the individual component repos instead.

mthalman:
The change to build.sh should only be made in the arcade repo. This is because it's within the eng/common directory. Edits within that directory should only be made in arcade and dependency flow will flow those changes to all the other repos, including the VMR.

mthalman:
> The change to build.sh should only be made in the arcade repo. 

By arcade repo, I mean https://github.com/dotnet/arcade, not the src/arcade directory in the VMR.

ViktorHofer:
Sorry, @mthalman that I have to disagree here. We have a different policy:

- Changes to Versions.props & Version.Details.xml must not be made inside the VMR (unless you really know what you're doing -> they don't sync back to the repo)
- Changes to eng/common files must be made under src/arcade/eng/common only. The VMR repos build with those build scripts (except for the VMR orchestrator & SBRP which use the root eng/common ones).
- Updating the .NET SDK or the Arcade.Sdk in VMR root's or src/arcade's global.json or updating the root eng/common files must only be done as part of a re-bootstrap PR (we have automation for that).

Based on that, @omajid please revert all the eng/common changes except for the one under src/arcade.

... and yes, we should document this. Also see https://github.com/dotnet/dotnet?tab=readme-ov-file#where-to-make-changes

ellahathaway:
> * Changes to eng/common files must be made under src/arcade/eng/common only. The VMR repos build with those build scripts (except for SBRP which uses the root eng/common ones).
> * Updating the .NET SDK or the Arcade.Sdk in VMR root's or src/arcade's global.json or updating the root eng/common files must only be done as part of a re-bootstrap PR (we have automation for that).

FYI - https://github.com/dotnet/source-build/issues/5185 will help prevent changes to eng/common.

omajid:
Thanks, @mthalman and @ViktorHofer ! I have reverted all eng/common changes except to src/arcade/eng/common. However, that means that building with `/p:OfficialBuilder='Red Hat'`  breaks when building SBRP, since it uses the root eng/common without the fixes. I guess I have to wait for the re-bootstrap?

mthalman:
> I guess I have to wait for the re-bootstrap?

That's correct.

mthalman (approved):

ViktorHofer (approved):

ViktorHofer:
Yes. If we have a passing build tonight, I can do another re-bootstrap tomorrow. I want one for other reasons as well.

PR #678, opened 2025-05-21
Use non-XL pool for tall stacks

mmitche:
We don't actually need the XL pool for this, and the XL pool only has a few machines.


jkoritzinsky (approved):

ViktorHofer (approved):
Interesting. Why don't we need an XL pool for internal?

mmitche:
> Interesting. Why don't we need an XL pool for internal?

The internal pools already have 8 cores by default.

PR #719, opened 2025-05-22
Manually apply of "Remove MicroBuild workaround"

ViktorHofer:
Manual cherry-pick of https://github.com/dotnet/arcade/commit/12d3a9f5d6138e22270694574e73e4c58a815795

This fixes the broken official builds. We normally don't touch the toolset eng/common layout but this is one of the cases when we should.


ViktorHofer:
Force merge to unblock builds. And this change isn't validated here.

PR #726, opened 2025-05-22
Do not set RESTORENOHTTPCACHE

mmitche:


mmitche:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2714441&view=results

mmitche:
Doesn't solve our issue, but doesn't show any additional problems.

ViktorHofer:
We usually don't make changes to the eng/common bootstrap layout but I think that's OK in this case. It doesn't really matter.

PR #727, opened 2025-05-22
Can't error on warning if there is no warning. Checkmate msbuild

mmitche:


rainersigwald (approved):
In some build systems this would still break . . . but I don't think this is one, fortunately.

PR #760, opened 2025-05-23
Add validation for changes to protected files

ellahathaway:
Closes https://github.com/dotnet/source-build/issues/5185


jkoritzinsky (commented):
Implementing this using a .NET task and project seems a little excessive.

We should be able to do this with one or two lines of bash (just running the `git diff` command looking at only specific subpaths or using `grep` on the output of `git diff`).

For small actions like these, I strongly lean towards basic scripts instead of wrapping everything in C# and MSBuild.

ellahathaway:
@dotnet/source-build - can I get a review on this please?

.github/workflows/protected-files-validation.yml
```diff
@@ -0,0 +1,81 @@
...
+      - eng/Version.Details.xml
+      - eng/Versions.props
+
+permissions:
+  pull-requests: read
+
+jobs:
+  check-version-details:
+    runs-on: ubuntu-22.04
+    if: ${{ !startsWith(github.event.pull_request.title, 'rebootstrap') && !startsWith(github.event.pull_request.title, 're-bootstrap') && !startsWith(github.event.pull_request.title, 'release') && !startsWith(github.event.pull_request.title, '.NET Source-Build ') }}
```
mthalman: I haven't tested this but am offering it as a suggested alternative that is more compact: 

```suggestion
    if: ${{ !contains(fromJSON('["rebootstrap", "re-bootstrap", "release", ".NET Source-Build "]', github.event.pull_request.title) }}
```

mthalman: Is it possible to have this condition around the `jobs` field instead of each `job`?

.github/workflows/protected-files-validation.yml
```diff
@@ -0,0 +1,81 @@
...
+        working-directory: ${{ needs.setup.outputs.repo-path }}
+        run: |
+          if git diff --name-only origin/${{ github.event.pull_request.base.ref }} | grep '^eng/Version.Details.xml$' >/dev/null; then
+            echo "Version.Details.xml has changed."
+            exit 1
+          else
+            echo "No changes in Version.Details.xml."
+          fi
+
+  check-versions-props:
```
mthalman: Is it necessary to have a separate job for each of these checks? Can't they all be run in a single job?

.github/workflows/protected-files-validation.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Fetch base branch from upstream
+        run: |
+          git fetch origin ${{ github.event.pull_request.base.ref }}
+
+      - name: Check for Version.Details.xml changes
```
mthalman: Is it necessary to explicitly check whether a file change occurred? Given that the workflow only runs on the path filters you've indicated (and it meets the PR title requirements), then you know it's in violation and can simply produce an error without having to check anything more.

.github/workflows/protected-files-validation.yml
ellahathaway: I'll give that a shot! I played around with a few different ways, and what I have is the best I could come up with. Hopefully your suggestion works :)

.github/workflows/protected-files-validation.yml
```diff
@@ -0,0 +1,81 @@
...
+          fetch-depth: 0
+
+      - name: Fetch base branch from upstream
+        run: |
+          git fetch origin ${{ github.event.pull_request.base.ref }}
+
+      - name: Check for Versions.props changes
+        working-directory: ${{ needs.setup.outputs.repo-path }}
+        run: |
+          if git diff --name-only origin/${{ github.event.pull_request.base.ref }} | grep '^eng/Versions.props$' >/dev/null; then
```
MichaelSimons: There are scenarios in which the Versions.props file can have valid edits outside of a rebootstrap/release PR.  e.g. you are working on the tasks and need to add a new dependency to the system. What is the expected workflow for these scenarios?  Is merge on red the only option?  Would it make sense to allow a "special" comment in the PR to indicate the edit is allowed by this check?

.github/workflows/protected-files-validation.yml
ellahathaway: My goal was to have it be clear which file path has the changes. But if we don't care about the specific path, then yes we can run this as a single job OR do what you suggested in https://github.com/dotnet/dotnet/pull/760#discussion_r2112460509

.github/workflows/protected-files-validation.yml
ellahathaway: I was picturing a "merge on red" scenario. This would require someone with elevated permissions (aka someone with knowledge on the file) to review and merge the PR.

That said, I'm open to having a "special" comment in the PR instead. Any ideas on what it should be?

.github/workflows/protected-files-validation.yml
ellahathaway: Echoing my reply in https://github.com/dotnet/dotnet/pull/760#discussion_r2112473188, I think that this is a UX issue. If we don't care about indicating which specific file has the changes, then yes, I think that the approach you're suggesting is appropriate. If we want to alert the user that a specific file cannot have changes, then we need to keep these specific diff checks.

.github/workflows/protected-files-validation.yml
mthalman: Yeah, as long as the error message states what is not allowed to be edited (the full list of disallowed paths), then I would say that is good enough. It doesn't need to figure out which specific file was in violation. So I think the suggestion at https://github.com/dotnet/dotnet/pull/760#discussion_r2112460509 would be adequate.

.github/workflows/protected-files-validation.yml
ViktorHofer: IMO merge on red is sufficient (requiring someone with elevated permissions to merge).

MichaelSimons: I am fine with merge on red as well.  I wanted to raise this to ensure an explicit decision was made.

MichaelSimons: > I am fine with merge on red as well. I wanted to raise this to ensure an explicit decision was made.

.github/workflows/protected-files-validation.yml
ellahathaway: I cannot find documentation to support that: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions

ellahathaway: Unfortunately, this syntax isn't case sensitive, and there's no good way to convert the PR title to a lower-case option. As a result, I'm going to keep the current if condition as is.

.github/workflows/protected-files-validation.yml
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/760/commits/9f8aca2189a7d5acc977428eae07bdf8ba5fe3e6

.github/workflows/protected-files-validation.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Check Protected Files for Changes'
+
+on:
+  pull_request:
+    paths:
```
mthalman, line 6: Add a comment here to update the error message below if these paths change.

.github/workflows/protected-files-validation.yml
```diff
@@ -0,0 +1,22 @@
...
+permissions:
+  pull-requests: read
+
+jobs:
+  check-protect-files:
+    runs-on: ubuntu-22.04
+    if: ${{ !startsWith(github.event.pull_request.title, 'rebootstrap') && !startsWith(github.event.pull_request.title, 're-bootstrap') && !startsWith(github.event.pull_request.title, 'release') && !startsWith(github.event.pull_request.title, '.NET Source-Build ') }}
+    steps:
+      - name: Protected File has Changes
+        working-directory: ${{ needs.setup.outputs.repo-path }}
```
mthalman: This `working-directory` shouldn't be necessary. Besides, it's referencing a job that doesn't exist.

.github/workflows/protected-files-validation.yml
mthalman: What about a regex?

```suggestion
    if: ${{ !contains(github.event.pull_request.title, '^(rebootstrap|re-bootstrap|release|\\.NET Source-Build )') }}
```

ellahathaway: GitHub Actions does not support regex matching in if conditional expressions

ellahathaway: I could move the condition into the step? That would make the expression better

premun:
@ellahathaway what about a PR like this one?

https://github.com/dotnet/dotnet/pull/832

It is perfectly fine, will it get through?

ellahathaway:
> @ellahathaway what about a PR like this one?
> 
> #832
> 
> It is perfectly fine, will it get through?

Good catch. The validation will fail in this case. I'll add this PR title to the list of allowed ones.

Edit: conditioned on the author instead, which addresses this case: https://github.com/dotnet/dotnet/pull/760/commits/038170226c07fff42c606acd76db0a1ea03178a4

.github/workflows/protected-files-validation.yml
mmitche: Does it make more sense to just check against the PR submitter? dotnet-sb-bot? Or does that catch cases that should not update the banned files

ellahathaway: That could work! We would have to include both `dotnet-sb-bot` and `dotnet-maestro`. SB bot should be fine, since that only touches these files in rebootstrap and release PRs. I think maestro should be fine too.

ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/760/commits/038170226c07fff42c606acd76db0a1ea03178a4

mmitche (approved):

ellahathaway:
@dotnet/product-construction - the failing PR checks are unrelated to the changes in this PR. Can someone please merge this PR for me?

PR #787, opened 2025-05-26
Fix from-vmr switch in nuget-client build.sh

ViktorHofer:
Fixes https://github.com/dotnet/source-build/issues/5218


ViktorHofer:
Only the SB legs build nuget-client and the online one already passed. That's enough validation.

PR #797, opened 2025-05-26
Fix bash array expansion on macOS

omajid:
Fix broken builds on macOS. macOS ships with an older version of bash, which doesn't expand empty arrays correctly. So check that arrays are not empty before expanding them.

### DIFF
--- a/src/arcade/eng/common/build.sh
+++ b/src/arcade/eng/common/build.sh
@@ -257,7 +257,7 @@ function Build {
     /p:Sign=$sign \
     /p:Publish=$publish \
     /p:RestoreStaticGraphEnableBinaryLogger=$binary_log \
-    "${properties[@]}"
+    ${properties[@]+"${properties[@]}"}
 
   ExitWithExitCode 0
 }
--- a/src/fsharp/eng/build.sh
+++ b/src/fsharp/eng/build.sh
@@ -295,7 +295,7 @@ function BuildSolution {
 
     BuildMessage="Error building tools"
     # TODO: Remove DotNetBuildRepo property when fsharp is on Arcade 10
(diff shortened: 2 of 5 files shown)


omajid:
See https://stackoverflow.com/questions/7577052/unbound-variable-error-in-bash-when-expanding-empty-array/7577209#7577209

omajid:
> Looks good, thanks for fixing. I don't fully understand the syntax. Would you mind explaining that?

(Did you mean documenting it in the code itself, like Copilot's feedback?)

The version of bash on macOS is so old that it follows a very strict interpretation of `"{array[@]}"` syntax: it tries to expand it to the individual elements of the array, then it errors out because there are no elements in the array, finally falling back to an even more legacy approach of treating `array[@]` as the actual variable name, and the variable with that name doesn't exist.

There are many solutions possible. They all boil down to "only expand the elements of the array if the array is not empty" to not break different versions of bash. https://stackoverflow.com/questions/7577052/unbound-variable-error-in-bash-when-expanding-empty-array/7577209 has a set of them. I picked the one that seems to be the most broadly compatible with most/all versions of bash based on the research there.

The fix uses a few pieces of syntax together.

- `"${array[@]}"` (with double quotes) expands to all elements of the array, each individual element quoted correctly.
- `${variable}` or `${array[@]}` (without double quoted) expands to the full value of the variable/array, without any quoting.
- `${variable+alternate}` expands to `alternate` if `variable` is set. If `variable` is unset or empty, it expands to nothing.

If you put it together, you get: `${array[@]+"${array[@]}"}`. Which says: expand array, and see if it is non-empty, if is not-empty, use the alternate value (`"${array[@]}"`) which is the elements of the array properly quoted.

There are other possible ways of expressing the same thing using different bash syntax, but as the StackOverflow post says, the majority of other ways run into issues with some version of bash.

For more details on the array and variable substitution syntax see https://web.archive.org/web/20230403055152/https://wiki.bash-hackers.org/syntax/arrays and https://web.archive.org/web/20230404084543/https://wiki.bash-hackers.org/syntax/pe#use_an_alternate_value

ViktorHofer:
Thanks a lot for providing all that context. Today I learned something :)

PR #812, opened 2025-05-27
Use P2Ps in VMR orchestrator for improved build performance

ViktorHofer:
- Define P2Ps to the msbuild task projects that are used by the repo-projects targets
- Remove the `test/Microsoft.DotNet.Installer.Tests/installer-tests-prereqs.proj` workaround now that we can build task projects via P2Ps from repo-projects.
- Simplify the short stack logic by inlining the logic into dotnet.proj
- Remove the now unnecessary tools.proj file
- Remove the incorrect use of the Microsoft.NET.Sdk in efcore.proj. Repo-projects should use the Microsoft.Build.NoTargets SDK.

---

Looks like this fixes the long standing issue that caused runtime to build later than expected:

```
  [13:52:27.29] DirSize After CleanupRepo arcade
 
  [13:52:37.76] Building cecil
  [13:55:12.79] DirSize After CleanupRepo cecil
 
  [13:52:35.96] Building command-line-api
  [13:58:26.72] DirSize After CleanupRepo command-line-api
 
  [13:58:26.69] Building runtime
```


ViktorHofer:
cc @rainersigwald @AR-May

repo-projects/Directory.Build.targets
```diff
@@ -55,6 +55,11 @@
     <MinimalConsoleLogOutput Condition="'$(MinimalConsoleLogOutput)' == ''">$(BuildInParallel)</MinimalConsoleLogOutput>
   </PropertyGroup>
 
+  <!-- Use the P2P protocol for RepositoryReference items. -->
+  <ItemGroup>
+    <ProjectReference Include="@(RepositoryReference->'%(Identity).proj')" IsRepositoryReference="true" BuildReference="false" />
+  </ItemGroup>
```
mthalman, line 61: Now that we have this, I wonder whether we can take advantage of MSBuild's traversal of the project references to replace or simplify the logic here:

https://github.com/dotnet/dotnet/blob/8efffbde026eccfca4eed0b453380dd91993c824/repo-projects/Directory.Build.targets#L58-L79

Does MSBuild provide a representation of the flattened list of dependencies?

ViktorHofer: Yes, I want us to eventually use the P2P graph for RepositoryReference as well (the building part) but I would rather do that in a follow-up as that has the potential to break join points.

mthalman (approved):

PR #823, opened 2025-05-27
[release/10.0.1xx-preview5] Enable v4 Publishing

jkoritzinsky:
Enable v4 publishing for UnifiedBuild jobs in the VMR.

Internal build https://dev.azure.com/dnceng/internal/_build/results?buildId=2720207&view=results


repo-projects/sdk.proj
```diff
@@ -18,13 +18,16 @@
...
+    <BaseAssetsPath Condition="'$(DotNetBuildPass)' != '' and '$(DotNetBuildPass)' != '1'">$(PreviousBuildPassAssetsDir)</BaseAssetsPath>
+    
+    <BuildArgs>$(BuildArgs) /p:PublicBaseURL=file:%2F%2F$(BaseAssetsPath)</BuildArgs>
     <BuildArgs>$(BuildArgs) /p:UsePortableLinuxSharedFramework=false</BuildArgs>
 
     <BuildArgs Condition="'$(PgoInstrument)' == 'true'">$(BuildArgs) /p:PgoInstrument=true</BuildArgs>
     <BuildArgs Condition="'$(EnablePoison)' == 'true' or '$(SkipUsingCrossgen)' == 'true'">$(BuildArgs) /p:SkipUsingCrossgen=true</BuildArgs>
 
-    <BuildArgs Condition="'$(BuildWorkloads)' == 'true'">$(BuildArgs) /p:BuildWorkloads=true</BuildArgs>
+    <BuildArgs Condition="'$(BuildWorkloads)' == 'true' or '$(DotNetBuildPass)' == '3'">$(BuildArgs) /p:BuildWorkloads=true /p:PreviousBuildPassShippingPackagesDir=$(PreviousBuildPassShippingPackagesDir)/$(RepositoryName)</BuildArgs>
```
ViktorHofer: This looks wrong? /p:BuildWorkloads=true should only be passed-in in BuildPass2

ViktorHofer:
If this is already ready for review, please provide a PR description so that we better understand the changes i.e. to the layout.

eng/common/core-templates/job/publish-build-assets.yml
ViktorHofer, line 1: eng/common change

eng/VmrLayout.props
```diff
@@ -28,6 +28,10 @@
     <ArtifactsAssetManifestsDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'manifests', '$(Configuration)'))</ArtifactsAssetManifestsDir>
     <AssetManifestsIntermediateDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsObjDir)', 'manifests', '$(Configuration)'))</AssetManifestsIntermediateDir>
     <ArtifactsAssetsDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'assets', '$(Configuration)'))</ArtifactsAssetsDir>
+    <PreviousBuildPassAssetManifestsDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'manifests'))</PreviousBuildPassAssetManifestsDir>
+    <PreviousBuildPassShippingPackagesDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'packages', 'shipping'))</PreviousBuildPassShippingPackagesDir>
+    <PreviousBuildPassNonShippingPackagesDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'packages', 'nonshipping'))</PreviousBuildPassNonShippingPackagesDir>
+    <PreviousBuildPassAssetsDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'assets'))</PreviousBuildPassAssetsDir>
```
ViktorHofer, line 36: What does previous build pass mean? Consider adding a comment.

ViktorHofer, line 36: Can we use the same casing and have the folders use a `$(Configuration)` like in the existing ones? Consider naming the root folders differently so that it's obvious what those folders are used for.

eng/Publishing.props
```diff
@@ -3,28 +3,30 @@
...
+    <PublishingVersion>3</PublishingVersion>
+    <PushToLocalStorage>true</PushToLocalStorage>
+    <AssetsLocalStorageDir>$(ArtifactsStagingDir)/artifacts/assets</AssetsLocalStorageDir>
+    <AssetManifestsLocalStorageDir>$(ArtifactsStagingDir)/artifacts/manifests</AssetManifestsLocalStorageDir>
+    <ShippingPackagesLocalStorageDir>$(ArtifactsTmpDir)/staging/packages/$(Configuration)/Shipping</ShippingPackagesLocalStorageDir>
+    <NonShippingPackagesLocalStorageDir>$(ArtifactsTmpDir)/staging/packages/$(Configuration)/NonShipping</NonShippingPackagesLocalStorageDir>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(PublishingVersion)' == '4'">
+    <FutureArtifactName Condition="'$(VerticalName)' != ''">$(VerticalName)_Artifacts</FutureArtifactName>
```
ViktorHofer, line 29: What does Future mean here?

eng/Publishing.props
```diff
@@ -3,28 +3,30 @@
...
     <AutoGenerateSymbolPackages>false</AutoGenerateSymbolPackages>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' != 'true'">
-    <SourceBuiltShippingPackagesDir>$(ArtifactsStagingDir)/packages/$(Configuration)/Shipping</SourceBuiltShippingPackagesDir>
-    <SourceBuiltNonShippingPackagesDir>$(ArtifactsStagingDir)/packages/$(Configuration)/NonShipping</SourceBuiltNonShippingPackagesDir>
-  </PropertyGroup>
   <!--
     When building from source, the Private.SourceBuilt.Artifacts archive will contain the nuget packages, so don't push them where we'll upload from.
+    As we place the packages in a different directory structure than the staging directory,
+    we can't use PublishingVersion 4 as the paths to the packages won't be a subdirectory of the artifacts (which is fine as we don't publish assets for source-only builds).
```
ViktorHofer, line 17: From what I understand, the eventual goal is to get everyone on V4 publishing and make it the default. Given that, we would need to solve this for SB as well. Please file a tracking issue for removing this.

eng/common/core-templates/job/publish-build-assets.yml
jkoritzinsky: I already cleared this with @mmitche that this is fine for this PR (I have https://github.com/dotnet/arcade/pull/15862 open for main)

eng/Publishing.props
jkoritzinsky: `Future` here means the pipeline artifact name that we'll create on Azure DevOps when uploading the artifacts at the end of the job.

eng/VmrLayout.props
jkoritzinsky: I wanted to use the default settings for v4 publishing here, which doesn't include the Configuration and uses this naming (majority of the work in this PR has been accounting for those differences).

repo-projects/sdk.proj
jkoritzinsky: Yes I need to spilt these into two conditions

```suggestion
    <BuildArgs Condition="'$(BuildWorkloads)' == 'true'">$(BuildArgs) /p:BuildWorkloads=true</BuildArgs>
    <BuildArgs Condition="'$(BuildWorkloads)' == 'true' or '$(DotNetBuildPass)' == '3'">$(BuildArgs) /p:PreviousBuildPassShippingPackagesDir=$(PreviousBuildPassShippingPackagesDir)/$(RepositoryName)</BuildArgs>
```

eng/Publishing.props
jkoritzinsky: I'll file the tracking issue along with the PR into the main branch.

eng/Publishing.props
mmitche: It doesn't exist yet at the time Publish.proj is run, so that's why I named it Future. Renaming is possible.

eng/Publishing.props
mmitche: Isn't `ArtifactsDir` a parent of both `ArtifactsStagingDir` and `ArtifactsTmpDir`? Could that be used for the future publish root, which would enable v4 publishing to be used?

Either way, I don't think the manifest functionality is interesting, right? If not, then we should be able to carve out some way in PushToBuildStorage so that v4 publishing can be used, without the FutureArtifactName/publish base path. Essentially making the task only interesting for copying around files.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -317,8 +317,8 @@ jobs:
     # Using build artifacts to enable publishing the vertical manifests to a single artifact from different jobs
     - ${{ if ne(parameters.buildSourceOnly, true) }}:
       - output: buildArtifacts
```
mmitche, line 319: Is it necessary to use build artifacts vs. a pipeline artifacts here? I thought you had found some way of downloading all artifacts from the pipeline that matched some pattern.

eng/Publishing.props
jkoritzinsky: ArtifactsStagingDir is under a separate directory from ArtifactsDir when running on Azdo.

I can change to using something like "ArtifactsStagingTmpDir" so it will live in a subdirectory.

eng/pipelines/templates/stages/vmr-build.yml
```diff
@@ -67,32 +67,28 @@ parameters:
...
-            symbolPublishingAdditionalParameters: '/p:PublishSpecialClrFiles=true'
+    scope: ${{ parameters.scope }}
+
+- ${{ if and(parameters.isBuiltFromVmr, eq(variables['System.TeamProject'], 'internal'), ne(variables['Build.Reason'], 'PullRequest')) }}:
+  - stage: Publish_Build_Assets
+    displayName: Publish Assets
+    dependsOn: VMR_Vertical_Build
+    jobs:
+    - template: /eng/common/templates-official/job/publish-build-assets.yml
+      parameters:
```
mmitche, line 83: Need to add publishingVersion: 4 here?

eng/pipelines/templates/jobs/vmr-build.yml
jkoritzinsky: Yes, but I would have needed to use eng/common.

Also, @ViktorHofer and others asked for an artifact that had all of the vertical manifests easily accessible for diagnostic purposes, so we'll still end up with one build artifact.

eng/tools/BuildComparer/AssetComparer.cs
```diff
@@ -19,7 +19,6 @@ public class AssetComparer : BuildComparer
     public AssetComparer(
         bool clean,
         AssetType? assetType,
-        string vmrManifestPath,
```
mmitche, line 22: The reason for getting rid of the merged manifest here was that we run the comparison before we publish assets?

eng/Publishing.props
mmitche: You might not even need that. PushToBuildStorage only checks for relative paths if FutureArtifactPublishBasePath is specified. Maybe we just don't need to specify it for SB since we don't need the asset manifest with that info?

eng/tools/BuildComparer/AssetComparer.cs
jkoritzinsky: Yep

eng/Publishing.props
jkoritzinsky: FutureArtifactPublishBasePath is initialized too late for me to be able to easily intercept.

Let's leave fixing this to main branch PR.

eng/VmrLayout.props
ViktorHofer: Can we change these default settings in Arcade? Those small differences in casing and the Configuration will lead to confusion.

eng/Publishing.props
ViktorHofer: Thanks for elaborating. Please add a comment.

eng/tools/BuildComparer/AssetComparer.cs
ViktorHofer: We could run the comparison tool after publish. Not sure if that's desirable.

eng/VmrLayout.props
jkoritzinsky: Yes, but at that point this is definitely not going into P5 as I'm going to have to undo the vast majority of the PR.

I'll put out a PR to Arcade to adjust the paths and update the PR for the main branch to undo most of the work

jkoritzinsky:
Opened https://github.com/dotnet/dotnet/pull/935 so I'll close this one.

PR #845, opened 2025-05-28
fix charecters typo

SimonCropp:


ViktorHofer (approved):

ViktorHofer:
@SimonCropp thanks for the comment typo fix. Note that we don't yet take community PRs in dotnet/dotnet aside from a few exceptions. This one is fine but wanted to share to set the expectation right.

akoeplinger:
Also we'd generally recommend that changes which touch only a single repo should still be done in the repo itself since there is more test coverage there.

PR #847, opened 2025-05-28
Output poison leak as message instead of warning

mthalman:
Fixes https://github.com/dotnet/source-build/issues/5221

Poison leaks show up as warnings which fail the build because warnings are now treated as errors due to the changes in https://github.com/dotnet/dotnet/pull/798.


ViktorHofer (approved):
Thanks for fixing

ViktorHofer:
Btw do we understand the poison leak reported files?

mthalman:
> Btw do we understand the poison leak reported files?

Yes, these are known and baselined and deemed acceptable: https://github.com/dotnet/dotnet/pull/230

PR #848, opened 2025-05-28
Move BuildComparer from VersionTools to build manifest package

mmitche:


mmitche:
Fixes build errors here https://dev.azure.com/dnceng/internal/_build/results?buildId=2716458&view=results

PR #852, opened 2025-05-28
Build RepositoryReferences as P2Ps and use NuGet's transitive P2P calculation

ViktorHofer:
Continuation of https://github.com/dotnet/dotnet/commit/1801a223d0cc621cb94e43a272014985bee87113

This improves scheduling the inner repos and with that reduces the overall build time by getting runtime (which is on the hot path) to start to build earlier. Additionally, this makes the VMR build more aligned with the rest of the product by using P2Ps to build dependencies.

---

NuGet is responsible for calculating transitive project references in general. A project's project.assets.json file contains a node for the transitive p2ps. At build time a target then runs that reads that information and adds the transitive p2ps to an item group so that transitives are also getting built as part of the `ResolveProjectReferences` target. That's how the product works.
 
In the VMR we weren't defining P2P for the repo projects and therefore we never had that information available in the "default" way. Instead we had a custom target that retrieved that information. My PR basically removes that custom target and gets us onto the common path. Of course with some additional complexity... mostly because of BuildPass stuff.

The PR also removes the `BuildRepoReferences` target as we now build RepositoryReference items as P2Ps directly and rely on the `ResolveProjectReferences` SDK target.

### DIFF
--- a/repo-projects/Directory.Build.props
+++ b/repo-projects/Directory.Build.props
@@ -16,6 +16,8 @@
 
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
+    <NoTargetsDoNotReferenceOutputAssemblies>false</NoTargetsDoNotReferenceOutputAssemblies>
+
     <ProjectDirectory>$([MSBuild]::NormalizeDirectory('$(SrcDir)', '$(RepositoryName)'))</ProjectDirectory>
 
     <!-- Paths to the version props files -->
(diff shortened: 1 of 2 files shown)


ViktorHofer:
This one is ready. Failures are unrelated (dry run signing issue in aspnetcore and android leg that times out)

akoeplinger:
> Failures are unrelated

I trust you but given we haven't tested the later build passes here and you're touching them, mind running an official build (with signing disabled) to validate it?

ViktorHofer:
> I trust you but given we haven't tested the later build passes here and you're touching them, mind running an official build (with signing disabled) to validate it?

I ran the `dotnet-unified-build-full` pipeline in this PR and inspected the `win-x64 BuildPass2` binlog. I couldn't find any regressions. The last two commits are just clean-up (adding comments and removing non necessary code) and shouldn't affect execution.

https://dev.azure.com/dnceng-public/public/_build/results?buildId=1057706&view=results

PR #930, opened 2025-06-03
Keep subdirectory for repo log files

akoeplinger:
I noticed in a recent PR that the log files for source-build-externals were flattened and not all logs kept due to name clashes.

We need to keep the directory structure when moving the log files.


ViktorHofer (approved):

akoeplinger:
looks good now:
image

PR #934, opened 2025-06-03
Doc: Feature Band Support Work Plan

mthalman:
Related to https://github.com/dotnet/dotnet/issues/1011


docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+* Since Arcade repo will not be available, all repos will need to use the bootstrap Arcade (`UseBootstrapArcade=true`).
+* While the `scenario-tests` repo doesn't have differing code for each feature band, it will still need to exist in each of the VMR's feature band branches because it is built as part of the pipeline run and used to validate the product. This is ok since it is not part of the shipped binaries.
+
+❓ Open Question:
+There will need to be some way to detect when we are in one of these 2xx branches. What mechanism should be used for that? Maybe checking `packageVersion` of the sdk in `source-manifest.json`?
+
+❓ Open Question:
+How to handle the multi-pass build legs where they are dependent on a shared component (e.g. `Windows_x86_BuildPass2`, `Windows_x64_BuildPass2`, `Windows_Workloads_x64_BuildPass2`)?
+
+#### Source-Only Builds
```
MichaelSimons, line 73: It would be helpful to clarify what is supported from a toolset/artifacts perspective.  Some of this can be inferred from the scenarios below but the document would benefit from being more explicit.  For example what version of the toolset and artifacts do we support building 2xx with?  Is it 1xx or 2xx or both?

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+For source-only builds, you're going to need to build all the source, including the source from the 1xx branch in order to produce a release from a 2xx branch.
+
+There are a variety of scenarios to consider:
+
+* Distro maintainer has not produced any 1xx assets before
+  * Distro maintainer needs to bootstrap 2xx for the first time:
+    * Actions:
+      1. Build 1xx branch using Microsoft artifacts
+      1. Rebuild 1xx branch using output of first build as input
+      1. Build 2xx branch using Microsoft artifacts and source-built 1xx artifacts as input
```
MichaelSimons: If this is a 200 build, the Microsoft artifacts would not be needed right?  They would only be needed for >200 for the first time?

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+      1. Rebuild 2xx branch using output of previous build and source-built 1xx artifacts as input
```
MichaelSimons: Can you clarify what 1xx artifacts are needed?  Perhaps clarifying artifacts vs toolset (SDK, arcade) would be benficial.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+This document builds on top of [Managing SDK Bands](https://github.com/dotnet/dotnet/blob/main/docs/VMR-Managing-SDK-Bands.md) with a concrete work plan for providing feature band support in the VMR. Please read that doc as a prerequisite.
+
+Document convention:
+* There can be a variable number of feature band releases (e.g. 10.0.1xx, 10.0.2xx, 10.0.3xx, etc). For purposes of this document, we mainly care about the distinction between the 1xx branch and all the other non-1xx feature band branches. For simplicity, 2xx will be used as a notation to mean any non-1xx feature band branch.
+
+## Plan Details
+
+### Branch Management
+
+The 2xx branch sources should accurately reflect what gets produced in the output. Since the shared component are only being produced in the 1xx branch, those associated repos will be excluded from the 2xx branch upon creation.
```
mmitche: Right now I'm on the side of avoiding subsetting the sources. Just being able to branch the repo and create a 2xx or 3xx branch, without deleting a bunch of files, probably has a ton of value. It keeps the shared history and means you can merge from 2xx back to 1xx if you wanted to.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+    VMR_1xx->>VMR_2xx: Dependency flow <br />(runtime build output packages)
+    VMR_1xx->>VMR_3xx: Dependency flow <br />(runtime build output packages)
+```
+
+This will require changes to Maestro to allow output of the VMR to flow to another branch of the VMR.
+
+### Build
+
+* In addition to NuGet package, we also need to provide a way to have access to the runtime and ASP.NET Core artifacts (e.g. `dotnet-runtime-<version>-linux-x64.tar.gz`) from 2xx branches. This is necessary for the SDK to be able to bundle things together. This should be able to be done using https://ci.dot.net, just as it is done in product repo builds.
+* Since Arcade repo will not be available, all repos will need to use the bootstrap Arcade (`UseBootstrapArcade=true`).
```
mmitche: I think arcade could be included if we wanted. There have been a few cases in the past where we've needed arcade updates in 2xx+ to implement new features in the SDK.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+* While the `scenario-tests` repo doesn't have differing code for each feature band, it will still need to exist in each of the VMR's feature band branches because it is built as part of the pipeline run and used to validate the product. This is ok since it is not part of the shipped binaries.
+
+❓ Open Question:
+There will need to be some way to detect when we are in one of these 2xx branches. What mechanism should be used for that? Maybe checking `packageVersion` of the sdk in `source-manifest.json`?
```
mmitche: Where do we need this knowledge?

docs/VMR-Feature-Band-Support-Work-Plan.md
mmitche: Yeah, I don't think we'd need MS 200 artifacts to build 200 for the first time, unless the 2xx branch takes a dependency on the 2xx SDK

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Branch Management
+
+The 2xx branch sources should accurately reflect what gets produced in the output. Since the shared component are only being produced in the 1xx branch, those associated repos will be excluded from the 2xx branch upon creation.
+
+There should be tooling to help with this exclusion process for the VMR maintainers. Similarly, the associated projects in the `repo-projects` directory should be deleted along with updating the appropriate `RepositoryReference` usages.
+
+### Dependency Flow
+
+Repos in the 2xx branch of the VMR which have dependencies on repos solely defined in the 1xx branch (e.g. `sdk` repo's dependency on `runtime`) will need to have dependency flow that reflects the version of the package produced from the VMR's 1xx branch.
```
mmitche: Open question is how/where the 1xx dependencies are defined in the 2xx branch

docs/VMR-Feature-Band-Support-Work-Plan.md
mthalman: * If we're going to be keeping the shared component repos around, then I assume we're going to have the entire repo graph, `RepositoryReferences`, left untouched as well. So we'll need a way to trim the graph conditionally.
* SDK needs this in order to construct the URIs for gathering the redist artifacts. If in a 1xx branch, it will use a local path. If in a 2xx branch, it [should use an artifacts package feed](https://github.com/dotnet/sdk/blob/427918c5c680e1ee087438e85fcb572e509ea26d/src/Layout/redist/targets/RestoreLayout.targets#L308-L320).

docs/VMR-Feature-Band-Support-Work-Plan.md
mthalman: I was generalizing this for any 2xx release. I can add a specific scenario for just the 200 release.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+Document convention:
+* There can be a variable number of feature band releases (e.g. 10.0.1xx, 10.0.2xx, 10.0.3xx, etc). For purposes of this document, we mainly care about the distinction between the 1xx branch and all the other non-1xx feature band branches. For simplicity, 2xx will be used as a notation to mean any non-1xx feature band branch.
+
+## Plan Details
+
+### Branch Management
+
+The 2xx branch sources should accurately reflect what gets produced in the output. Since the shared component are only being produced in the 1xx branch, those associated repos will be excluded from the 2xx branch upon creation.
+
+There should be tooling to help with this exclusion process for the VMR maintainers. Similarly, the associated projects in the `repo-projects` directory should be deleted along with updating the appropriate `RepositoryReference` usages.
```
ViktorHofer: It might be better to not delete the repo-projects but condition them based on the feature band. This would allow to forward merge changes from 10.0.2xx to 11.0.1xx (the dotnet/sdk model basically) without a ton of conflicts.

docs/VMR-Feature-Band-Support-Work-Plan.md
NikolaMilosavljevic: Can we have something like `/Dependencies/Dependency/Branch` in `Version.Details.xml`? If present it would point at dependent VMR branch, and will be updated when we flow lower SDK level to a higher one.

premun: Product dependencies in the root of the 2xx eng/V.D.xml?

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+    participant VMR_2xx as VMR<br />release/10.0.2xx
+    participant VMR_3xx as VMR<br />release/10.0.3xx
+
+    runtime->>VMR_1xx: Flow of runtime change
+    Note over VMR_1xx: 📦 VMR build output packages are built
+    VMR_1xx->>VMR_2xx: Dependency flow <br />(runtime build output packages)
+    VMR_1xx->>VMR_3xx: Dependency flow <br />(runtime build output packages)
+```
+
+This will require changes to Maestro to allow output of the VMR to flow to another branch of the VMR.
```
premun: We can already flow packages between the branches if they are defined in `V.D.xml`

docs/VMR-Feature-Band-Support-Work-Plan.md
ViktorHofer: The `packageVersion` metadata will be removed from source-manifest.json pretty soon. Mostly because the information is incorrect for repos that do assetless publish.

https://github.com/dotnet/arcade-services/issues/2428

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,92 @@
...
+
+* In addition to NuGet package, we also need to provide a way to have access to the runtime and ASP.NET Core artifacts (e.g. `dotnet-runtime-<version>-linux-x64.tar.gz`) from 2xx branches. This is necessary for the SDK to be able to bundle things together. This should be able to be done using https://ci.dot.net, just as it is done in product repo builds.
+* Since Arcade repo will not be available, all repos will need to use the bootstrap Arcade (`UseBootstrapArcade=true`).
+* While the `scenario-tests` repo doesn't have differing code for each feature band, it will still need to exist in each of the VMR's feature band branches because it is built as part of the pipeline run and used to validate the product. This is ok since it is not part of the shipped binaries.
+
+❓ Open Question:
+There will need to be some way to detect when we are in one of these 2xx branches. What mechanism should be used for that? Maybe checking `packageVersion` of the sdk in `source-manifest.json`?
+
+❓ Open Question:
+How to handle the multi-pass build legs where they are dependent on a shared component (e.g. `Windows_x86_BuildPass2`, `Windows_x64_BuildPass2`, `Windows_Workloads_x64_BuildPass2`)?
```
akoeplinger: Workloads will definitely need some changes, e.g. the .msi's from the Windows_Workloads_x64_BuildPass2 in 1xx need to flow into 2xx so they can get used by BP3 there

docs/VMR-Feature-Band-Support-Work-Plan.md
mthalman: Will there be any canonical way to determine the SDK version at the VMR root level?

/cc @premun

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -9,9 +9,16 @@ Document convention:
...
+* Avoids confusion about which source is relevant in the production of the build
+* Avoids the need to determine if or when repo sources should be updated with the latest from the 1xx branch.
+* Avoids any compliance issues with the code getting stale if it hasn't been updated (e.g. CodeQL alerts).
 
-There should be tooling to help with this exclusion process for the VMR maintainers. Similarly, the associated projects in the `repo-projects` directory should be deleted along with updating the appropriate `RepositoryReference` usages.
+One exception is the Arcade repo which will be present in all branches. It's not technically required to exist in the 2xx branch in all cases. But there have historically been cases where Arcade changes are necessary in order to support new SDK features not in the 1xx branch. To account for those cases, it'll be a smoother process to simply always have Arcade present in those branches. There isn't an impact in the output of the build since Arcade isn't a shipping component.
```
premun, line 17: (How) will the Arcade be updated? We should also distinguish between root-level Arcade dependency and `src/arcade` and describe their lifecycle.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -9,9 +9,16 @@ Document convention:
...
+
+Rather than deleting the associated projects in the `repo-projects` directory, it may make sense to keep those and trim the dependency graph of `RepositoryReferences` using a condition based on whether it's the 1xx branch or not.
+
+There should be tooling to help with this exclusion process for the VMR maintainers. This can potentially be integrated into the broader workflow that is necessary when branching such as branding updates and pipeline changes. The tooling could potentially make use of the dependency graph described by the `repo-projects` directory to determine which repos need to be removed.
```
premun, line 21: If there should be tooling for snapping bands, we should possibly describe what the snap looks like in detail?
I know something like this is already in the original design documents but maybe we know more now.

premun:
For the roadmap / schedule section - it would be good if we understood when we can expect what kind of intermediate results, e.g. "when can we start testing the release process which depends on having some sort of a 2xx build".

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,136 @@
...
+```mermaid
+flowchart TD
+  VerProps[Import Versions.props]
+  ImportPVP[<b>Override with Previous.Versions.Props</b><br/><br/><i>Includes versions of assets from root VMR's Version.Details.xml file that are also present in the repo's Version.Details.xml</i>]
+  ImportCVP[<b>Override with Current.Versions.Props</b><br/><br/><i>Includes versions of assets built during the current VMR build that are also present in the repo's Version.Details.xml</i>]
+  VerProps-->ImportPVP-->ImportCVP
+```
+
+- Import Versions.props
+- Import generated Previous.Versions.props -
```
MichaelSimons: What dependencies are included here?  
Will it be all artifacts, only shared components, other?  
Is this going to change the MSFT build behavior?  Specifically would this lift dependencies that are not intended to be lifted?  
When do these get updated?  
Does this imply the versions from PSB will not get read dynamically?  I am concerned about potentially breaking distro maintainer scenarios where versions may be different.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,136 @@
...
+- **Distro partner previously-source-built update process** - Right now distro partners will build a given release by taking a set of previously SB artifacts (provided by MS or source-built by the distro partner), and pass these to the build script via --with-packages and --with-sdk. Going forward, this step will either be preceded by an explicit step to update the root `Version.Details.xml` with these packages, or this switch will imply such a step. Therefore, when the build is running, the previously source-built inputs and their versions are specified by the `Version.Details.xml` file, and NuGet will find those packages in the directory passed via --with-packages. Essentially, we are eliminating some implicit behavior.
+
+### Toolset
+
+❓ Open Question:
+A policy is needed which defines what toolset version each feature band branch can have. For example, should all feature band branches be required to use only the 1xx toolset? Can there be a mix of some using 1xx while others use a higher version?
```
MichaelSimons: Per discussions, this will have to be the n-1 toolset from the feature band.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,136 @@
...
+❓ Open Question:
+There will need to be some way to detect when we are in one of these 2xx branches. What mechanism should be used for that? Maybe checking `packageVersion` of the sdk in `source-manifest.json`?
+
+#### Source-Only Builds
+
+For source-only builds, you're going to need to build all the source, including the source from the 1xx branch in order to produce a release from a 2xx branch.
+
+There are a variety of scenarios to consider:
+
+1. Has the distro maintainer produced any 1xx assets before from the previous release?
```
MichaelSimons, line 79: Would it be possible to skip the building of the tools?  A distro maintainer only releasing 2xx sdk wouldn't need to build all of the tools.

MichaelSimons: Already asked later on in https://github.com/dotnet/dotnet/pull/934/files#diff-b114dee811ac74882f25eeef24c22981b35aba9f6809cd5ed797cebf26de072dR112.

docs/VMR-Feature-Band-Support-Work-Plan.md
mthalman: > (How) will the Arcade be updated?

It would be updated in the same way it's updated in the 1xx branch. Is that an issue?

mthalman: It's specifically referring to the presence of `src/arcade`. The root level Arcade stuff is required to be there regardless and will be updated via rebootstrapping.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,136 @@
...
+
+In order to facilitate the actions that will need to be taken by a distro maintainer or developer, new scripts should be provided (or existing ones updated). Specifically, provide a way to combine the output artifacts from multiple builds to be used as input. An alternative to combining artifacts is to have separate tarball inputs. See the implications of this design choice in [poison leak detection](#poison-leak-detection).
+
+❓ Open Question:
+
+Should a distro maintainer be required to build the *entire* 1xx branch VMR (including the SDK)? That's easier to implement but also potentially wasteful if they're not going to be producing a release for that feature branch. Otherwise, there'd need to be some way to filter which repos get built such that only the repos excluded from the 2xx branch get built. Including all packages, including tools, rather than just shared components, has implications for [poison leak detection](#poison-leak-detection) which need to be considered.
+
```
NikolaMilosavljevic, line 101: It would be good to add clarity around build command requirements. I think `OfficialBuildId` becomes mandatory if building 2xx or higher branches. Versioning packages with `-ci` or `-dev` would not be enough, to differentiate between those built in 1xx (and brought via 1xx PSB archive) and packages build in 2xx + - this will only affect SDK/Roslyn packages.

mthalman:
Moving this out of draft as just a formality since people have already been reviewing.

docs/VMR-Feature-Band-Support-Work-Plan.md
premun: > It would be updated in the same way it's updated in the 1xx branch. Is that an issue?

Arcade doesn't snap branches for different bands. So backflow won't work. Will we be forbidden from making changes to it in the 2xx branch?

mthalman: > Arcade doesn't snap branches for different bands. So backflow won't work.

So how does this work today for servicing? @mmitche said that there can changes made to Arcade: https://github.com/dotnet/dotnet/pull/934#discussion_r2125150010

> Will we be forbidden from making changes to it in the 2xx branch?

No, the whole reason for including it is to allow modifications to support new SDK features.

premun: > It would be updated in the same way it's updated in the 1xx branch. Is that an issue?

This means every Arcade commit forward flows to 1xx and 2xx.
We can't backflow from 2xx because we can't have 1:N bi-directional code flow.

If we change 2xx `src/arcade` folder in the VMR, we must consider that future forward-flows of Arcade to 1xx might hit conflicts then.

mmitche: Usually the scenario is that the 2xx or 3xx branch (in preview) needs a new arcade, while 1xx is in a code complete/frozen state for the next release. So the 2xx branch is just moving ahead of what 1xx has.

We would only allow backflow from 1 branch (probably 1xx). Note that Arcade is not unique in this regard. When roslyn starts inserting into 10.0.2xx and 11.0.1xx, we will need to choose 1 backflow source, and avoid changes in the other branch.

**If** we could solve this problem and allow backflow from multiple branches (like a merge from multiple topic branches back to a common base) then things get a LOT easier.

docs/VMR-Feature-Band-Support-Work-Plan.md
mmitche: > Is this going to change the MSFT build behavior? Specifically would this lift dependencies that are not intended to be lifted?

After chatting with Michael yesterday, I have this concern as well. I think we will need some experimentation. I don't think we fully comprehend the various problems yet. Understanding them more fully may show us where our current VMR dependency system is lacking in expressivity.

Laying out the problem:

Dependency **flow** is not explicit the Version.Details.xml files. It's only "potential" flow, really. If a build containing X or Y asset is applied to a V.D.xml containing an unpinned X or Y asset, then X or Y will get updated. But whether those actually get updated is a function of build order in the VMR, subscriptions in old-style builds, and additional metadata on subscriptions like excludeAssets. Some information about whether a given asset should get updated before building is not visible just by looking at the Version.Details.xml file.

Let's take templating, for instance. It has a number of test-only dependencies on the runtime listed in its Version.Details.xml file:
- The subscription dependency flow says that these should not get updated when you backflow from VMR->templating. The repo sticks with 9.0.3 versions.
- The MS VMR build dependency flow agrees that these should **not** get updated when you build, because templating doesn't depend on runtime
- The distro partner VMR build dependency flow says these **should** get updated when you build, because templating does depend on runtime.

These requirements are obviously in tension, but the resolution of this tension is via various definitions of how flow works. Now let's take 2xx, which doesn't build the runtime. The proposed solution is that we essentially specify "override any dependency produced by the 1xx build that matches a dependency in a 2xx repo by using a root V.D.xml file".
- For distro partner builds, this is **correct**. It's also making the previous implicit behavior (override any matching dep using PSB) explicit.
- For MS VMR builds, this is **not correct**. The proposed method is essentially making an all-assets "subscription" from runtime->templating, which is not what we do in 1xx. 1xx uses a Maestro partial assets subscription to only update a few things.

This is just the ongoing tension between how distro partner VMR builds and MS VMR builds work. Fundamentally they have different dependency flow, and we don't have the right tools to express that dependency flow in just the V.D.xml alone. The proposed solution drags MS towards the distro partner methodology. IMO, this is not a bad thing. It would be awesome to be more coherent overall. But we should recognize that this will likely break things. There are likely repos which stay on old versions of input packages for legitimate product reasons.

Let's say, then, that we didn't flow 1xx directly to 2xx. We instead flowed 1xx outputs to individual 2xx repos via dependency flow, then FF-ed into the 2xx VMR. What we would see then is:
- Distro partner VMR builds would use 1xx outputs via `--with-packages`, and dependency flow will work as desired (since this is just essentially a bootstrap flow).
- Because 1xx outputs would flow to various 2xx repos via traditional dependency flow, we would have the ability to use excludeAssets to encode the **future** VMR dependency flow (after FF) at a higher fidelity than just a big hammer of "all 1xx goes to 2xx". **Side note: excludeAssets isn't available on traditional subscriptions at this time. It was never implemented because we thought it was an anti-pattern. These days, though, we have some cases where a repo wants to take 10.0 inputs for testing, but stay on an 8.0 arcade, so we're going to add it. Is this an anti-pattern? hmmmm.**

Of course, this is **not** a good solution. It essentially has the same coherency issues we have with .NET 9 and prior builds, but worse. Slower and less obvious when things are incoherent.

So what to do? Things seem to be pointing towards us needing to be able to more completely describe dependency flow in a way that satisfies both distro partner and MS builds. Right now we have 4 sources of dependency flow info:
- Which subscriptions are present/what assets do they flow (from Maestro)
- The build order of distro partner builds (from VMR)
- The build order of MS builds (from VMR)
- The unpinned dependencies a repo's V.D.xml (from repo)

docs/VMR-Feature-Band-Support-Work-Plan.md
mthalman: Can we just not support backflow of Arcade except for the 1xx branch? And if you want to make updates to Arcade in the 2xx branch, you can do that but you need to manually backport it.

docs/VMR-Feature-Band-Support-Work-Plan.md
mthalman: I don't think there's any restriction from using ci/dev builds as input. In that scenario, the 1xx branch would produce a package version like `10.0.104-dev`, which wouldn't conflict with the packages from 2xx which would have `10.0.200-dev`.

docs/VMR-Feature-Band-Support-Work-Plan.md
mthalman: How would that work due to the lifecycle of the feature bands? For example, if 3xx uses 2xx that will be problematic since 2xx will eventually become unsupported while 3xx is still supported.

docs/VMR-Feature-Band-Support-Work-Plan.md
mmitche: Yeah, we wouldn't have backflow except to 1xx.

docs/VMR-Feature-Band-Support-Work-Plan.md
MichaelSimons: I will try and explain what I was envisioning that would allow feature bands to make use of the latest toolset.

When .200 is built, it will build with the current 1xx toolset.  For illustration purposes suppose it is .106. When .201 is built,  is will build with .200.  A similar pattern will be used for 3xx.  .300 will build with the current 2xx toolset, say .206.  .301 will use .300.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,146 @@
...
+In order to facilitate the actions that will need to be taken by a distro maintainer or developer, new scripts should be provided (or existing ones updated). Specifically, provide a way to combine the output artifacts from multiple builds to be used as input. An alternative to combining artifacts is to have separate tarball inputs. See the implications of this design choice in [poison leak detection](#poison-leak-detection).
+
+❓ Open Question:
+
+Should a distro maintainer be required to build the *entire* 1xx branch VMR (including the SDK)? That's easier to implement but also potentially wasteful if they're not going to be producing a release for that feature branch. Otherwise, there'd need to be some way to filter which repos get built such that only the repos excluded from the 2xx branch get built. Including all packages, including tools, rather than just shared components, has implications for [poison leak detection](#poison-leak-detection) which need to be considered.
```
mthalman: I lean towards not supporting this for the following reasons:
* We're not going to be able to verify that the output of a partial build works from the 1xx branch. For example, we can't run scenario tests on it.
* It would require us to have more build legs to cover this scenario. We already have a requirement to publish the output of the 1xx build so that it can be consumed by the 2xx build. These would be the legs that would do that. But we'd also need build legs that still do the full SDK build as part of the normal 1xx release validation. So we'd be adding a lot more build legs to ensure that this flow works.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Source-Only Build Legs
+
+Given that the packages outputted from the 1xx branch are needed in order to source build the 2xx branch, changes will be needed to the pipelines. In order to make the build efficient, we can collect the output of the 1xx branch's official build and feed that in as input to the 2xx branch build. There's no need to build the 1xx branch again from source since it's already been done.
+
+In order to do this, there needs to be a mechanism to publish the source-built packages from a 1xx official build so that they can be consumed by a 2xx build. There are a variety of options that can be considered:
+* Publish each source-built NuGet package to a NuGet feed (probably specific to source build). These would then be downloaded all together by using [NuGet search functionality](https://learn.microsoft.com/nuget/reference/nuget-client-sdk#search-packages) by filtering on the package version.
+* Publish each source-built artifacts tarball to blob storage. This would then be downloaded by constructing a URL that contains the necessary version.
```
mthalman, line 124: The benefit of using a tarball is that we get the benefit or storing other files besides just packages. For example, we may end up needing the VerticalManifest.xml file in the tarball so that we know which repo each package came from. That's necessary for source build poison leak detection where we would allow packages from shared components to be "leaked" but nothing else.

docs/VMR-Feature-Band-Support-Work-Plan.md
mmitche: Yeah, I don't think we should support this. Also, wouldn't this not work? A repo owner would need to build the full 1xx branch in order to get a bootstrap SDK in order to rebuild 1xx to get the runtime bits. In theory, it means that when building the servicing N+1, they could use an old 1xx SDK they already bootstrapped to build the runtime bits, but it seems more efficient to just build the whole set each time. In addition, based on build graph, you would only eliminate a few nodes. You'd still need to build roslyn, nuget-client, msbuild because the runtime repos depend on them.

docs/VMR-Feature-Band-Support-Work-Plan.md
mmitche: Yes, I think it makes a ton of sense to use this method. This also reduces the possibility of SB packages with the same identities as MS's but different content leaking somewhere.

mthalman: We'd probably store it in https://ci.dot.net. How does the retention work for other archive files that get published there? Do they just remain there indefinitely? Do we see any issue with doing the same with this scenario? The important thing we want to allow from a retention policy is that any person should be able to go to a past release tag from a 2xx release and still be able to build it with access to the 1xx artifacts associated with it at that time.

mmitche: Retained forever.

docs/VMR-Feature-Band-Support-Work-Plan.md
premun: This will have to be set somewhere in `Versions.props`

docs/VMR-Feature-Band-Support-Work-Plan.md
mmitche: The natural gesture is just to have all the deps and not embed any VMR or branch specific info in there.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -0,0 +1,139 @@
...
+  * That's ok because the static version defined in Version.Details is the one we want to consume
+
+**Considerations**
+
+- **Validation of 2xx+ repos** - A repo is validated on backflow based on what it built against. So a 1xx SDK repo branch receives the source changes in the 1xx VMR sdk sources, as well as the outputs of the VMR build, which includes the runtime that the SDK built against in that VMR build. This means that on backflow, the SDK PR is validating the SDK's sources against the updated SDK dependencies. For 2xx+ branches, only a subset of outputs is produced, so a backflow PR would not include all the dependencies that the 2xx SDK needs. To deal with this, we will attempt to use the CoherentParentDependency feature. CoherentParentDependency attributes will be placed on the SDK dependencies that are not produced in the 2xx build, tying them to a dependency produced in the 2xx build. When backflow happens and the dependencies are updated for a repo, the CPDs will resolve to the versions specific in the root VMR Version.Details.xml if they are not produced in the current build.
```
mthalman: @mmitche - Is this still relevant? Since we'll represent all the dependencies each repo's V.D.xml instead of a root-level file, that V.D.xml info already does backflow to the product repo. So does this just work?

mmitche: No longer relevant based on the feature above.

docs/VMR-Feature-Band-Support-Work-Plan.md
```diff
@@ -44,28 +44,24 @@ This could be done a few different ways, but there are some explicit goals that
...
+To achieve 2xx+ builds, we will use the "classic" dependency flow methodology that .NET has used for many years but with a variation to account for VMR-to-VMR flow. When flow happens from the the VMR's 1xx to 2xx branch, all dependencies across all relevant repos will be updated in one PR. This ensures coherency. The dependencies being updated are applied to the individual repos' Version.Details.xml files. There will not be a root VMR Version.Details.xml that aggregates all of these dependencies.
 
-When building a repo within the VMR, the following version flow will be applied:
+Let's use roslyn's dependency on command-line-api as an example to demonstrate the flow.
```
mmitche, line 49: @premun What do you think a feature like this would cost in Maestro? Essentially, when 1xx flows to 2xx, we would have subscription metadata that describes the flow to each of the 2xx VMR repo dirs, complete with ExcludeAssets.

mmitche: The main problem with putting the info in the root V.D.xml file is that it changes the build behavior for 2xx

premun: Would it mean we would have a dedicated subscription per each target folder in the 2xx branch? Each could have different exclusions?

I was sort of accounting for something like this 2 years ago when adding the `SourceDirectory` and `TargetDirectory` attributes on subscriptions.

Maybe we could allow non-code-enabled subscriptions to define a `TargetDirectory` and it would just hit that `src/[TargetDirectory]`?



I'd say ~2 weeks of work?

mmitche: We'd also want these to come in as a single PR. We could batch them? But it seems possible to screw up.

mthalman: BTW, this work can be tracked with https://github.com/dotnet/dotnet/issues/1108

premun: Batching is not a problem - only we need to force-trigger maybe somehow so that Maestro doesnt' wait for the builds to fail

mthalman:
I think the doc accurately reflects the outline of the work now. Issues have been created and available from the epic at https://github.com/dotnet/dotnet/issues/1011. Further details on implementation can be provided in those issues.

I still need an approval to get this PR merged.

PR #935, opened 2025-06-03
Enable v4 publishing in the VMR

jkoritzinsky:
Enable v4 publishing for UnifiedBuild jobs in the VMR.

Based on feedback from https://github.com/dotnet/dotnet/pull/823, I'm going to try to keep the existing artifact paths to simplify the change.

Internal build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2726405&view=results

Fixes https://github.com/dotnet/dotnet/issues/1062


eng/pipelines/templates/stages/source-build-and-validate.yml
```diff
@@ -207,45 +207,45 @@ stages:
               downloadFilePatterns: assets/Release/**/*.tar.gz
               copyDestination: $(Build.SourcesDirectory)/artifacts
               ${{ if eq(leg.distro, 'centos') }}:
-                artifactName: ${{ format(leg.buildNameFormat, variables.centOSStreamName) }}_${{ leg.targetArchitecture}}_Artifacts
+                artifactName: ${{ format(leg.buildNameFormat, variables.centOSStreamName) }}_${{ leg.targetArchitecture}}_BuildLogs_Attempt1
```
ViktorHofer: I guess the obvious reaction here is whether the Attemp1 hardcode will cause issues with reruns?

jkoritzinsky: Yeah I'd rather have these logs copied into the artifacts directory instead of logs being used as inputs to later jobs.

This change is also due to this PR fixing https://github.com/dotnet/source-build/issues/5072

ViktorHofer: > Yeah I'd rather have these logs copied into the artifacts directory instead of logs being used as inputs to later jobs.

If I understand you correctly you are saying that we should put logs into the _Artifacts payload and revert this change? What was the intentional reason for having two separate payloads?

jkoritzinsky: I think we should have the files that are consumed here put in the Artifacts payload and download them from there instead of putting them in the build logs payload.

src/aspire/eng/Publishing.props
```diff
@@ -1,7 +1,6 @@
 <Project>
 
    <PropertyGroup>
-      <PublishingVersion>3</PublishingVersion>
```
akoeplinger, line 4: other repos still set this, why is aspire special?

jkoritzinsky: Aspire uses the CouldBeStable metadata, and Matt recommended moving to v4 to propagate that. Other repos that explicitly opt into v3 can (and probably should) be moved to v4 in the VMR, but it wasn't necessary for completing this work.

PR #950, opened 2025-06-04
Fix CreateBaselineUpdatePR tool to preserve baseline files

Winniexu01:
Fixes: https://github.com/dotnet/source-build/issues/5123

Previously, the `CreateBaselineUpdatePR` tool would delete the baseline files when no updated baseline file was produced, which was incorrect behavior. Now the tool will preserve the baseline files when there is no updated baseline file.


eng/tools/CreateBaselineUpdatePR/PRCreator.cs
```diff
@@ -14,6 +14,7 @@ public class PRCreator
     private readonly string _repoOwner;
     private readonly string _repoName;
     private readonly GitHubClient _client;
+    private readonly List<string> originalBaselineFile = new List<string>();
```
mthalman: There should be no need to define this as a field. It can be a local variable in the `UpdateBaseFileAsync` method.

It can also be simplified as:

```c#
List<string> originalBaselineFile = [];
```

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
```diff
@@ -245,6 +248,41 @@ private async Task<List<NewTreeItem>> UpdateFileAsync(List<NewTreeItem> tree, st
...
+            {
+                originalBaselineFile.Add(Path.Combine(path,item.Name));
+            }
+        }
+        foreach(var item in updatedTestsFiles)
+        {
+            originalBaselineFile.RemoveAll(file => item.Value.Any(v => v.Contains(Path.GetFileNameWithoutExtension(file))));
+        }
+
+        if (originalBaselineFile.Count > 0)
```
mthalman: This `if` is unnecessary since the `foreach` will simply not enumerate anything if it's empty.

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
```diff
@@ -245,6 +248,41 @@ private async Task<List<NewTreeItem>> UpdateFileAsync(List<NewTreeItem> tree, st
         return tree;
     }
 
+    private async Task<List<NewTreeItem>> UpdateBaseFileAsync(List<NewTreeItem> tree, string path, string targetBranch, Dictionary<string, HashSet<string>> updatedTestsFiles)
```
mthalman, line 252: It'd be helpful to add comments stating what's being done and why.

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
```diff
@@ -245,6 +248,41 @@ private async Task<List<NewTreeItem>> UpdateFileAsync(List<NewTreeItem> tree, st
...
+        foreach(var item in updatedTestsFiles)
+        {
+            originalBaselineFile.RemoveAll(file => item.Value.Any(v => v.Contains(Path.GetFileNameWithoutExtension(file))));
+        }
+
+        if (originalBaselineFile.Count > 0)
+        {
+            foreach (var file in originalBaselineFile)
+            {
+
```
mthalman:
```suggestion
```

Winniexu01:
Hi @ellahathaway , from a logical perspective, the issue is not related to  [this](https://github.com/dotnet/dotnet/blob/0faf106b5d8e6f3677df85a2b0f8d3011b7d87ce/eng/tools/CreateBaselineUpdatePR/PRCreator.cs#L222-L229) code. It will not be executed when no updated baseline file (i.e., Updated *.*) is generated, it only works when a Updated*.* be produced with no meaningful content and the corresponding baseline file is already included in the file tree.

In the [earlier code](https://github.com/dotnet/dotnet/blob/f09dba3ccec38b8ab6000fee4804f3783771aba6/eng/tools/CreateBaselineUpdatePR/PRCreator.cs), due to the large size of the dotnet repository and limitations of the GitHub API, the full file tree under the path `test/Microsoft.DotNet.SourceBuild.Tests/assets/LicenseScanTests` was not being gathered (this issue has been fixed by https://github.com/dotnet/dotnet/pull/331). As a result, the [testResultsTreeItems](https://github.com/dotnet/dotnet/blob/f09dba3ccec38b8ab6000fee4804f3783771aba6/eng/tools/CreateBaselineUpdatePR/PRCreator.cs#L45C13-L45C33) filed was empty. According to the subsequent logic, if the test did not generated a file matching the pattern 'Updated' then no new item would be added to the new file tree, please check this [code](https://github.com/dotnet/dotnet/blob/0faf106b5d8e6f3677df85a2b0f8d3011b7d87ce/eng/tools/CreateBaselineUpdatePR/PRCreator.cs#L124-L134). Consequently, the new branch created based on `testResultsTreeItems` would exclude the corresponding 
 baseline file, which appeared as deleted in the pull request.

Now in the latest code, the file tree is built using using a layered traversal approach, which ensures completeness. Therefore, even in timeout scenarios, the file tree remains intact and the new branch includes the related baseline file. 

The issue occurred because the `Updated` baseline file was not generated, which led to the baseline file appearing as deleted in the PR. However, the actual reason is that the file tree for the relevant path in the ealir code was empty, so the baseline file was never be added into the new branch. It was not that the code explicitly deleted the file, but rather that it was omitted from the new tree, resulting in it being interpreted as a deletion in the PR. So this is why I previously decided to close the issue with https://github.com/dotnet/dotnet/pull/331.

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
Winniexu01: Update: https://github.com/dotnet/dotnet/pull/950/commits/a68f8298e9821ad98abfbfa9e30b56db02986f8c

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
Winniexu01: Update: https://github.com/dotnet/dotnet/pull/950/commits/a68f8298e9821ad98abfbfa9e30b56db02986f8c

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
Winniexu01: Update: https://github.com/dotnet/dotnet/pull/950/commits/a68f8298e9821ad98abfbfa9e30b56db02986f8c

eng/tools/CreateBaselineUpdatePR/PRCreator.cs
Winniexu01: Update: https://github.com/dotnet/dotnet/pull/950/commits/a68f8298e9821ad98abfbfa9e30b56db02986f8c

ellahathaway:
I see what you're saying. As long as the file exists in the original tree, the file only gets replaced if it has an associated "Updated*" file. Because the original tree was previously being "cut off", the updated tree was reflecting the "cut off" tree and thus deleting the relevant files.

I verified that this is the behavior by running the PR creator locally. We can close this PR.

Winniexu01:
Got it, I will close this PR and the related issue.

PR #951, opened 2025-06-04
Fix dotnet version in arcade global.json

akoeplinger:
Mistake from https://github.com/dotnet/dotnet/pull/933


ViktorHofer:
Feel free to fast merge as that version isn't used in the VMR anyway.

PR #1080, opened 2025-06-09
VMR product versioning document

mmitche:


docs/VMR-Product-Versioning.md
```diff
@@ -0,0 +1,162 @@
+# The Unified Build Almanac (TUBA) - Product Versioning
+
+## Purpose
+
+This document details how the product is versioned over a given major version lifecycle.
+
+## Terminology
+
+This document uses the following terminology:
+- **Branding** - Refers to the variosu product version information. Typically MAJOR.MINOR.PATCH, pre-release and build-varying versioning information if present.
```
akoeplinger:
```suggestion
- **Branding** - Refers to the various product version information. Typically MAJOR.MINOR.PATCH, pre-release and build-varying versioning information if present.
```

docs/VMR-Product-Versioning.md
```diff
@@ -0,0 +1,162 @@
...
+
+#### Customer UX
+
+A simple version number is easier to understand/address than a more complex one.
+
+#### Compatibility with distributions not built by Microsoft
+
+All builds of .NET all have strong version dependencies between components. For instance, a package dependency references a specific version of another package. When stable branding is not in use, two successive builds of .NET with different official build ids will have different package ID for System.Text.Json. Therefore any references to System.Text.Json will also vary. This means that that builds of .NET with different official build ids will not necessarily work seamlessly with each other. This is an important consideration when we recognize that Microsoft is *not* the only builder of .NET. Our Linux distro partners also build their own SDKs and runtimes, and those SDKs and runtimes have dependencies on assets produced by Microsoft (e.g. packages on nuget.org).
```
akoeplinger:
```suggestion
All builds of .NET have strong version dependencies between components. For instance, a package dependency references a specific version of another package. When stable branding is not in use, two successive builds of .NET with different official build IDs will have different package versions for System.Text.Json.

Therefore any references to System.Text.Json will also vary. This means that builds of .NET with different official build IDs will not necessarily work seamlessly with each other. This is an important consideration when we recognize that Microsoft is *not* the only builder of .NET. Our Linux distro partners also build their own SDKs and runtimes, and those SDKs and runtimes have dependencies on assets produced by Microsoft (e.g. packages on nuget.org).
```

docs/VMR-Product-Versioning.md
```diff
@@ -0,0 +1,162 @@
...
+  - 10.0.0 - Stable
+  - 10.0.25172 (patch varies build to build based on date+revision) - Non-Stable
+  - 10.0.0-preview.2 - Stable
+  - 10.0.100-preview.2.1234.5 - Non-Stable
+
+## NuGet, Runtime and SDK Versioning Principles
+
+### Major, Minor and Patch Versions
+
+.NET follows [Semantic Versioning](https://semver.org/) to communicate the nature of changes in each release. The package version numbers are structured as `MAJOR.MINOR.PATCH`, though since .NET Core 3.1, only the MAJOR and PATCH components are actively used. There is no plan to use MINOR versiosn going forward. The MAJOR version is incremented for yearly releases that introduce breaking changes or significant new features that may not be backward compatible, while the PATCH version is incremented for releases that include only backward-compatible bug fixes or small improvements.
```
MichaelSimons: > .NET follows [Semantic Versioning](https://semver.org/) to communicate the nature of changes in each release.

I don't like this statement as this is not true for the sdk feature bands.

docs/VMR-Product-Versioning.md
```diff
@@ -0,0 +1,162 @@
...
+
+In order for external partners to produce distributions that are compatible with Microsoft's, we need to either communicate the build-varying information to them (the official build ID), or we need to use stable branding to ensure that the build-varying information does not influence product functionality.
+
+#### Changes in .NET 10
+
+In .NET 10, preview releases will be stable branded. This is to ensure cross-distribution compatibility, as noted above.
```
MichaelSimons, line 55: If I follow correctly, the versions of non-shipping assets will still vary between builds.  I don't think this is the behavior most distro maintainers will want and therefore will want to provide a build id.

docs/VMR-Product-Versioning.md
mmitche: Good point. Bad AI.

docs/VMR-Product-Versioning.md
mmitche: If they want to align with our build ID, then that's fine, though it won't matter whether they align.

docs/VMR-Product-Versioning.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Open questions
+
+There are a few open questions that may require some experience before settling on a final answer:
+
+### How should branding be updated? Centrally in VMR or individual repos?
+
+Branding has been historically updated in individual repos. With the VMR, we could now do branding centrally, checking in an update in a single commit.
+
+**Proposal: Update branding centrally in the VMR and backflow to repos**
```
ViktorHofer, line 152: I assume you describe an eventual goal here given that Versions.props changes (which contain the branding information) can't be made in the VMR today with the current flow mechanism?

docs/VMR-Product-Versioning.md
```diff
@@ -0,0 +1,164 @@
...
+
+In the past, we would activate stable branding for a repository by changing its Versions.props. All future builds of that branch would be stable. The VMR now has options to queue a build with stable branding. Does stable branding need to be checked in any longer? Or can it just be activated on-demand with a switch?
+
+**Proposal: Check-in stable branding in preview and servicing branches. This ensures that rolling CI is producing the stable bits required for release at all times.**
+
+### If checked in, how should stable branding be activated?
+
+There are options in the VMR build pipeline to activate stable branding (preview or rtm release branding) via pipeline parameter. In addition, the stable switches could also be activated in the VMR repo-project Directory.Build.props, or individual repo Versions.props.
+
+**Proposal: Use YAML pre-processor directives to change defaults for stable branding to produce 'prerelease stable' in release/preview.N or release/rc.N, and 'rtm stable' in release/N, and non-stable everywhere else.**
```
ViktorHofer, line 164: When checking out the release/10.0.1xx branch in the VMR let's say in November, I won't need to pass in an additional arg to produce stable branded bits or would I? And would that only get enabled when doing a local OfficialBuildId build?

mmitche: You would. The CI would pass -branding rtm by default.

docs/VMR-Product-Versioning.md
mmitche: Correct.

mmitche:
Any more feedback here?

PR #1093, opened 2025-06-10
Update to CentOS Stream 10

ellahathaway:
Related to https://github.com/dotnet/source-build/issues/4890

[Test pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2727363&view=results)

### DIFF
--- a/eng/pipelines/templates/jobs/vmr-build.yml
+++ b/eng/pipelines/templates/jobs/vmr-build.yml
@@ -465,9 +465,7 @@ jobs:
           fi
 
           if [[ '${{ parameters.withPreviousSDK }}' == 'True' ]]; then
-            # Source-built artifacts are from CentOS 9 Stream. We want to download them without
-            # downloading portable versions from the internet.
-            # Update previous comment on next rebootstrap or release - https://github.com/dotnet/source-build/issues/5238
+            # Using previous SDK implies we do not want to bootstrap.
             customPrepArgs="${customPrepArgs} --no-sdk --no-bootstrap"
             prepSdk=false
           elif [[ '${{ length(parameters.reuseBuildArtifactsFrom) }}' -gt '0' ]]; then


ellahathaway:
Two aspnetcore issues in the CentOSStream10_Online_PreviousSourceBuiltSdk_x64 and CentOSStream10_Offline_PreviousSourceBuiltSdk_x64 legs:

```
warning: Building of NodeJS projects is disabled since node is not detected on Path and no BuildNodeJs or NoBuildNodeJs setting is set explicitly.
  warning: Some managed projects depend on NodeJS projects. Building NodeJS is disabled so the managed projects will fallback to using the output from previous builds. The output may not be correct or up to date.
```

and

```
##[error]src/aspnetcore/src/Framework/App.Runtime/src/Microsoft.AspNetCore.App.Runtime.sfxproj(0,0): error NU1102: (NETCORE_ENGINEERING_TELEMETRY=Restore) Unable to find package Microsoft.NETCore.App.Crossgen2.linux-x64 with version (= 10.0.0-preview.6.25310.111)
```

MichaelSimons:
> Two aspnetcore issues in the CentOSStream10_Online_PreviousSourceBuiltSdk_x64 and CentOSStream10_Offline_PreviousSourceBuiltSdk_x64 legs:

These legs can't be updated to 10 until you update PSB artifacts to be based on CentOSStream10

ellahathaway:
Addressed issue in https://github.com/dotnet/dotnet/pull/1093/commits/8a1966520c58ce843f4cdad7b45ef6194e936121. I think we should keep the new "previous" variables to make the next centos update easier. Thoughts? @MichaelSimons 

[Test pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2728086&view=results)

mthalman (approved):

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -467,6 +467,7 @@ jobs:
           if [[ '${{ parameters.withPreviousSDK }}' == 'True' ]]; then
             # Source-built artifacts are from CentOS 9 Stream. We want to download them without
```
MichaelSimons: I think this comment could be abstracted to not contain CentOS Stream 9.  

`Using previous SDK implies we do not want to bootstrap`

ellahathaway:
> I can see value in this yet it is unfortunate to have unreferenced variables. 

Agreed. I'd rather not have the unreferenced variables, so I'll remove them entirely during the next rebootstrap/release.

eng/pipelines/templates/jobs/vmr-build.yml
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1093/commits/12c9fccc3192d5bbf9f98ea15cedcdcf55334f89

Issue #1079, opened 2025-06-09
Installer tests are failing due to project dependency issues


NikolaMilosavljevic:
Dependencies of `scenario-tests` project are being built, despite: https://github.com/dotnet/dotnet/blob/4b3afe1cb9d8ba910a7de6661b1635491a1af276/test/Microsoft.DotNet.Installer.Tests/Microsoft.DotNet.Installer.Tests.csproj#L16 and https://github.com/dotnet/dotnet/blob/4b3afe1cb9d8ba910a7de6661b1635491a1af276/repo-projects/scenario-tests.proj#L21

Last successful build was on 6/3: https://dev.azure.com/dnceng/internal/_build/results?buildId=2722680

NikolaMilosavljevic:
Here's the diff containing the breaking change: https://github.com/dotnet/dotnet/compare/a08e7c8..62e9d32

Most likely root cause: https://github.com/dotnet/dotnet/pull/852

@ViktorHofer

ViktorHofer:
Thanks for failing. Yes, this regressed with https://github.com/dotnet/dotnet/pull/852. Submitted https://github.com/dotnet/dotnet/pull/1104 to collect binlogs. Hard to tell what's going on without a binlog available.

PR #1105, opened 2025-06-11
Disable transitive P2Ps when building scenario-tests standalone

ViktorHofer:
Fixes https://github.com/dotnet/dotnet/issues/1079

Microsoft.DotNet.Installer.Tests.csproj which is invoked by the vmr-validate-installers.yml references scenario-tests.proj and passes an additional global property in: SkipScenarioTestsDependencies=true.

NuGet intentionally drops additional properties from P2Ps to flatten the restore graph. Due to that, transitive P2Ps which are calculated and then provided by NuGet still flow into scenario-tests.proj.

Explicitly disable transitive P2Ps when SkipScenarioTestsDependencies=true.

### DIFF
--- a/repo-projects/scenario-tests.proj
+++ b/repo-projects/scenario-tests.proj
@@ -15,6 +15,7 @@
          as the scenario tests repo needs to be capable of being built by itself in a standalone validation job.
          This also requires the use of bootstrap arcade. -->
     <SkipScenarioTestsDependencies Condition="'$(DotNetSourceOnlyTestOnly)' == 'true'">true</SkipScenarioTestsDependencies>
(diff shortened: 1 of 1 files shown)


akoeplinger (approved):

NikolaMilosavljevic:
Thanks for fixing this. I was going to make a change to pass `SkipScenarioTestsDependencies` property in the build invocation command. It would have also worked, but your solution is scoped to scenario-tests project which is better.

ViktorHofer:
>  I was going to make a change to pass SkipScenarioTestsDependencies property in the build invocation command. It would have also worked, but your solution is scoped to scenario-tests project which is better.

Yes that would have been an option as well but wouldn't have worked locally unless you supply that arg there as well.

Issue #1118, opened 2025-06-11
Add schedule official builds that sign on Sat/Sun mornings


mmitche:
Add a schedule for Saturday and Sunday morning (UTC) to ci.yml. Then, update the vmr build variable template so that if the trigger method is scheduled, that signing is enabled by default.

PR #1119, opened 2025-06-11
Add signed main build on weekends

akoeplinger:
Fixes https://github.com/dotnet/dotnet/issues/1118

### DIFF
--- a/eng/pipelines/ci.yml
+++ b/eng/pipelines/ci.yml
@@ -145,7 +145,7 @@ extends:
 
     - template: /eng/pipelines/templates/stages/vmr-build.yml@self
       parameters:
-        ${{ if eq(variables.isScheduleTrigger, 'true') }}:
+        ${{ if and(eq(variables.isScheduleTrigger, 'true'), eq(variables['Build.CronSchedule.DisplayName'], 'Signed main build on weekends') }}:
           desiredSigning: Signed
         ${{ else }}:
           desiredSigning: ${{ parameters.desiredSigning }}
--- a/eng/pipelines/ci.yml
+++ b/eng/pipelines/ci.yml
@@ -10,6 +10,14 @@ trigger:
       - release/*
       - internal/release/*
 
+schedules:
+  - cron: "0 12 * * 0,6" # run every Saturday and Sunday at 12:00 (UTC)
+    displayName: Signed main build on weekends
+    branches:
+      include:
+      - main
+    always: true
+
 parameters:
 - name: desiredSigning
   displayName: 'Sign?'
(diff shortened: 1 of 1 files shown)


eng/pipelines/ci.yml
```diff
@@ -137,11 +145,12 @@ extends:
 
     - template: /eng/pipelines/templates/stages/vmr-build.yml@self
       parameters:
-        desiredSigning: ${{ parameters.desiredSigning }}
-        desiredIBC: ${{ parameters.desiredIBC }}
         ${{ if eq(variables.isScheduleTrigger, 'true') }}:
-          scope: full
-        ${{ elseif eq(variables.isPRTrigger, 'true') }}:
+          desiredSigning: Signed
```
mmitche, line 149: I would put this logic into the vmr-build variables file, where desiredSigning is interpreted

akoeplinger: hmm why? vmr-build "parses" the desiredSigning value, for me it makes more sense to have setting the value here.

mmitche: I was thinking mainly because that template also deals with types of triggers too. Anyways, I think this is fine then.

mmitche (approved):

eng/pipelines/ci.yml
jkoritzinsky, line 149: Do we want to condition this on the specific "signed build on weekends" schedule name in case we add other scheduled builds in the future?

ViktorHofer: Good idea. We will also eventually be adding a scheduled stage2 windows build.

I wonder whether separate pipelines for these kind of things would make sense. Would it be possible to set pipeline variables in AzDo to automatically set desiredSigning=true with a separate pipeline?

akoeplinger: We can use `Build.CronSchedule.DisplayName` to check for the schedule trigger or we can also make a separate pipeline and check `Build.DefinitionName`.

I don't really have a preference, thoughts?

akoeplinger: Actually, I'm not sure if we can condition the `schedule` itself based on a variable so a separate pipeline might be harder (we'd probably need a separate .yml, or define the schedule in the AzDO UI override). I'm also not sure what would happen with the compliance tooling in a separate "official" pipeline.

I'll add the check for displayname here

ViktorHofer: Looks like setting the var in AzDO pipeline variables would work (to keep the YML simple):

image

akoeplinger: Setting the variable isn't the issue, the question is whether conditioning `schedule` based on it works

ViktorHofer: I was assuming that with a separate pipeline we have a separate thin YML that sets the schedule similar to what we have in runtime: https://github.com/dotnet/runtime/blob/main/eng/pipelines/coreclr/pgostress.yml

No need to figure this out now but we additional post merge validation (stage2, etc) we will need a better story here. I don't think overloading one YML with all possible entry points is a good solution long term.

akoeplinger: Yes with a separate YML it works of course. I think the other validation pipelines differ from this one in that they don't need to use the 1ES PT Official template and all that implies.

ViktorHofer (approved):

eng/pipelines/ci.yml
jkoritzinsky: I was thinking of conditioning in the cron display name so we don't end up with a billion pipelines we need to track.

jkoritzinsky (approved):

eng/pipelines/ci.yml
mmitche: Separate pipelines doesn't work because of potentially overlapping Official build IDs. We can really only have one official build pipeline without jumping through quite a lot of hoops.

mmitche (approved):

PR #1132, opened 2025-06-12
Refactor SBRP reference only package handling logic

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/5193

These changes separate the reference only packages (reference and target) produced in SBRP from the text-only packages.  Text-only packages are now treated like any other package.  Reference only packages are still copied to prereqs/package/reference just as before.  They are no longer included in the artifacts tarball.

To accomplish this, SBRP puts the reference only packages in `artifacts\packages\<config>\ReferenceOnly` so that they do not get picked up by publishing.  A new VMR target was added to copy the reference only packages to `prereqs\package\reference`.  The special publishing logic for SBRP was then removed so that the text-only packages are treated like any other package.


src/source-build-reference-packages/src/referencePackages/Directory.Build.props
```diff
@@ -34,6 +34,9 @@
     <LangVersion>latest</LangVersion>
 
     <CommonSrc>$([MSBuild]::NormalizeDirectory('$(MSBuildThisFileDirectory)', 'common'))</CommonSrc>
+
+    <IsShippingPackage>false</IsShippingPackage>
```
MichaelSimons, line 38: Not sure if this should be set or not.  It is not necessary but wanted to help call out these are "not shipping" artifacts.

repo-projects/Directory.Build.targets
```diff
@@ -25,7 +18,6 @@
     <!-- Trim the trailing slash as it breaks argument parsing on Windows if this is the last argument. -->
     <CommonArgs>$(CommonArgs) /p:SourceBuiltNonShippingPackagesDir=$(RepoArtifactsNonShippingPackagesDir.TrimEnd('\'))</CommonArgs>
     <CommonArgs>$(CommonArgs) /p:SourceBuiltPdbArtifactsDir=$(RepoArtifactsPdbArtifactsDir.TrimEnd('\'))</CommonArgs>
-    <CommonArgs>$(CommonArgs) /p:SourceBuiltNonShippingPackagesDir=$(RepoArtifactsNonShippingPackagesDir.TrimEnd('\'))</CommonArgs>
```
MichaelSimons, line 28: This appears to be a duplicate of the line 2 above.

MichaelSimons:
@mmitche, @ViktorHofer - would appreciate your feedback on this approach vs adding a special package concept to the manifest and publishing logic.

mthalman:
Stage 2 is failing:

```
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package NETStandard.Library. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package Microsoft.Win32.Registry. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package System.Memory. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package System.Runtime.CompilerServices.Unsafe. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package System.Security.Principal.Windows. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package System.Threading.Tasks.Extensions. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package System.Buffers. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package System.Security.Cryptography.Cng. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package System.Security.AccessControl. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
/__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj : error NU1101: Unable to find package Microsoft.Bcl.HashCode. No packages exist with this id in source(s): /__w/1/s/.dotnet/library-packs, /__w/1/s/prereqs/packages/prebuilt/, /__w/1/s/prereqs/packages/previously-source-built/
  Failed to restore /__w/1/s/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj (in 340 ms).
```

repo-projects/Directory.Build.targets
```diff
@@ -608,7 +598,7 @@
     </GetKnownArtifactsFromAssetManifests>
 
     <ItemGroup>
-      <ProducedPackageFromDependentVertical ReferenceOnly="$([MSBuild]::ValueOrDefault('$(ReferenceOnlyRepoArtifacts)', 'false'))" />
+      <ProducedPackageFromDependentVertical ReferenceOnly="false" />
```
mthalman: For both this item group and `ProducedPackage`, it seems you can just get rid of this `ReferenceOnly` metadata and then just update the logic which current consumes that metadata. That's in the `CreateBuildInputProps` target. So since it's always `false` now, the filtering that's done in that target can be removed.

repo-projects/Directory.Build.targets
```diff
@@ -683,6 +673,18 @@
...
+  <!-- Copy reference only packages to prereqs/package/reference -->
+  <Target Name="CopyReferenceOnlyPackages"
+          Condition="'$(ReferenceOnlyRepoArtifacts)' == 'true'">
+    <ItemGroup>
+      <_ReferenceOnlyPackages Include="$(RepoArtifactsDir)packages\$(Configuration)\ReferenceOnly\**\*" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(_ReferenceOnlyPackages)"
+          DestinationFolder="$(ReferencePackagesDir)"
+          Condition="'@(_ReferenceOnlyPackages)' != ''" />
```
mthalman: It might make sense to move this condition to be part of the target's condition.

repo-projects/Directory.Build.targets
```diff
@@ -683,6 +666,18 @@
...
+  </Target>
```
ViktorHofer: nit: consider moving this target into source-build-reference-packages.proj as it's specific to only that one project.

The target could have a Before/AfterTargets hook.

MichaelSimons: Thanks for the feedback.  That was something I was contemplating.

MichaelSimons: Fixed in https://github.com/dotnet/dotnet/pull/1132/commits/014e1d4fbac20bd6add445e9372f7323d6e89d7f

repo-projects/Directory.Build.targets
MichaelSimons: Addressed with https://github.com/dotnet/dotnet/pull/1132/commits/84212bfe721eb79bb4d5798bb3878bc6d337b667.

repo-projects/Directory.Build.targets
MichaelSimons: `_ReferenceOnlyPackages` is a local ItemGroup so it cannot be part of the target's condition.

MichaelSimons:
> Stage 2 is failing:

I knew this was going to be an issue but I thought it would have to be addressed during the next bootstrap.  I overlooked stage 2 builds.  I added a fix in https://github.com/dotnet/dotnet/pull/1132/commits/6222f3bd63f219725e45eb1c1fe9b6462215677d.  This will have to be cleaned up after the next bootstrap.

repo-projects/source-build-reference-packages.proj
```diff
@@ -17,4 +14,16 @@
...
+  <Target Name="CopyReferenceOnlyPackages"
+          AfterTargets="RepoBuild"
+          BeforeTargets="CleanupRepo" >
+    <ItemGroup>
+      <_ReferenceOnlyPackages Include="$(RepoArtifactsDir)packages\$(Configuration)\ReferenceOnly\**\*" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(_ReferenceOnlyPackages)"
+          DestinationFolder="$(ReferencePackagesDir)"
+          Condition="'@(_ReferenceOnlyPackages)' != ''" />
```
ViktorHofer: nit: is this condition necessary? I assume SBRP should always produce these

repo-projects/source-build-reference-packages.proj
```diff
@@ -17,4 +14,16 @@
     <BuiltSdkPackage Include="Microsoft.Build.Traversal" />
   </ItemGroup>
 
+  <!-- Copy reference only packages to prereqs/package/reference -->
+  <Target Name="CopyReferenceOnlyPackages"
+          AfterTargets="RepoBuild"
+          BeforeTargets="CleanupRepo" >
+    <ItemGroup>
+      <_ReferenceOnlyPackages Include="$(RepoArtifactsDir)packages\$(Configuration)\ReferenceOnly\**\*" />
```
ViktorHofer: nit: why the `**\*` glob? I would expect the ReferenceOnly dir to be a flat folder.

repo-projects/source-build-reference-packages.proj
```diff
@@ -17,4 +14,16 @@
...
+    </ItemGroup>
+
+    <Copy SourceFiles="@(_ReferenceOnlyPackages)"
+          DestinationFolder="$(ReferencePackagesDir)"
+          Condition="'@(_ReferenceOnlyPackages)' != ''" />
+  </Target>
 </Project>
```
ViktorHofer, line 42: style nit: we put an empty line before the project closing tag.

ViktorHofer (approved):

ViktorHofer (commented):
One slight annoyance is that the [`LogRepoArtifacts`](https://github.com/dotnet/dotnet/blob/0da37f99e59167516b19c39d38675ae87bef7348/repo-projects/Directory.Build.targets#L580) target doesn't log the reference-only packages anymore. You could add some indirection to log these as well in that target by adding to an item group.

repo-projects/source-build-reference-packages.proj
MichaelSimons: This was something that felt more apporpriate when the infra was generic.  Removed with https://github.com/dotnet/dotnet/pull/1132/commits/2c502e4c80567b5d119c3582c203c71eb0568cf1

repo-projects/source-build-reference-packages.proj
MichaelSimons: Fixed with https://github.com/dotnet/dotnet/pull/1132/commits/2c502e4c80567b5d119c3582c203c71eb0568cf1

repo-projects/source-build-reference-packages.proj
MichaelSimons: Removed with https://github.com/dotnet/dotnet/pull/1132/commits/2c502e4c80567b5d119c3582c203c71eb0568cf1

mthalman (approved):

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
```diff
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework Condition="'$(DotNetBuildSourceOnly)' != 'true'">netstandard2.0</TargetFramework>
+    <TargetFramework Condition="'$(DotNetBuildSourceOnly)' == 'true'">$(NetCurrent)</TargetFramework>
```
ViktorHofer: What is that about?

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard2.1</TargetFramework>
```
mmitche: Nit: Might want to put a comment here as to why ns2.1 is desired over other options.

MichaelSimons:
> One slight annoyance is that the [`LogRepoArtifacts`](https://github.com/dotnet/dotnet/blob/0da37f99e59167516b19c39d38675ae87bef7348/repo-projects/Directory.Build.targets#L580) target doesn't log the reference-only packages anymore. So it's not obvious whether everything got produced just from looking at the build log.
> 
> You could add some indirection to log these as well in that target by adding to an item group. Or probably better, introduce an additional target in source-build-reference-packages.proj that runs after LogRepoArtifacts and also lists the SBRP reference only artifacts that got produced.

This was addressed in https://github.com/dotnet/dotnet/pull/1132/commits/4cb0e6cbc31141a6bf495a3122694628f3e4e97d.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
MichaelSimons: Fixed in https://github.com/dotnet/dotnet/pull/1132/commits/a7e1403f66a053b2194cb0123fc490000c48fbb7

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
```diff
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <!-- Use a NetStandard TFM bundled in the SDK to avoid prebuilts in source builds. -->
+    <TargetFramework>netstandard2.1</TargetFramework>
```
MichaelSimons: This change is needed to support SB.  netstandard2.0 is not bundled with the SDK.  It is included as a SBRP target pack.  I would like to eliminate the SBRP reference package from PSB as it eliminates special logic to handle them and increases the PSB size.  NS2.1 is bundled with the SDK so upgrading solves these problems.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
MichaelSimons: This was a point in time change trying to flush out a solution that would work for SB.  See https://github.com/dotnet/dotnet/pull/1132/files#r2145480187 for details.

src/source-build-reference-packages/src/referencePackages/Directory.Build.props
mmitche: I think that all of SBRP is non-shipping? Maybe move this to the root SBRP D.B.props?

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
ViktorHofer: I think NS20 was chosen so that the msbuild sdk resolver works on both .NETCoreApp msbuild and desktop (VS) msbuild. If you upgrade beyond netstandard2.0 and with that lock out VS msbuild, you could also just go all up to NetCurrent.

mmitche (approved):

src/source-build-reference-packages/src/referencePackages/Directory.Build.props
MichaelSimons: No it's not.  Text-only packages are considered shipping as will be the SBE packages when we consolidate the repo.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver/Microsoft.DotNet.UnifiedBuild.MSBuildSdkResolver.csproj
MichaelSimons: It looks like NS20 is required for at least winforms because the resolver is loaded into the repo level msbuild process.  We can use NetCurrent for non-windows to support SB scenarios and NS20 on windows.

ViktorHofer: Any repo that uses desktop msbuild. aspnetcore falls into that bucket too. I don't think we can move away from netstandard2.0 unless we multi-target the resolver (net472 + NetCurrent) and use the right assembly based on the MSBuildRuntimeType.

MichaelSimons: I will update the comment.  If we start bundling the NS2.0 target pack in the SDK, we can eliminate this condition and go back to always using NS2.0

> I don't think we can move away from netstandard2.0 unless we multi-target the resolver (net472 + NetCurrent) and use the right assembly based on the MSBuildRuntimeType.

I'm not following the impact.  Are you alright with the proposed change or do you want net472 + NetCurrent for non-windows?

MichaelSimons: Updated comment in https://github.com/dotnet/dotnet/pull/1132/commits/5d50321d6e851dcfb9add612d63f1527d3171c10

MichaelSimons:
Validation passed, it would be good to take another look @mthalman, @mmitche, and @ViktorHofer.

ViktorHofer (approved):

PR #1133, opened 2025-06-12
Add explanation how to manually download an msi version

ViktorHofer:


docs/builds-table.md
```diff
@@ -21,6 +20,9 @@ Reference notes:
 > * [Host, Host FX Resolver, and Shared Framework](https://github.com/dotnet/runtime/blob/main/docs/project/dogfooding.md#nightly-builds-table)
 > * [ASP.NET Core Shared Framework](https://github.com/aspnet/AspNetCore/blob/main/docs/DailyBuilds.md)
 
+#### Manually construct a link to download an SDK msi (example for dotnet-sdk msi for win-x64):
```
akoeplinger, line 23: nit: it's a bit confusing that we call this an msi and then link to an .exe. maybe just call this SDK installer here

PR #1151, opened 2025-06-13
Retry all errors to handle curl (18)

Winniexu01:
Fixes: https://github.com/dotnet/source-build/issues/5240

curl 18 is CURLE_PARTIAL_FILE and doesn't count as a retry condition: https://curl.se/docs/manpage.html


prep-source-build.sh
```diff
@@ -175,7 +175,20 @@ function DownloadArchive {
...
+        if curl -f --retry 5 -O "$archiveUrl"; then
+          exit 0
+        else
+          case $? in
+            18)
+              sleep 3
+              ;;
+          esac
+        fi
+      done
```
ViktorHofer: If I read this correctly the loop applies to not just case 18 but any failure?

Winniexu01: Oh, yes, I will update it to apply only to case 18.

Winniexu01: Update: https://github.com/dotnet/dotnet/pull/1151/commits/9cd9bf4b988f339d31503c378ee8e3c94f4d032a

prep-source-build.sh
```diff
@@ -175,7 +175,23 @@ function DownloadArchive {
...
+            18)
+              sleep 3
+              ;;
+            *)
+              exit 1
+              ;;
+          esac
+        fi
+      done
+    )
```
akoeplinger: I think we could simplify this by passing `--retry-delay 3 --retry-all-errors`

Winniexu01: Yes, I saw this option. Since there was a note in its [description](https://curl.se/docs/manpage.html#--retry-all-errors): `there may be unintended consequences such as sending or receiving duplicate data.`, I decided to use a for loop to handle the curl (18). But since you brought it up, I'll update the code now.

Winniexu01: Update: https://github.com/dotnet/dotnet/pull/1151/commits/0d9dd007641082d6794e1fc751333b6eecf74481

akoeplinger: yeah the docs mention that only happens when you redirect/pipe the curl output. since we explicitly use the -O option to let curl write the file we're save and it will handle deleting existing files

akoeplinger (approved):

ViktorHofer:
I assume that `retry-all-errors` would retry even in case of a 404?

Winniexu01:
Yes, `--retry-all-errors` with `-f` will retry even in case of a 404.

ViktorHofer:
I didn't see any other cases that required a more sophisticated retry than for the curl 18 error. Not sure I like the sledgehammer approach here but I don't have a strong opinion.

akoeplinger:
Yeah it's a bit unfortunate, ideally we'd find out why the blob container download is unreliable, but I think this is ok too for this use case.

ViktorHofer:
This caused a regression:

>   Downloading source-built Artifacts from https://builds.dotnet.microsoft.com/source-built-artifacts/assets/Private.SourceBuilt.Artifacts.10.0.100-preview.6.25302.104.centos.9-x64.tar.gz...
curl: option --retry-all-errors: is unknown
curl: try 'curl --help' or 'curl --manual' for more information

https://dev.azure.com/dnceng/internal/_build/results?buildId=2730661&view=logs&j=391e3a46-1d0e-50f6-4b48-986aa9de4d73&t=62649f07-1bbf-51a6-0d31-d342655156c6

Curl might be too old. I would go back to the other approach that you had in this PR.

PR #1154, opened 2025-06-13
Update vmr-validate-installers.yml

ViktorHofer:
I at least temporarily need a binlog even when jobs time out to diagnose an issue further: https://dev.azure.com/dnceng/internal/_build/results?buildId=2729497&view=results


ViktorHofer:
Force merge as this path isn't entered in PR validation.

PR #1185, opened 2025-06-16
Don't use --retry-all-errors in prep-source-build.sh

akoeplinger:
Switch back to first version from https://github.com/dotnet/dotnet/pull/1151 as --retry-all-errors was only added in curl 7.71.0

Fixes https://github.com/dotnet/source-build/issues/5251


akoeplinger:
prep-source-build.sh passed in `VMR Source-Only Build CentOSStream10_Online_MsftSdk_x64`, fast merging.

PR #1189, opened 2025-06-16
Allow partial VMR verifications

NikolaMilosavljevic:
Many repo owners want to run VMR verifications in their repo PRs. For a lot of repos, doing the full set of verifications is too slow and resource intensive. This PR allows repos to select specific verification jobs.

For instance SBRP repo can simply specify source-build stage 1 job, using:
```
    verifications: [ "source-build-stage1" ]
```

Other repos, like WinForms, can elect to verify a single Windows leg, with:
```
    verifications: [ "unified-build-windows-x64" ]
```

For repos that prefer a full set of verifications, `verifications` parameter can be omitted.

To further simplify pipeline customizations on the repo side, another PR will be coming soon that will remove the existing pipeline template from arcade (https://github.com/dotnet/arcade/blob/main/eng/common/templates/vmr-build-pr.yml) and create a repo-specific template in the appropriate source location. Repo owners will be able to simply copy this template from whichever repo they choose and modify the list of verifications if necessary. Template will have the necessary documentation explaining the `verifications` parameter use cases.

This PR also enables creation of various VMR pipelines, one of them might be "vmr-lite" that would run just the verification jobs.

## Validations

Full VMR build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2729174

### Repo-level verifications

Full set of verification jobs: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1069116
Just source-build jobs: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1066740
One unified-build job: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1069248

### DIFF
--- a/eng/pipelines/templates/stages/source-build-stages.yml
+++ b/eng/pipelines/templates/stages/source-build-stages.yml
@@ -30,26 +30,27 @@ parameters:
   default: []
 
 stages:
-- template: source-build-and-validate.yml
-  parameters:
-    pool_Linux: ${{ parameters.pool_Linux }}
-    pool_LinuxArm64: ${{ parameters.pool_LinuxArm64 }}
-    useMicrosoftBuildAssetsForTests: ${{ parameters.useMicrosoftBuildAssetsForTests }}
-    isBuiltFromVmr: ${{ parameters.isBuiltFromVmr }}
-    
-    # Description of the source build legs to run.
-    # This is described here as a parameter to allow two separate stages to be produced from this list (one for building
-    # and one for validation). Because this template is not directly within a pipeline, stage, or job, we can't reference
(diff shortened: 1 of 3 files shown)


NikolaMilosavljevic:
@mmitche I don't think this is needed in Preview 6 branch - correct?

MichaelSimons:
>  I don't think this is needed in Preview 6 branch - correct?

No it isn't.

NikolaMilosavljevic:
I will have a follow up PR to fix a minor issue with running a repo pipeline in a CI, with a subset of verifications. The goal is to enable this for repo PRs, but CI runs should still be a valid option I think, as the pipeline runs as CI, automatically, when it gets created.

NikolaMilosavljevic:
@mmitche there is a follow-up PR to cover the CI scenario for the new verification pipelines - https://github.com/dotnet/dotnet/pull/1214

PR #1216, opened 2025-06-17
Split out the IBC repo name from the repository name

mmitche:
Allow for IBC drops to be accessed under a different repo name than the RepositoryName, which is used for other things like the manifest repo origin. Most repos align on these names. Msbuild does not


jkoritzinsky (approved):

ViktorHofer (approved):
LGTM

PR #1236, opened 2025-06-18
Migrate SBE packages to SBRP

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/5194
Fixes https://github.com/dotnet/source-build/issues/5256

1. Hooks up the SBRP externalPackages added with https://github.com/dotnet/source-build-reference-packages/pull/1273 to the build.
2. Disabled SBE from the build and remove references to it.  

To make this PR easier to review, I will make a separate follow-up PR to remove the actual SBE source after this PR is merged.  Also immediately after merging the SBE code flow subscriptions will be removed.

See individual commits for a smaller breakdown of the work.


src/source-build-reference-packages/Directory.Build.targets
```diff
@@ -11,7 +11,8 @@
     contains everything. It's generated when the package is initially deconstructed and tends to be
     unique per package, so there's no benefit to this Arcade util.
   -->
-  <Target Name="InitializeStandardNuspecProperties" />
+  <Target Name="InitializeStandardNuspecProperties"
+          Condition="'$(IsPackable)' == 'true'" />
```
MichaelSimons: I'm not in love with reusing IsPackable here...

ViktorHofer: Agreed. These targets are only needed when using manual nuspecs in NuGet Pack instead of just a normal dotnet pack. Might be cleaner to create a property to express that and set it to true in the necessary places...

ViktorHofer: Or you could separate this out into a separate .targets file and import it in the needed places.

src/source-build-reference-packages/eng/Build.props
```diff
@@ -9,10 +9,12 @@
...
-    <!-- The BuildInParallel="false" flag makes sure that the targeting packs get built first. -->
-    <ProjectToBuild Include="$(RepoRoot)src\targetPacks\ILsrc\**\*.csproj" BuildInParallel="false" />
-    <ProjectToBuild Include="$(RepoRoot)src\textOnlyPackages\src\*\*\*.csproj" />
-    <ProjectToBuild Include="$(RepoRoot)src\referencePackages\src\**\*.csproj" />
+    <!-- BuildSteps are defined to ensure the correct dependency order across package types. -->
+    <ProjectToBuild Include="$(RepoRoot)eng\tasks\**\*.csproj" BuildStep="infra" />
+    <ProjectToBuild Include="$(RepoRoot)src\targetPacks\ILsrc\**\*.csproj" BuildStep="targetPacks" />
+    <ProjectToBuild Include="$(RepoRoot)src\textOnlyPackages\src\*\*\*.csproj" BuildStep="generalPackages" />
+    <ProjectToBuild Include="$(RepoRoot)src\referencePackages\src\**\*.csproj" BuildStep="generalPackages" />
+    <ProjectToBuild Include="$(RepoRoot)src\externalPackages\projects\*.proj" BuildStep="externalPackages" />
```
ViktorHofer: This looks interesting. Can you explain in a bit more detail why this dependency order is needed? I.e. I would expect the task packages under eng\tasks to be automatically built when needed via a P2P. I thought I had allowed for that already in SBE.

src/source-build-reference-packages/Directory.Build.targets
MichaelSimons: Creating a separate targets was what I was thinking as well. I think I will go that route.  Thanks for the feedback.

src/source-build-reference-packages/eng/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/Microsoft.DotNet.SourceBuild.Tasks.XPlat.csproj
```diff
@@ -7,11 +7,13 @@
          msbuild task assembly is for build only purposes. -->
     <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
     <Deterministic>False</Deterministic>
+    <IsPackable>false</IsPackable>
```
ViktorHofer: That shouldn't be necessary. IsPackable defaults to false when using the Arcade SDK.

```suggestion
```

MichaelSimons: This is related to the root Directory.Build.targets running - will address per https://github.com/dotnet/dotnet/pull/1236/files/0a602d8bd1805133d90cd902c37dfe170c2523e0#r2159655149

src/source-build-reference-packages/Directory.Build.targets
MichaelSimons: Fixed with https://github.com/dotnet/dotnet/pull/1236/commits/3d934c0be0c8efaea8536b6221257ca3b1f81965

src/source-build-reference-packages/eng/tasks/Microsoft.DotNet.SourceBuild.Tasks.XPlat/Microsoft.DotNet.SourceBuild.Tasks.XPlat.csproj
MichaelSimons: Fixed with https://github.com/dotnet/dotnet/pull/1236/commits/3d934c0be0c8efaea8536b6221257ca3b1f81965

src/source-build-reference-packages/Directory.Build.props
```diff
@@ -16,6 +16,7 @@
     <CustomizationsSourceFile>$(CustomizationsPrefix).cs</CustomizationsSourceFile>
 
     <ArtifactsReferenceOnlyPackagesDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsPackagesDir)','ReferenceOnly'))</ArtifactsReferenceOnlyPackagesDir>
+    <ManualNuspecTargets>$([MSBuild]::NormalizePath('$(RepositoryEngineeringDir)','ManualNuspec.targets'))</ManualNuspecTargets>
```
ViktorHofer: Super tiny nit. You don't need a function here as `RepositoryEngineeringDir` is already correctly normalized (It uses the NormalizeDirectory function in the arcade SDK).

```suggestion
    <ManualNuspecTargets>$(RepositoryEngineeringDir)ManualNuspec.targets</ManualNuspecTargets>
```

src/source-build-reference-packages/eng/Build.props
MichaelSimons: I found that I needed to specify the project, I can't explain why the P2P reference is not building in SBRP when it was in SBE.  I verified the [path](https://github.com/dotnet/dotnet/pull/1236/files#diff-86606b76f4b94a7136550fdd9a259189ce879b3fb2b238790422fc0b26b8fce6R72) is correct and the ProjectReference appears in the binlog.  Is it possibly related to the usage of BuildSteps?  I can share a binlog if you would like.

src/source-build-reference-packages/Directory.Build.props
MichaelSimons: Absolutely.  Silly oversight on my part.

src/source-build-reference-packages/eng/Build.props
ViktorHofer: Pushed a fix. You were missing the DependsOnTargets in the new target.

ViktorHofer: Regarding the other build steps here, I assume the external packages need to run after the referencePackages?

ViktorHofer (commented):
I love this. I think this is a great step towards simplification and unification. This will also enable easier cross-cutting changes between SBE and SBRP assets.

The one thing that I would advise to change is in Build.props not relying on the `BuildStep` metadata. I would rather set the `BuildInParallel=false` metadata on those items. Reason for that is that MSBuild would target batch BuildStep items which makes things harder to debug. Also, perf-wise it shouldn't matter as the cost of running textOnlyPackages and referenceOnlyPackages in serial should be roughly equivalent with the separate MSBuild task invocations due to the target batching.

Alternatively, we could define these dependencies in the tree (as P2Ps) but that would be a bigger change with implications on incremental builds so I would leave that as a follow-up to explore.

src/source-build-reference-packages/src/externalPackages/projects/Directory.Build.props
```diff
@@ -4,19 +4,27 @@
...
     <MinimalConsoleLogOutput Condition="'$(MinimalConsoleLogOutput)' == ''">true</MinimalConsoleLogOutput>
     <RepoConsoleLogFile>$(ArtifactsLogRepoDir)$(InnerRepoName).log</RepoConsoleLogFile>
     <RedirectRepoOutputToLog Condition="'$(MinimalConsoleLogOutput)' == 'true'">&gt; $(RepoConsoleLogFile) 2&gt;&amp;1</RedirectRepoOutputToLog>
+
+    <NuGetConfigFile>$(BaseIntermediateOutputPath)NuGet.config</NuGetConfigFile>
+
+    <XPlatSourceBuildTasksName>Microsoft.DotNet.SourceBuild.Tasks.XPlat</XPlatSourceBuildTasksName>
+    <XPlatSourceBuildTasksAssembly>$([MSBuild]::NormalizePath('$(ArtifactsBinDir)', '$(XPlatSourceBuildTasksName)', '$(Configuration)', '$(NetCurrent)', '$(XPlatSourceBuildTasksName).dll'))</XPlatSourceBuildTasksAssembly>
+
+    <BuildInParallel Condition="'$(BuildInParallel)' == ''">false</BuildInParallel>
```
mthalman: Is this going to be needed since you're making use of `BuildStep` for `ProjectToBuild` or is that not relevant?

MichaelSimons:
> I love this. I think this is a great step towards simplification and unification. This will also enable easier cross-cutting changes between SBE and SBRP assets.
> 
> The one thing that I would advise to change is in Build.props not relying on the `BuildStep` metadata. I would rather set the `BuildInParallel=false` metadata on those items. Reason for that is that MSBuild would target batch BuildStep items which makes things harder to debug. Also, perf-wise it shouldn't matter as the cost of running textOnlyPackages and referenceOnlyPackages in serial should be roughly equivalent with the separate MSBuild task invocations due to the target batching.
> 
> Alternatively, we could define these dependencies in the tree (as P2Ps) but that would be a bigger change with implications on incremental builds so I would leave that as a follow-up to explore.

Thanks for the feedback @ViktorHofer, I didn't realize the downside to utilizing `BuildStep`.  I replaced the usage with `BuildInParallel=false` as you suggested - https://github.com/dotnet/dotnet/pull/1236/files#diff-07d7974c36a473747cac37379f4e59b40a9c476c60539d1c68467811d0c1b303.

src/source-build-reference-packages/src/externalPackages/projects/Directory.Build.props
MichaelSimons: I don't think this is relevant anymore - removed.

mthalman (approved):

src/source-build-reference-packages/eng/Build.props
```diff
@@ -9,10 +9,11 @@
...
 
   <ItemGroup Condition="'$(GeneratePackageSource)' != 'true' and '$(Test)' != 'true'">
-    <!-- The BuildInParallel="false" flag makes sure that the targeting packs get built first. -->
+    <!-- The BuildInParallel="false" flag makes sure the correct dependency order between package types. -->
     <ProjectToBuild Include="$(RepoRoot)src\targetPacks\ILsrc\**\*.csproj" BuildInParallel="false" />
-    <ProjectToBuild Include="$(RepoRoot)src\textOnlyPackages\src\*\*\*.csproj" />
-    <ProjectToBuild Include="$(RepoRoot)src\referencePackages\src\**\*.csproj" />
+    <ProjectToBuild Include="$(RepoRoot)src\textOnlyPackages\src\*\*\*.csproj" BuildInParallel="false" />
+    <ProjectToBuild Include="$(RepoRoot)src\referencePackages\src\**\*.csproj" BuildInParallel="false" />
+    <ProjectToBuild Include="$(RepoRoot)src\externalPackages\projects\*.proj" />
```
ViktorHofer: Oh I didn't realize the globs here. This will mean that all the projects in each line here will build in serial in contrast to what I was stating before. I don't think we want that as that would have a big impact on perf.

@MichaelSimons sorry for causing the confusion here. You can either go back to the previous solution and we can follow-up post merge or you can add traversal projects for each type:

```
src\targetPacks\targetPacks.proj (uses Microsoft.Build.Traversal and references all projects underneath)
src\textOnlyPackages\textOnlyPackages.proj
src\referencePackages\referencePackages.proj
src\externalPackages\externalPackages.proj
```

I guess this makes more sense to do in a follow-up.

MichaelSimons: I reverted back to use BuildSteps as it preserves the performance and is functionally correct.  We can follow-up post merge.

MichaelSimons: I logged the following Epic sub-issue to track this - https://github.com/dotnet/source-build/issues/5271

ViktorHofer (approved):

PR #1270, opened 2025-06-20
Add PR pipeline stage to validate VMR changes

adamzip:
https://github.com/dotnet/arcade-services/issues/5155

### DIFF
--- a/eng/tools/ChangeValidation/Validation.cs
+++ b/eng/tools/ChangeValidation/Validation.cs
@@ -98,7 +98,7 @@ internal static class Validation
 
         string mergeBaseCommit = (await pm.ExecuteGit(repoPath, ["merge-base", "pr-head", targetBranch])).StandardOutput.Trim();
 
-        var diffOutput = (await pm.ExecuteGit(repoPath, ["diff", "--name-only", mergeBaseCommit, "HEAD"])).StandardOutput.Trim();
+        var diffOutput = (await pm.ExecuteGit(repoPath, ["diff", "--name-only", mergeBaseCommit, "pr-head"])).StandardOutput.Trim();
 
         var changedFiles = diffOutput
             .Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries)
@@ -106,7 +106,7 @@ internal static class Validation
             .ToImmutableList();
 
         Console.WriteLine($"Merge base commit is {mergeBaseCommit}");
-        Console.WriteLine($"Found modifications to {changedFiles.Count} file(s) in PR base branch");
+        Console.WriteLine($"Found modifications to {changedFiles.Count} file(s) in PR head branch");
 
         return new PrInfo(targetBranch, changedFiles);
     }


premun:
@adamzip as we spoke earlier - it would probably be good to comment out the other stages so that we don't block compute for no reason (and since there are iterations to be expected)

jkoritzinsky (commented):
Like with the other change validation tooling, I think this could be better written as a one or two line script in a bash script.

premun:
@jkoritzinsky please note that there will be a whole suite of checks we will run and many will use existing functionality from Maestro to parse `source-manifest` and `source-mappings`: https://github.com/dotnet/arcade-services/issues/2950

For this reason, C# will be more tenable long-term. Existing GHA checks from Ella will be ported to C# too.

.github/workflows/protected-files-validation.yml
```diff
@@ -1,22 +0,0 @@
-name: 'Check Protected Files for Changes'
```
premun, line 1: This is probably not desired? Or is it one of the tests?

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,194 @@
...
+            catch (Exception)
+            {
+                toolingMessages = new List<ProcessingMessage>();
+                AddProcessingMessage(processingMessages, Error("Error: Failed to perform validation on tooling files."));
+            }
+            processingMessages.AddRange(toolingMessages);
+
+            // Run Submodule Validation
+            Console.WriteLine("");
+            Console.WriteLine("");
```
premun:
```suggestion
            Console.WriteLine();
            Console.WriteLine();
```

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,194 @@
...
+            Console.WriteLine("2. Starting Submodule Validation...");
```
premun: Shouldn't this be rather logged from within the method itself?

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,194 @@
...
+            {
+                submoduleMessages = new List<ProcessingMessage>();
+                AddProcessingMessage(processingMessages, Error("Error: Failed to perform validation on submodule files."));
+            }
+            processingMessages.AddRange(submoduleMessages);
+
+            // Run Exclusion File Validation
+            Console.WriteLine("");
+            Console.WriteLine("");
+            Console.WriteLine("3. Starting Exclusion File Validation...");
```
premun: The repetition of these "sections" is not great. Especially since when we will be defining more of these.

I'd imagine we define a list of validations and then have code to loop through them and run/output them.
The list could just be a list of types and the validations could inherit from an interface/base class so that we just run some `Validate()` method on each one. We can instantiate each from DI.

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,194 @@
...
+
+    internal static void AddProcessingMessage(List<ProcessingMessage> processingMessages, ProcessingMessage processingMessage)
+    {
+        var lines = processingMessage.Message.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+
+        if (processingMessage.WarningLevel == WarningLevel.Error)
+        {
+            foreach (var line in lines)
+            {
+                Console.WriteLine($"##vso[task.logissue type=error] {line}");
```
premun: Looking at this page (where devs usually go first): https://github.com/dotnet/dotnet/pull/1270/checks?check_run_id=46502889155

we should consider which lines should be outputted with `##error` only to make it more readable

eng/pipelines/pr.yml
```diff
@@ -46,3 +46,27 @@ stages:
       scope: lite
     ${{ else }}:
       scope: full
+
+- stage: ValidateSubmodules
+  displayName: 'Validate VMR Submodules'
```
premun: nit: quotation not necessary in YAML (for most cases)

```suggestion
  displayName: Validate VMR Submodules
```

eng/tools/ChangeValidation/ChangeValidation.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net8.0</TargetFramework>
```
premun: This should probably be `$NetCurrent`

```suggestion
        <TargetFramework>$(NetCurrent)</TargetFramework>
```

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,180 @@
...
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json;
+
+using static ValidateVmrChanges.Validate;
+
+namespace ValidateVmrChanges;
+
+internal class ExclusionFileValidation
+{
+    private static readonly string SourceMappingsPath = "src/source-mappings.json";
```
premun: This should come from DarcLib's constants

eng/tools/ChangeValidation/SubmoduleValidation.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    // <summary>
+    // Verifies that the PR does not include any changes to submodule files
+    // </summary>
+    internal static List<ProcessingMessage> VerifySubModuleFileChanges(List<string> fileNames)
+    {
+        List<ProcessingMessage> messages = [];
+
+        var submoduleChanges = fileNames
+            .Where(f => f.Equals(".gitmodules", StringComparison.OrdinalIgnoreCase) || f.Contains("submodules/") || f.EndsWith("/.gitmodules"))
```
premun: This check is wrong, there's no rule about submodules being in the `submodules` directory:
- We should verify no submodules are added - which means changes to `.gitmodules` (so the first part of the condition)
- We should use DarcLib to read `source-manifest.json` and check the `submodules` list there to see if any files of the inlined submodules have been changed.

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,180 @@
...
+            if (jsonRoot.TryGetValue("mappings", out JToken? mappingsToken) && mappingsToken?.Type == JTokenType.Array)
+            {
+                mappings = (JArray)mappingsToken;
+            }
+            else
+            {
+                Console.WriteLine($"Cannot find mappings in {SourceMappingsPath}");
+                return [];
+            }
+        }
```
premun: We could just parse the file using DarcLib into an existing POCOs with an existing schema to be more type safe. This whole method could be a 2 liner then.

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,194 @@
...
+        else
+        {
+            Console.WriteLine($"VMR Change Validation completed successfully with {warnings.Count} warning(s).");
+            return 0;
+        }
+    }
+
+
+
+    internal static string RunGitCommand(string arguments)
```
premun: It might be better to re-use `ProcessManager` since we will need DarcLib anyway

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        if (process.ExitCode != 0)
+        {
+            throw new Exception($"Git command failed: {arguments}\n{err}");
+        }
+
+        return output;
+    }
+
+    internal static void AddProcessingMessage(List<ProcessingMessage> processingMessages, ProcessingMessage processingMessage)
```
premun: Are we buffering the messages just for the sake of counting them at the end or is there another reason?

I think it'd be just better to output them right away and if we need the count, increment some counter then?

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        if (originalExclusionRules == null || !originalExclusionRules.Any())
+        {
+            AddProcessingMessage(messages, Error($"No exclusion rules found in `{SourceMappingsPath}` on the target branch, or the file does not exist."));
+            return messages;
+        }
+
+        var newExclusionRules = GetExclusionPatternsFromBranch("HEAD");
+
+        if (newExclusionRules == null || !newExclusionRules.Any())
```
premun: It's entirely possible we will not have any exclusions and that is a valid scenario

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,180 @@
...
+        {
+            var gitPath = Path.Combine(dir.FullName, ".git");
+            if (Directory.Exists(gitPath) || File.Exists(gitPath))
+            {
+                return dir.FullName;
+            }
+            dir = dir.Parent;
+        }
+        throw new InvalidOperationException("Could not find the Git repository root.");
+    }
```
premun: This method is probably also available in Darclib

.github/workflows/protected-files-validation.yml
adamzip: This is the github workflow that checks the tooling files that the VMR change validation now checks instead

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    internal ExclusionFileValidation()
+    {
+        _processManager = new ProcessManager(NullLogger<ProcessManager>.Instance, "git");
+        VmrInfo vmrInfo = new VmrInfo(_processManager.FindGitRoot(""), "");
+        ILogger<VmrDependencyTracker> nullLogger = NullLogger<VmrDependencyTracker>.Instance;
+        SourceMappingParser parser = new SourceMappingParser(vmrInfo, new FileSystem());
+        SourceManifest sourceManifest = new SourceManifest([], []);
+        _dependencyTracker = new VmrDependencyTracker(vmrInfo, new FileSystem(), parser, sourceManifest, nullLogger);
+        _repoRoot = _processManager.FindGitRoot(AppContext.BaseDirectory);
```
premun: Please create a DI container instead so that we don't need to create these manually in every validation.
Then instantiate the Validations from the DI container.

Directory.Packages.props
```diff
@@ -8,6 +8,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageVersion Include="Microsoft.DotNet.DarcLib" Version="$(MicrosoftDotNetDarcLibVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="$(MicrosoftExtensionsHostingVersion)" />
```
premun: I think `Microsoft.Extensions.Hosting` should go lower where the rest of the extensions are and DarcLib should get the `<!-- arcade-services dependencies -->` header

eng/tools/ChangeValidation/ChangeValidation.csproj
```diff
@@ -0,0 +1,17 @@
...
+        <Nullable>enable</Nullable>
+        <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging" />
+    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" />
+    <PackageReference Include="Microsoft.DotNet.DarcLib" />
+    <PackageReference Include="Newtonsoft.Json" />
+    <PackageReference Include="Microsoft.Extensions.Hosting" />
```
premun: nit: Usually we order these alphabetically

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,123 @@
...
+            .Where(file => newExcludedFiles.Contains(NormalizePath(file)))
+            .ToList();
+
+        var filesMatchingNewExclusionRules = newExcludedFiles
+            .Where(file => !originalExcludedFiles.Contains(NormalizePath(file)) && !prInfo.ChangedFiles.Contains(NormalizePath(file)))
+            .ToList();
+
+        foreach (var file in excludedFilesInPr.Take(maxDisplayedFiles).ToList())
+        {
+            LogError($"This PR modifies the file `{file}`, which is part of the excluded files defined in {VmrInfo.DefaultRelativeSourceMappingsPath}. If these changes are necessary, please contact the repository organizers.");
```
premun:
```suggestion
            LogError($"This PR modifies the file `{file}`, which is part of the excluded files defined in {VmrInfo.DefaultRelativeSourceMappingsPath}. If these changes are necessary, please contact @dotnet/product-construction");
```

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        var excludedFilesInPr = prInfo.ChangedFiles
+            .Where(file => newExcludedFiles.Contains(NormalizePath(file)))
+            .ToList();
+
+        var filesMatchingNewExclusionRules = newExcludedFiles
+            .Where(file => !originalExcludedFiles.Contains(NormalizePath(file)) && !prInfo.ChangedFiles.Contains(NormalizePath(file)))
+            .ToList();
+
+        foreach (var file in excludedFilesInPr.Take(maxDisplayedFiles).ToList())
```
premun, line 61: It would be better to have a header saying "These files are in excluded locations (or something)":
- file 1
- file 2
- ...

Rather than having this long sentence 20 times in a row where it's hard to fish out the file names.

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        if (excludedFilesInPr.Count > maxDisplayedFiles)
+        {
+            LogError($"... {excludedFilesInPr.Count - maxDisplayedFiles} more excluded files detected in the PR. Only showing the first {maxDisplayedFiles} files.");
+        }
+
+        foreach (var file in filesMatchingNewExclusionRules.Take(maxDisplayedFiles).ToList())
+        {
+            LogError($"The new exclusion rules defined in {VmrInfo.DefaultRelativeSourceMappingsPath} include the VMR file `{file}`. If this file is not needed in the VMR, consider deleting it as part of the PR.");
```
premun: Same, consider adding a header and a plain list of files below

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        string originalBranch = (await _processManager.ExecuteGit(_repoRoot, "rev-parse", "abbrev-ref HEAD")).StandardOutput;
+        try
+        {
+            await _processManager.ExecuteGit(_repoRoot, "checkout", branchName);
+            await _dependencyTracker.RefreshMetadata(_repoRoot);
+            var sourceMappings = _dependencyTracker.Mappings;
+
+            return sourceMappings.Select(mapping => mapping.Exclude)
+                .SelectMany(exclude => exclude)
```
premun: This can be done directly:

```suggestion
            return sourceMappings
                .SelectMany(mapping => mapping.Exclude)
```

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,123 @@
...
+                .Distinct()
+                .ToList();
+        }
+        finally
+        {
+            await _processManager.ExecuteGit(_repoRoot, $"checkout {originalBranch}");
```
premun: I think this might be quoted wrong:

```suggestion
            await _processManager.ExecuteGit(_repoRoot, ["checkout", originalBranch]);
```

eng/tools/ChangeValidation/SubmoduleValidation.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        var directory = new InMemoryDirectoryInfo("", prInfo.ChangedFiles);
+
+        var result = matcher.Execute(directory);
+
+        var modifiedSubmoduleFiles = result.Files.Select(f => f.Path).ToList();
+
+        foreach(var file in modifiedSubmoduleFiles.Take(maxDisplayedFiles).ToList())
+        {
+            LogError($"This PR modifies the submodule file `{file}`, which is not allowed. Submodule files are managed by the repository maintainers and should not be modified directly.");
```
premun, line 64: Same - header + plain list preferrable

eng/tools/ChangeValidation/SubmoduleValidation.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        if (modifiedSubmoduleFiles.Any())
+        {
+            return false;
+        }
+        else
+        {
+            return true;
+        }
```
premun:
```suggestion
        return !modifiedSubmoduleFiles.Any();
```

eng/pipelines/pr.yml
```diff
@@ -46,3 +46,28 @@ stages:
...
+
+- stage: ValidateUserChanges
+  displayName: 'Validate user changes in VMR'
+  dependsOn: []
+  condition: and(eq(variables['isPRTrigger'], 'true'), ne(variables['Build.RequestedForEmail'], 'maestro-dotnet-bot@microsoft.com'))
+  jobs:
+  - job: Validate
+    displayName: 'Run Validation Script'
+    pool:
+      vmImage: 'windows-latest'
```
premun: nit quoting not necessary in YAML

```suggestion
  displayName: Validate user changes in VMR
  dependsOn: []
  condition: and(eq(variables['isPRTrigger'], 'true'), ne(variables['Build.RequestedForEmail'], 'maestro-dotnet-bot@microsoft.com'))
  jobs:
  - job: Validate
    displayName: Run Validation Script
    pool:
      vmImage: windows-latest
```

eng/pipelines/pr.yml
```diff
@@ -46,3 +46,28 @@ stages:
...
+        Write-Host "Fetching target branch: $(System.PullRequest.TargetBranch)"
+        git fetch origin $(System.PullRequest.TargetBranch)
+
+        . "$(Build.SourcesDirectory)\eng\common\tools.ps1"
+        InitializeDotNetCli -install $true
+
+        & "$env:DOTNET_INSTALL_DIR\dotnet.exe" run --project eng/tools/ChangeValidation/ChangeValidation.csproj
+      displayName: 'Run ChangeValidation tool'
+      env:
+        SYSTEM_PULLREQUEST_TARGETBRANCH: $(System.PullRequest.TargetBranch)
```
premun: I'd maybe split it into three build steps so it's immediately obvious what failed (fetch, tooling init, tool run)

eng/tools/ChangeValidation/PrInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+internal class PrInfo
+{
+    internal string TargetBranch { get; }
+    internal List<string> ChangedFiles { get; }
+    internal PrInfo(string targetBranch, List<string> changedFiles)
+    {
+        TargetBranch = targetBranch;
+        ChangedFiles = changedFiles;
+    }
+}
```
premun: Consider using records: `internal record PrInfo(string TargetBranch, List<string> ChangedFiles);` is equal to this whole class.

eng/tools/ChangeValidation/PrInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System;
+using System.Linq;
+
+namespace ValidateVmrChanges;
+
+internal class PrInfo
+{
+    internal string TargetBranch { get; }
+    internal List<string> ChangedFiles { get; }
```
premun: I'd make `ChangedFiles` a read-only type

eng/tools/ChangeValidation/ToolingFilesValidation.cs
```diff
@@ -0,0 +1,45 @@
...
+
+internal class ToolingFilesValidation : IValidationStep
+{
+    public string DisplayName => "Tooling Files Validation";
+
+    private static readonly string[] ToolingFilePatterns = new[]
+    {
+        "^eng/Version\\.Details\\.xml$",
+        "^src/source-manifest\\.json$",
+        "^eng/Version\\.props$",
```
premun, line 18:
```suggestion
        "^eng/Version\\.props$",
        "^eng/Version\\.Details.props$",
```

eng/tools/ChangeValidation/ExclusionFileValidation.cs
premun: Looking at this further, seems like all Validations have a similar behaviour - maybe we could have a validation base class to which you give a header and the list of offending files (maybe the max display size too) so we deduplicate this a bit?

adamzip: I didn't find a way to have multi-line error messages that work on ADO and GH.
When I tried to make a multiline error message in ADO, only the first line shows up as an error in red text, and when I preceded every line with ##vso[task.logissue type=error], then it showed up as separate errors on the GH PR. That's why I chose to make 1 standalone error per file. Is there a better solution?

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,152 @@
...
+            {
+                var pm = sp.GetRequiredService<IProcessManager>();
+                var repoRoot = pm.FindGitRoot(AppContext.BaseDirectory);
+                return new VmrInfo(repoRoot, "tmp");
+            });
+            services.AddScoped<ISourceManifest>(sp => new SourceManifest([], []));
+            services.AddScoped<IVmrDependencyTracker, VmrDependencyTracker>();
+            services.AddTransient<ISourceMappingParser, SourceMappingParser>();
+            services.AddSingleton<IFileSystem, FileSystem>();
+
```
premun: As we spoke yesterday, I think this all should be replaced `AddSingleVmrSupport()`

Directory.Packages.props
```diff
@@ -30,6 +31,7 @@
     <PackageVersion Include="xunit.extensibility.core" Version="$(XUnitVersion)" />
     <PackageVersion Include="xunit.extensibility.execution" Version="$(XUnitVersion)" />
     <PackageVersion Include="Octokit" Version="$(OctokitVersion)" />
+    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="$(MicrosoftExtensionsHostingVersion)" />
```
premun: I think this belongs to the `runtime` dependencies.

eng/tools/ChangeValidation/ExclusionFileValidation.cs
premun: I see. In that case, leave as-is, this makes sense.

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            return sourceMappings.SelectMany(mapping => mapping.Exclude)
+                .Distinct()
+                .ToList();
+        }
+        finally
+        {
+            await _processManager.ExecuteGit(_repoRoot, ["checkout", originalBranch]);
+        }
+    }
```
premun: It'd be better to inject `ILocalGitRepoFactory`, create a repo handler and do `vmr.CheckoutAsync(originalBranch)` or `vmr.GetCheckedOutBranchAsync()`

eng/tools/ChangeValidation/ExclusionFileValidation.cs
```diff
@@ -0,0 +1,123 @@
...
+    
+    private HashSet<string> FindMatchingFiles(List<string> globPatterns)
+    {
+        var repoRoot = _processManager.FindGitRoot(AppContext.BaseDirectory);
```
premun: You could inject `IVmrInfo` and if you initialize the container well in the beginning, you can use `IVmrInfo.VmrPath` to get the path instead of looking for the root always.

eng/Versions.props
```diff
@@ -26,6 +26,7 @@
     -->
     <PrivateSourceBuiltSdkVersion>10.0.100-preview.7.25372.107</PrivateSourceBuiltSdkVersion>
     <PrivateSourceBuiltArtifactsVersion>10.0.100-preview.7.25372.107</PrivateSourceBuiltArtifactsVersion>
+    <MicrosoftDotNetDarcLibVersion>1.1.0-beta.25404.1</MicrosoftDotNetDarcLibVersion>
```
premun, line 31:
```suggestion
    <!-- arcade-services dependencies -->
    <MicrosoftDotNetDarcLibVersion>1.1.0-beta.25404.1</MicrosoftDotNetDarcLibVersion>
```

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        string? targetBranch = Environment.GetEnvironmentVariable("SYSTEM_PULLREQUEST_TARGETBRANCH") ?? "main";
+
+        if (string.IsNullOrEmpty(targetBranch))
+        {
+            throw new ArgumentException("Cannot determine PR target branch.");
+        }
+
+        await pm.ExecuteGit(repoPath, ["fetch", $"origin {targetBranch}"]);
+        //string diffOutput = (await pm.ExecuteGit(repoPath, ["diff",  $"--name-only" ,$"origin/{targetBranch}...HEAD"])).StandardOutput;
```
premun:
```suggestion
```

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        string mergeBase = (await pm.ExecuteGit(repoPath, ["merge-base", $"origin/{targetBranch}", "HEAD" ])).StandardOutput.Trim();
+        Console.WriteLine("Merge base commit:" + mergeBase);
+        string diffOutput = (await pm.ExecuteGit(repoPath, ["diff", "--name-only", mergeBase, "HEAD" ])).StandardOutput;
+
+        Console.WriteLine("diff output length:" +diffOutput.Length);
+
+        var changedFiles = diffOutput
+            .Split('\n', StringSplitOptions.RemoveEmptyEntries)
```
premun: Shouldn't we split by `Environment.NewLine`?

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        await pm.ExecuteGit(repoPath, ["fetch", $"origin {targetBranch}"]);
+        //string diffOutput = (await pm.ExecuteGit(repoPath, ["diff",  $"--name-only" ,$"origin/{targetBranch}...HEAD"])).StandardOutput;
+
+        string mergeBase = (await pm.ExecuteGit(repoPath, ["merge-base", $"origin/{targetBranch}", "HEAD" ])).StandardOutput.Trim();
+        Console.WriteLine("Merge base commit:" + mergeBase);
+        string diffOutput = (await pm.ExecuteGit(repoPath, ["diff", "--name-only", mergeBase, "HEAD" ])).StandardOutput;
+
+        Console.WriteLine("diff output length:" +diffOutput.Length);
```
premun: Is this useful to log? If anything, it'd rather be more useful to log how many files changed?

premun: Ah, that's below

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,135 @@
...
+        var changedFiles = diffOutput
+            .Split('\n', StringSplitOptions.RemoveEmptyEntries)
+            .Select(f => f.Replace('\\', '/').Trim())
+            .ToImmutableList();
+
+        Console.WriteLine("changed files count: " + changedFiles.Count);
+
+        foreach (var file in changedFiles)
+        {
+            Console.WriteLine(file);
```
premun: Is this useful?

eng/pipelines/pr.yml
```diff
@@ -46,3 +46,28 @@ stages:
       scope: lite
     ${{ else }}:
       scope: full
+
+- stage: ValidateUserChanges
+  displayName: Validate user changes in VMR
+  dependsOn: []
+  condition: and(eq(variables['isPRTrigger'], 'true'), ne(variables['Build.RequestedForEmail'], 'maestro-dotnet-bot@microsoft.com'))
```
premun: The original GitHub action from Ella also exempted `dotnet-sb-bot`

eng/tools/ChangeValidation/Validate.cs
```diff
@@ -0,0 +1,135 @@
...
+using Microsoft.DotNet.DarcLib.VirtualMonoRepo;
+using Microsoft.DotNet.DarcLib.Models.VirtualMonoRepo;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+
+namespace ValidateVmrChanges;
```
premun: The namespace differs from project name - maybe everything should be `ChangeValidation`? Even the class below? Though usually the entrypoint classes in C# are named `Program`

eng/tools/ChangeValidation/Validate.cs
adamzip: No - it's my debugging

eng/tools/ChangeValidation/ExclusionFileValidation.cs
adamzip: BTW, just a note that checking out between the PR branch and main can take quite a while, since it needs to updates tens or hundreds of thousands of files. I've experienced it when running locally too. When creating helpers in darclib we should keep this in mind and maybe use "git show <...>" as much as possible (which ties in to a point we talked about this week - that whenever we know specific commit SHAs from where we want to fetch metadata, we should have efficient methods for getting the metadata files based on those SHAs)

eng/Versions.props
```diff
@@ -27,6 +27,10 @@
     -->
     <PrivateSourceBuiltSdkVersion>10.0.100-preview.7.25372.107</PrivateSourceBuiltSdkVersion>
     <PrivateSourceBuiltArtifactsVersion>10.0.100-preview.7.25372.107</PrivateSourceBuiltArtifactsVersion>
+    <!-- arcade-services dependencies -->
+    <MicrosoftDotNetDarcLibVersion>1.1.0-beta.25404.1</MicrosoftDotNetDarcLibVersion>
```
premun, line 31: I think you need to merge main because this will now have t be added in `Version.Details.props`

eng/pipelines/pr.yml
```diff
@@ -46,3 +46,37 @@ stages:
...
+                    ) 
+                )
+  jobs:
+  - job: Validate
+    displayName: Run Validation Script
+    pool:
+      vmImage: windows-latest
+    steps:
+    - checkout: self
+      fetchDepth: 0
```
premun: Did you mention we're supposed to be doing the full clone?

premun (approved):

eng/pipelines/pr.yml
adamzip: if I'm not mistaken fetchDepth 0 is a full clone and fetchDepth N is a shallow copy that includes the last N commits

eng/Versions.props
adamzip: I've rebased on main quite recently, I have `Version.Details.props` on the PR branch. Should I remove DarcLibVersion here and add it under `<!-- arcade-services dependencies -->` in `Version.Details.props` instead?

premun: Yes, it needs to be in the other file

eng/pipelines/pr.yml
```diff
@@ -46,3 +46,40 @@ stages:
...
+      pool:
+        vmImage: windows-latest
+      steps:
+        - checkout: self
+          fetchDepth: 0
+
+        - powershell: |
+            Write-Host "Fetching target branch: $(System.PullRequest.TargetBranch)"
+            git fetch origin $(System.PullRequest.TargetBranch)
+          displayName: Fetch target branch
```
premun, line 72: Do we need this with the full clone?

Directory.Packages.props
premun: This one is still out of place

eng/pipelines/pr.yml
adamzip: Yes it's needed. My understanding is that `checkout: self` with `fetchDepth: 0` gives you the full history of the fetched branch (the PR branch), but not the history of other branches

premun:
@adamzip we will have to force-merge tomorrow

PR #1293, opened 2025-06-24
Fix VMR signing validation

mmitche:
A few fixes for VMR signing validation:
- Move files rather than copying to save space
- Only check externally visible, release shipping files
- Fix a couple failures where Dir.EnumerateFiles could not find the input directory because it did not exist.
- Because we pre-delete files on Mac, do not log warnings when a file could not be found. This causes an AzDO failure.
- Fix the deletion logic so that it correctly evaluates the glob paths on non-Windows.

### DIFF
--- a/eng/pipelines/templates/steps/vmr-validate-signing.yml
+++ b/eng/pipelines/templates/steps/vmr-validate-signing.yml
@@ -42,7 +42,7 @@ steps:
     displayName: Remove Source-Build Artifacts
     continueOnError: ${{ parameters.continueOnError }}
 - ${{ else }}:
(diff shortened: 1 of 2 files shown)


mmitche:
https://github.com/dotnet/arcade/pull/15928 is another fix required for a green sign check. The dotnet-hosting-options msi does not have a Binary table.

mmitche:
Latest test (without arcade fix): https://dev.azure.com/dnceng/internal/_build/results?buildId=2736209&view=results

mmitche:
Based on the results of the latest test, p6 is green.

PR #1304, opened 2025-06-24
Shallow clone in vmr-build.yml

akoeplinger:


jkoritzinsky:
In dotnet/runtime we do a checkout fetch depth of 20. Should we match that here or do something in the middle?

akoeplinger:
I think that is some leftover from when AzDO was not reliably checking out the same commit for all jobs (depending on when the job started) and we manually reset the commit. I don't think we need this anymore.

ViktorHofer (approved):

PR #1317, opened 2025-06-25
Do not clear the pre-release label in VMR builds

mmitche:
Clearing the label is unnecessary; this package is only for transport in VMR builds. This avoids both the VMR and repo producing the same version number on successive days.


ViktorHofer (approved):

hoyosjs (approved):
Is this just a temporary work around?

mmitche:
> Is this just a temporary work around?

Permanant. We don't have a great solution otherwise and there is pushback on changing the revision numbers

PR #1325, opened 2025-06-26
Use secret arg to determine if reverse pinvoke is ILStub or not

hoyosjs:
Modified JIT_ReversePInvokeEnterTrackTransitions to use the given secret arg to identify if a method desc needs unwrapping and is a stub.


jkotas (approved):
Thank you!

Issue #1110, opened 2025-06-11
[feature bands] Publish source-built packages


mthalman:
Official builds of the VMR's 1xx branch need to publish the source-built packages that get produced from it so that those packages can be provided as input to the source build of non-1xx branches (2xx, 3xx, etc). The exact method of publishing still needs to be determined and is discussed in https://github.com/dotnet/dotnet/pull/934. But publishing should be limited to the 1xx branch only.

PR #1355, opened 2025-06-30
Publish source build artifacts

mthalman:
Fixes https://github.com/dotnet/dotnet/issues/1110

This updates the UB pipeline to publish source built artifacts. This is necessary in order for source-only builds of the 2xx branch to be able to consume the packages that were produced by the 1xx branch. For example, the 1xx branch would produce the System.Text.Json package from the runtime repo which can then be pulled in as a dependency for the 2xx branch's source build, since runtime source isn't included in that branch.

Originally I tried to include the artifacts tarball in the existing VerticalManifest.xml file that gets produced by a source-only job. That's problematic because of a chicken-and-egg problem. The artifacts tarball includes the VerticalManifest.xml file in it. But since we want the VerticalManifest.xml file to include a reference to this tarball, we can't produce the VerticalManifest.xml file without the artifacts tarball already existing.

Instead a different approach is taken where a separate manifest file is produced. This allows the artifacts tarball to be produced after the VerticalManifest.xml file just as before. And then a new manifest file gets generated that only references the artifacts tarball. This is done by making two passes of the `Publish` target, referencing different manifests and configuration to include the appropriate files.

Prior to these changes all manifests from source-only jobs were excluded from the processing in the `Publish` stage. These changes update that to specifically exclude the original VerticalManifest.xml file and only include the manifest file containing the artifacts tarball, whose name is based on the job name for uniqueness purposes.

Some other things that needed to be fixed with these changes:
* Offline builds need to pass the `-E` option to `sudo` in order to flow the system environment variables to the build. Without this option set, AzDO variables like `BUILD_BUILDID` were not set which are necessary in order to produce a correct manifest file.
* Roslyn publishing needs an update to skip explicitly adding `FilesToPublishToSymbolServer` items for source-only builds. These symbols shouldn't be published for source-only and were incorrectly showing up in the manifests for the artifacts tarball without these changes.


src/roslyn/eng/Publishing.props
```diff
@@ -6,7 +6,7 @@
 
   <!-- Update Artifacts with Kind=Package to have additional metadata item Category="ToolingPackage".
        Depending on channel configuration, this means that these assets could be pushed to a different feed. -->
-  <ItemGroup>
+  <ItemGroup Condition="'$(DotNetBuildSourceOnly)' != 'true'">
```
ViktorHofer, line 10: Consider adding a comment for this condition's purpose?

eng/Publishing.props
```diff
@@ -70,7 +71,7 @@
     </ItemGroup>
 
     <!-- Create Artifact items for produced packages and assets -->
-    <ItemGroup>
+    <ItemGroup Condition="'$(IsPublishPass2)' == 'false'">
```
ViktorHofer: Check on `!= true` as in eng/PublishSourceBuild.props? That makes the property a true boolean instead of checking on true, false and other values.

eng/Publishing.props
```diff
@@ -81,7 +82,7 @@
       </Artifact>
     </ItemGroup>
 
-    <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
+    <ItemGroup Condition="'$(IsPublishPass2)' == 'false' and '$(DotNetBuildSourceOnly)' == 'true'">
```
ViktorHofer:
```suggestion
    <ItemGroup Condition="'$(IsPublishPass2)' != 'true' and '$(DotNetBuildSourceOnly)' == 'true'">
```

src/roslyn/eng/Publishing.props
```diff
@@ -5,8 +5,9 @@
   </PropertyGroup>
 
   <!-- Update Artifacts with Kind=Package to have additional metadata item Category="ToolingPackage".
-       Depending on channel configuration, this means that these assets could be pushed to a different feed. -->
-  <ItemGroup>
+       Depending on channel configuration, this means that these assets could be pushed to a different feed.
+       Do not include these for source-only builds since these files aren't intended to be published in that mode. -->
```
ViktorHofer, line 9: Just curious, why are those not intended to get published? Also this item group here only does an Update not an Include. Why is that a problem?

mthalman: If you look at the pipelines artifacts of any source build job (today, without these changes), you'll see that it contains a PDB directory that contains this PDB. This is the only PDB of all the repos that get built in the VMR that shows up there. Similarly, this is the only PDB that exist in the VerticalManifest.xml file that gets output as well. So clearly that's not right. Source-only builds have their own symbols tarball output in which this PDB does exist.

The specific issue is the `FilesToPublishToSymbolServer` definitions here which is an include, not update.

PR #1370, opened 2025-07-01
Change Roslyn from main-vs-deps to main

dkurepa:


akoeplinger:
I noticed we were missing a couple changes done directly in the VMR that weren't yet flown back to roslyn and would've been overwritten, I cherry-picked those into the PR.

The only exception are changes from https://github.com/dotnet/dotnet/pull/1240 done by @JoeRobich since it looked like they weren't needed.

akoeplinger:
@dotnet/source-build looks like we need to add Microsoft.CodeAnalysis.Common 4.12.0 to SBRP

MichaelSimons:
> @dotnet/source-build looks like we need to add Microsoft.CodeAnalysis.Common 4.12.0 to SBRP

SBRP is intended to be self-servicing.  [Here](https://github.com/dotnet/source-build-reference-packages?tab=readme-ov-file#reference) are the instructions for adding new reference packages.

akoeplinger:
@MichaelSimons thanks, I opened https://github.com/dotnet/source-build-reference-packages/pull/1290

PR #1372, opened 2025-07-01
Require user interaction when merge conflicts are encountered

ellahathaway:
This PR includes some updates to the bash and powershell scripts to handle merge conflicts in repos that _are_ included in the non-1xx branches.

### DIFF
--- a/eng/update-band.ps1
+++ b/eng/update-band.ps1
@@ -1,25 +1,41 @@
 [CmdletBinding(PositionalBinding=$false)]
 Param(
-    [Parameter(Mandatory = $true)]
+    [Parameter(Mandatory = $false)]
     [string]$Remote,
 
-    [Parameter(Mandatory = $true)]
-    [Alias("branch-1xx")]
+    [Parameter(Mandatory = $false)]
     [string]$Branch1xx,
+
+    [switch]$Continue = $false,
 
     [switch][Alias('h')]$help
 )
 
 function Get-Usage {
     Write-Host "  -Remote <name>           Git remote to pull from (e.g., upstream)"
-    Write-Host "  -1xx-branch <branch>     1xx branch name to merge in (e.g., main)"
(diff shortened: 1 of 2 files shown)


eng/update-band.ps1
```diff
@@ -1,25 +1,41 @@
...
     [string]$Remote,
 
-    [Parameter(Mandatory = $true)]
-    [Alias("branch-1xx")]
+    [Parameter(Mandatory = $false)]
     [string]$Branch1xx,
 
+    [switch]$Continue = $false,
+
     [switch][Alias('h')]$help
```
mthalman: Sorry I didn't catch this on the first review but there's no need to have a help option for PowerShell. It's integrated into PowerShell itself: `Get-Help update-band.ps1`. To provide descriptions for the parameters, you can use comment-based help: https://learn.microsoft.com/powershell/scripting/developer/help/examples-of-comment-based-help

eng/update-band.ps1
```diff
@@ -1,25 +1,41 @@
...
 )
 
 function Get-Usage {
     Write-Host "  -Remote <name>           Git remote to pull from (e.g., upstream)"
-    Write-Host "  -1xx-branch <branch>     1xx branch name to merge in (e.g., main)"
+    Write-Host "  -Branch1xx <branch>     1xx branch name to merge in (e.g., main)"
+    Write-Host "  -Continue                Continue after manually fixing merge conflicts"
```
mthalman: Consider a different name. `Continue` is a reserved word so this may cause confusion: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_continue

ellahathaway: Changed to `ContinueMerge` (pwsh) and `--continue-merge` (bash)

mthalman (approved):

Issue #1106, opened 2025-06-11
[feature bands] Setup mock 10.0.2xx branch for development testing


mthalman:
In order to test multiple feature bands, we're going to need a branch that simulates what a 10.0.2xx branch would look like. We can't wait for the actual branch before implementing the work to support multiple feature bands, so we'll need to create a mock branch.

This branch would also need to be manually configured to mimic what a dependency flow from a 1xx VMR build would look like since there wouldn't be actual dependency flow to this branch.

mthalman:
Branches that will be preserved in the 2xx branch:
* `arcade`
* `source-build-reference-packages`
* `scenario-tests`
* tooling repos (@mmitche - do you have a definitive list of these?)

mthalman:
> * tooling repos ([@mmitche](https://github.com/mmitche) - do you have a definitive list of these?)

Here is my best guess:

* ~~`diagnostics`~~
* `fsharp`
* `msbuild`
* `nuget-client`
* `razor`
* `roslyn`
* `roslyn-analyzers`
* `sdk`
* `templating`
* `vstest`

mthalman:
Part of this work should include a script for keeping this branch updated with new changes from main branch (obviously excluding any changes that were made in deleted repos).

mmitche:
I think:
- No diagnostics
- We had discussed that always putting arcade in made some sense.

Otherwise looks good.

ellahathaway:
Here's the dev branch: https://github.com/dotnet/dotnet/tree/10.0.2xx-dev

ellahathaway:
FYI - there's currently no protections on that branch. Do we want to set some up, and if so, what should they be?

mthalman:
> FYI - there's currently no protections on that branch. Do we want to set some up, and if so, what should they be?

Shouldn't be necessary. It's not a production branch.

ellahathaway:
> Part of this work should include a script for keeping this branch updated with new changes from main branch (obviously excluding any changes that were made in deleted repos).

Scripts have been merged. I'm not sure how and when these scripts will be triggered though. Do we want them to be triggered manually? If not, we'll need to add some automation. @mthalman, what are your thoughts?

mthalman:
Manual is fine. We can merge changes in as necessary.

mthalman:
I think this can be closed then?

ellahathaway:
> Manual is fine. We can merge changes in as necessary.
> I think this can be closed then?

[...]

Issue #1107, opened 2025-06-11
[feature bands] Create tool to assist in creating feature band branches


mthalman:
This tool would automate all the necessary steps needed to configure a feature band branch, such as removing shared component repos. It could potentially be expanded to include other common tasks such as branding as well.

mthalman:
@mmitche - It sounds like you're already thinking about branch/branding in the context of previews. Would your work incorporate feature bands as well?

mmitche:
Yeah will do

mmitche:
Finished in current local WIP tooling. Looking good there.

mmitche:
Completed

PR #1373, opened 2025-07-01
Condition repository refs for repos not built in non-1xx branches

ellahathaway:
Related to https://github.com/dotnet/dotnet/issues/1106

This avoids building repos that we don't include in non-1xx branches.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,7 +25,6 @@
     <ArtifactsObjDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'obj'))</ArtifactsObjDir>
     <ArtifactsBinDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'bin'))</ArtifactsBinDir>
     <RepositoryEngineeringDir>$([MSBuild]::NormalizeDirectory('$(RepoRoot)', 'eng'))</RepositoryEngineeringDir>
-    <SourcesDir>$([MSBuild]::NormalizeDirectory('$(RepoRoot)', 'src'))</SourcesDir>
 
     <!-- ProjectLayout.props -->
     <PlatformName Condition="'$(PlatformName)' == ''">$(Platform)</PlatformName>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,6 +25,7 @@
     <ArtifactsObjDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'obj'))</ArtifactsObjDir>
     <ArtifactsBinDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'bin'))</ArtifactsBinDir>
     <RepositoryEngineeringDir>$([MSBuild]::NormalizeDirectory('$(RepoRoot)', 'eng'))</RepositoryEngineeringDir>
+    <SourcesDir>$([MSBuild]::NormalizeDirectory('$(RepoRoot)', 'src'))</SourcesDir>
 
     <!-- ProjectLayout.props -->
     <PlatformName Condition="'$(PlatformName)' == ''">$(Platform)</PlatformName>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -59,6 +59,8 @@
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
     <!-- By default, the VMR builds with online sources when not building source-only. -->
     <DotNetBuildWithOnlineFeeds Condition="'$(DotNetBuildWithOnlineFeeds)' == '' and '$(DotNetBuildSourceOnly)' != 'true'">true</DotNetBuildWithOnlineFeeds>
+    <!-- Only build runtime components in 1xx feature band. -->
+    <DotNetBuildRuntimeComponents Condition="'$(DotNetBuildRuntimeComponents)' == '' and '$(VersionSDKMinor)' == '1'">true</DotNetBuildRuntimeComponents>
     <!-- Don't use Arcade's ExcludeFrom* build infra in the VMR orchestrator. -->
     <DisableArcadeExcludeFromBuildSupport>true</DisableArcadeExcludeFromBuildSupport>
     <LangVersion>latest</LangVersion>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -59,8 +59,8 @@
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
     <!-- By default, the VMR builds with online sources when not building source-only. -->
     <DotNetBuildWithOnlineFeeds Condition="'$(DotNetBuildWithOnlineFeeds)' == '' and '$(DotNetBuildSourceOnly)' != 'true'">true</DotNetBuildWithOnlineFeeds>
-    <!-- Only build runtime components in 1xx feature band. -->
-    <DotNetBuildRuntimeComponents Condition="'$(DotNetBuildRuntimeComponents)' == '' and '$(VersionSDKMinor)' == '1'">true</DotNetBuildRuntimeComponents>
+    <!-- Only build shared components in 1xx feature band. -->
+    <DotNetBuildSharedComponents Condition="'$(DotNetBuildSharedComponents)' == '' and '$(VersionSDKMinor)' == '1'">true</DotNetBuildSharedComponents>
     <!-- Don't use Arcade's ExcludeFrom* build infra in the VMR orchestrator. -->
     <DisableArcadeExcludeFromBuildSupport>true</DisableArcadeExcludeFromBuildSupport>
     <LangVersion>latest</LangVersion>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,4 +1,4 @@
-<Project InitialTargets="ValidateShortStackBuild">
+<Project>
 
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" Condition="'$(SkipArcadeSdkImport)' != 'true'" />
 
@@ -60,13 +60,5 @@
       <RuntimeArchivePath>%(RuntimeArchiveItem.Identity)</RuntimeArchivePath>
       <SourceBuiltRuntimeVersion>$(RuntimeFilename.Replace('$(RuntimeFilenamePrefix)','').Replace('-$(TargetRid)$(ArchiveExtension)',''))</SourceBuiltRuntimeVersion>
     </PropertyGroup>
-  </Target>
-
-  <Target Name="ValidateShortStackBuild" Condition="'$(ShortStack)' == 'true'" >
-    <!--
-      A short-stack build will only build runtime and its dependencies.
-      We should not and cannot have a short-stack build if we're not building the runtime components.
-    -->
-    <Error Text="Cannot have a short-stack build if we're not building the runtime components." Condition="'$(ShortStack)' == 'true' and '$(DotNetBuildRuntimeComponents)' != 'true'" />
   </Target>
 </Project>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,4 +1,4 @@
-<Project>
+<Project InitialTargets="ValidateFeatureBand">
 
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" Condition="'$(SkipArcadeSdkImport)' != 'true'" />
 
@@ -60,5 +60,13 @@
       <RuntimeArchivePath>%(RuntimeArchiveItem.Identity)</RuntimeArchivePath>
       <SourceBuiltRuntimeVersion>$(RuntimeFilename.Replace('$(RuntimeFilenamePrefix)','').Replace('-$(TargetRid)$(ArchiveExtension)',''))</SourceBuiltRuntimeVersion>
     </PropertyGroup>
+  </Target>
+
+  <Target Name="ValidateFeatureBand" Condition="'$(VersionSDKMinor)' != '1'" >
+    <!--
+      Non-1xx feature bands do not include runtime, and a short-stack build will build runtime and its dependencies only.
+      Therefore, we should not and cannot have a short-stack build on a non-1xx feature band.
+    -->
+    <Error Text="Cannot have a short-stack build on a non-1xx feature band." Condition="'$(ShortStack)' == 'true'" />
   </Target>
 </Project>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,4 +1,4 @@
-<Project InitialTargets="ValidateFeatureBand">
+<Project InitialTargets="ValidateShortStackBuild">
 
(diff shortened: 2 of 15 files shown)


ellahathaway:
This includes changes to files shared by the 1xx branch. Would it be better if this PR targeted the main branch?

repo-projects/dotnet.proj
```diff
@@ -12,17 +12,17 @@
   <ItemGroup Condition="'$(DotNetBuildPass)' == '' or '$(DotNetBuildPass)' == '1'">
     <!-- Short stack builds stop at runtime, not the whole SDK -->
     <RepositoryReference Include="sdk" Condition="'$(ShortStack)' != 'true'" />
-    <RepositoryReference Include="runtime" Condition="'$(ShortStack)' == 'true'" />
+    <RepositoryReference Include="runtime" Condition="'$(ShortStack)' == 'true' and Exists($([MSBuild]::NormalizeDirectory('$(SourcesDir)', 'runtime')))" />
```
jkoritzinsky: Instead of this condition, could we instead have a short-stack build on a .2xx branch result in an error?

jkoritzinsky (commented):
I think I'd rather this sort of condition be automatic on the `RepositoryReference` item or applied based on some metadata (like a "FeatureBand100Only" metadata key).

Is there a way we could have this only trigger in .2xx builds (ie do we have any versioning logic in the orchestrator we can use)?

mthalman:
> This includes changes to files shared by the 1xx branch. Would it be better if this PR targeted the main branch?

My thinking was that all the development should be done in main. The `10.0.2xx-dev` branch should be used only to verify the changes in a 2xx context.

mthalman:
> I agree with @jkoritzinsky that we should avoid these conditional Exists checks and consider adding metadata that would allow filtering out later. I.e. if you add a new metadata to those items, we could remove them then later in Directory.Build.targets at evaluation time. I.e.
> 
> ```
> <ItemGroup Condition="ISNOT1XXFEATUREBAND">
>   <RepositoryReference Remove="@(RepositoryReference->WithMetdataValue('BuildOnlyFor1XX', 'true'))" />
> </ItemGroup>
> ```

Agreed. The determination of whether it's a non-1xx feature band can be done by getting the `VersionSDKMinor` property from the sdk repo.

ViktorHofer:
> Agreed. The determination of whether it's a non-1xx feature band can be done by getting the VersionSDKMinor property from the sdk repo.

I'm not sure we can (and we probably also shouldn't) import src\sdk\eng\Versions.props in the VMR orchestrator. That would mutate a ton of state that might cause issues. Should we define that property in VMR's eng\Versions.props file as well?

mthalman:
> I'm not sure we can (and we probably also shouldn't) import src\sdk\eng\Versions.props in the VMR orchestrator. That would mutate a ton of state that might cause issues. Should we define that property in VMR's eng\Versions.props file as well?

I was thinking we just read it as XML on the file rather than through MSBuild. I'm leery of having to maintain duplicates. However, with [tooling](https://github.com/dotnet/dotnet/issues/1107) that would manage things, that may be more appealing.

ellahathaway:
> > I'm not sure we can (and we probably also shouldn't) import src\sdk\eng\Versions.props in the VMR orchestrator. That would mutate a ton of state that might cause issues. Should we define that property in VMR's eng\Versions.props file as well?
> 
> I was thinking we just read it as XML on the file rather than through MSBuild. I'm leery of having to maintain duplicates. However, with [tooling](https://github.com/dotnet/dotnet/issues/1107) that would manage things, that may be more appealing.

I think with tooling, the approach I prefer is to duplicate this property. That said, I'm open to reading `src\sdk\eng\Versions.props` as an XML as well.

MichaelSimons:
It seems like everyone is gravitating towards metadata indicating the feature band - eg `BuildOnlyFor1XX`.  I was thinking of a more higher level abstraction - e.g `BuildRuntimeComponents`.  I don't like coupling the source to branch/branding.  Additionally a higher level abstraction is more understandable to me as to why it may be excluded/included.

ellahathaway:
> I don't like coupling the source to branch/branding. Additionally a higher level abstraction is more understandable to me as to why it may be excluded/included.

The struggle that I have with this is that we'd still have to condition the references on the `VersionSDKMinor` prop, which is directly coupling the source to branch/branding. Because of this, I inherently learn towards using `BuildOnlyFor1XX`.

Is there an instance when we wouldn't want to build the runtime components outside of the non-1xx feature bands? I can see using a higher level of abstraction, like `BuildRuntimeComponents`, in that case.

MichaelSimons:
>Is there an instance when we wouldn't want to build the runtime components outside of the non-1xx feature bands? I can see using a higher level of abstraction, like BuildRuntimeComponents, in that case.

Perhaps a distro maintainer wants to only build the latest feature band and is willing to do the work to fix the shared components to build with the latest toolset.  In this case they would manually override `BuildRuntimeComponents`. Having this named `BuildOnlyFor1XX` is less than ideal.  In this case it feels appropriate that we would have defaulting logic based on the version/branching but decouple the property name itself.

eng/Versions.props
```diff
@@ -4,6 +4,7 @@
     <VersionPrefix>0.1.0</VersionPrefix>
     <PreReleaseVersionLabel>alpha.1</PreReleaseVersionLabel>
     <UseVSTestRunner>true</UseVSTestRunner>
+    <VersionSDKMinor>2</VersionSDKMinor>
```
ellahathaway: Added this per https://github.com/dotnet/dotnet/pull/1373#issuecomment-3025356926.

@mthalman - let me know if you think I should keep this or if I should read it as XML on the file instead.

Unrelated: If we keep this prop, I'll update the value to be `1` when I switch the target branch to main. I'll also submit a PR against the 10.0.2xx-dev branch to update the value to `2`.

repo-projects/dotnet.proj
ellahathaway: I added the error [here](https://github.com/dotnet/dotnet/pull/1373/files#diff-50e91c82311ea26f2a73202525dfdf2b0a89c178ee666b2bf3ad4c84ac4c06dcR65-R71).

I did keep [the metadata on the reference](https://github.com/dotnet/dotnet/pull/1373/files#diff-8fce0228eadb46e761b5e34ae2bf266a82abce96241d4570f25ae3d2dc4e8068R15) for consistency, but I can remove it if you think it's unnecessary.

Directory.Build.props
```diff
@@ -59,6 +59,8 @@
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
     <!-- By default, the VMR builds with online sources when not building source-only. -->
     <DotNetBuildWithOnlineFeeds Condition="'$(DotNetBuildWithOnlineFeeds)' == '' and '$(DotNetBuildSourceOnly)' != 'true'">true</DotNetBuildWithOnlineFeeds>
+    <!-- Only build runtime components in 1xx feature band. -->
+    <DotNetBuildRuntimeComponents Condition="'$(DotNetBuildRuntimeComponents)' == '' and '$(VersionSDKMinor)' == '1'">true</DotNetBuildRuntimeComponents>
```
ellahathaway: Related to https://github.com/dotnet/dotnet/pull/1373#issuecomment-3025576569

@MichaelSimons - let me know what you think about this. I think it provides the overridability that you were looking for while still disabling building the runtime components on non-1xx branches by default.

MichaelSimons: Yes, I like this direction.

Directory.Build.targets
```diff
@@ -1,4 +1,4 @@
-<Project>
+<Project InitialTargets="ValidateShortStackBuild">
```
ViktorHofer: This initial target would run for every project, unnecessarily. We also noticed msbuild scheduling issues when using InitialTargets. Consider moving this target directly into dotnet.proj and use a different hook point, i.e. `BeforeTargets="PrepareProjectReferences"`. You could condition on `RepositoryReference` being empty or on ShortStack being true. Both would work.

Directory.Build.props
mthalman: My preference for the name would be `DotNetBuildSharedComponents`.

eng/Versions.props
mthalman: I think this is fine. I'd like feedback from @mmitche too. @mmitche - the idea here was that duplicating this value between here and the sdk would be acceptable since we'll be having tooling for branding.

repo-projects/msbuild.proj
```diff
@@ -10,10 +10,10 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
-    <RepositoryReference Include="roslyn" />
-    <RepositoryReference Include="runtime" />
+    <RepositoryReference Include="roslyn" BuildRuntimeComponents="true" />
```
mthalman: I'd prefer to not have to set `BuildRuntimeComponents` for all these references across all the proj files. I'd like one central location where all the shared repos are defined. And then the dynamic behavior of removing references to those repos would be based on that.

Directory.Build.targets
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1373/commits/87e6a9b5e2df6967f58754338015d70106482a3e

repo-projects/msbuild.proj
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1373/commits/55c432b16c7d343bd1978518f82b2dd78caf321d

repo-projects/Directory.Build.props
```diff
@@ -227,4 +227,24 @@
...
+    <RuntimeRepositoryReference Include="aspire" />
+    <RuntimeRepositoryReference Include="aspnetcore" />
+    <RuntimeRepositoryReference Include="cecil" />
+    <RuntimeRepositoryReference Include="command-line-api" />
+    <RuntimeRepositoryReference Include="deployment-tools" />
+    <RuntimeRepositoryReference Include="diagnostics" />
+    <RuntimeRepositoryReference Include="efcore" />
+    <RuntimeRepositoryReference Include="emsdk" />
+    <RuntimeRepositoryReference Include="runtime" />
+    <RuntimeRepositoryReference Include="source-build-externals" />
```
ViktorHofer: That repo doesn't exist anymore. Are you maybe on an old HEAD?

repo-projects/dotnet.proj
```diff
@@ -9,20 +9,30 @@
...
 
+  <Target Name="ValidateShortStackBuild"
+          Condition="'$(ShortStack)' == 'true'"
+          BeforeTargets="PrepareProjectReferences" >
+    <!--
+      A short-stack build will only build runtime and its dependencies.
+      We should not and cannot have a short-stack build if we're not building the runtime components.
+    -->
+    <Error Text="Cannot have a short-stack build if we're not building the runtime components." Condition="'$(DotNetBuildRuntimeComponents)' != 'true'" />
+  </Target>
```
ViktorHofer: tiny readability nit: we usually put targets after properties and items

repo-projects/dotnet.proj
```diff
@@ -9,20 +9,30 @@
...
   </PropertyGroup>
 
+  <Target Name="ValidateShortStackBuild"
+          Condition="'$(ShortStack)' == 'true'"
+          BeforeTargets="PrepareProjectReferences" >
+    <!--
+      A short-stack build will only build runtime and its dependencies.
+      We should not and cannot have a short-stack build if we're not building the runtime components.
+    -->
+    <Error Text="Cannot have a short-stack build if we're not building the runtime components." Condition="'$(DotNetBuildRuntimeComponents)' != 'true'" />
```
ViktorHofer: Any reason why this condition isn't part of the target condition?

repo-projects/Directory.Build.props
ellahathaway: Yeah I'm targeting the 10.0.2xx dev branch right now, which hasn't been updated with that change yet. Will update this section when I update the PR to target main.

eng/Versions.props
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1373/commits/bf36a5ee0b3fc1b9d85df7b583c9e551a04c5e23

Directory.Build.props
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1373/commits/bf36a5ee0b3fc1b9d85df7b583c9e551a04c5e23

repo-projects/dotnet.proj
ellahathaway: The name of the target felt too generic to condition on the `DotnetBuildRuntimeComponents` prop as well. I'm open to moving the condition and/or updating the target name. Let me know what you think.

repo-projects/dotnet.proj
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1373/commits/ea939aa98739589ef73d46a17d0427b6cf8ad53f

ViktorHofer (approved):

repo-projects/Directory.Build.props
ViktorHofer: Now that the repos are merged how will this work? I assume both the SBRP and the SBE content will now always get built in 2xx, 3xx, 4xx? cc @MichaelSimons

MichaelSimons: > I assume both the SBRP and the SBE content will now always get built in 2xx, 3xx, 4xx?

That's correct.

mthalman (approved):

repo-projects/Directory.Build.props
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1373/commits/11338541af2b7f0b2042ecb5f8609cf511327be4

eng/Versions.props
ellahathaway: > Unrelated: If we keep this prop, I'll update the value to be 1 when I switch the target branch to main. I'll also submit a PR against the 10.0.2xx-dev branch to update the value to 2.

Addressed in https://github.com/dotnet/dotnet/pull/1373/commits/11338541af2b7f0b2042ecb5f8609cf511327be4

ellahathaway:
@dotnet/source-build - can I get help merging this?
- The protected files validation failure is expected because I touched the `eng/Versions.props` file.
- The SB stage 2 failure is a scenario test failure that appears unrelated to my changes:
```
Microsoft.DotNet.SourceBuild.Tests.WebScenarioTests.VerifyScenario(scenario: TestScenario { Commands = Build | Run, Language = CSharp, NoHttps = False, ScenarioName = "WebScenarioTests", Template = Web, ··· }) [FAIL]
  System.InvalidOperationException : Failed to execute /__w/1/s/artifacts/obj/extracted-dotnet-sdk/dotnet run --no-launch-profile /bl:/__w/1/s/artifacts/TestResults/Release/WebScenarioTests_Web_CSharp-run.binlog
  Exit code: 134
  fail: Microsoft.Extensions.Hosting.Internal.Host[11]
  Hosting failed to start
  System.IO.IOException: Failed to bind to address http://[::]:33431/: address already in use.
  ---> Microsoft.AspNetCore.Connections.AddressInUseException: Address already in use
```

ViktorHofer (approved):

Issue #1354, opened 2025-06-30
Signing comparison is failing to remove SB artifacts


ellahathaway:
[Pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2736934&view=logs&j=8634bcad-249c-59e6-8fe3-cd6482720080&t=7f1417e2-f9db-5e8b-866a-232b14940af1)

On mac:
```
/usr/local/bin/pwsh -NoLogo -NoProfile -NonInteractive -Command . '/Users/runner/work/_temp/48b83c65-c90b-4cce-bba3-56378745b843.ps1'
rm: /Users/runner/work/1/a/vmr-assets/*Sdk_*_Artifacts: No such file or directory

##[error]PowerShell exited with code '1'.
```

On linux:
```
/usr/bin/pwsh -NoLogo -NoProfile -NonInteractive -Command . '/mnt/vss/_work/_temp/e36350cb-c906-414e-9fd8-e8937b6432b7.ps1'
/usr/bin/rm: cannot remove '/mnt/vss/_work/1/a/vmr-assets/*Sdk_*_Artifacts': No such file or directory

##[error]PowerShell exited with code '1'.
```

mmitche:
Due to the quoting of the path.

mmitche:
Resolved with https://github.com/dotnet/dotnet/pull/1406

PR #1406, opened 2025-07-07
Fixup deletion command to pick up wildcarded directories

mmitche:
The `rm` form launches the executable and interprets the quoted version as a literal path. Avoid this by using Remove-Item


mmitche:
Resolves https://github.com/dotnet/dotnet/issues/1354

jkoritzinsky (approved):

PR #1421, opened 2025-07-08
Re-enable Spectre.Console

jjonescz:
Follow up on https://github.com/dotnet/dotnet/pull/1402#issuecomment-3046613402.


src/source-build-reference-packages/src/externalPackages/projects/spectre-console.proj
```diff
@@ -16,6 +16,7 @@
       <BuildCommandArgs>$(BuildCommandArgs) /p:TargetFrameworks=$(NetCurrent)</BuildCommandArgs>
       <BuildCommandArgs>$(BuildCommandArgs) /p:UseBuildTimeTools=false</BuildCommandArgs>
       <BuildCommandArgs>$(BuildCommandArgs) /p:IsAotCompatible=false</BuildCommandArgs>
+      <BuildCommandArgs>$(BuildCommandArgs) /p:PublicSign=true</BuildCommandArgs>
```
jjonescz: 📝 without PublicSign=true, errors like this appear: https://github.com/dotnet/arcade/issues/9633

MichaelSimons: Would it make sense to follow the pattern used by several other projects?  [example](https://github.com/dotnet/source-build-reference-packages/blob/main/src/externalPackages/projects/humanizer.proj#L14)

PR #1440, opened 2025-07-09
Add a dependency on the source-build build for publishing

mmitche:
Publishing now pushes out the source-built assets. Add the missing stage dependency

### DIFF
--- a/eng/pipelines/templates/stages/vmr-build.yml
+++ b/eng/pipelines/templates/stages/vmr-build.yml
@@ -97,7 +97,9 @@ stages:
 - ${{ if and(parameters.isBuiltFromVmr, eq(variables['System.TeamProject'], 'internal'), ne(variables['Build.Reason'], 'PullRequest')) }}:
   - stage: Publish_Build_Assets
     displayName: Publish Assets
-    dependsOn: VMR_Vertical_Build
+    dependsOn:
+    - VMR_Vertical_Build
+    - VMR_SourceOnly_Build
     jobs:
     - template: /eng/common/templates-official/job/publish-build-assets.yml
       parameters:


jkoritzinsky (approved):

ViktorHofer:
Is it possible to run the publishing stage even when the source-build stage fails? I wonder about cases when source-build is broken for a longer period of time which would halt nightly msft builds.

jkoritzinsky:
Maybe it would be worth it/possible to run the source build leg that gets published in the VMR Verticals stage so we don't depend on the rest of the SB jobs that don't get published?

mmitche:
> Maybe it would be worth it/possible to run the source build leg that gets published in the VMR Verticals stage so we don't depend on the rest of the SB jobs that don't get published?

@mthalman To comment, since this is about the multi-band SDK support.

mmitche:
> Is it possible to run the publishing stage even when the source-build stage fails? I wonder about cases when source-build is broken for a longer period of time which would halt nightly msft builds.

This doesn't gate publishing on the SB validation, but it does gate on SB build, which, when we have multi-band support for real, we cannot publish without. I do not think those should be separated. The MS produced artifacts for SB are artifacts we ship, just like System.Text.Json.

We might be able to, as @jkoritzinsky mentioned, pull out the SB job that produces those artifacts and avoid gating on stage 2, for instance.

ViktorHofer:
Yes, stage2 builds were my primary concern here

mthalman:
> Is it possible to run the publishing stage even when the source-build stage fails? I wonder about cases when source-build is broken for a longer period of time which would halt nightly msft builds.

I could say the same thing about the vertical legs blocking the SB legs from publishing. 😁

> Maybe it would be worth it/possible to run the source build leg that gets published in the VMR Verticals stage so we don't depend on the rest of the SB jobs that don't get published?

I'm not in favor of this. It would overly complicate things. There are a handful of SB jobs which produce assets to publish. Keep in mind that SB validation is dependent on the SB legs too. We should either keep things as they are or integrate all the verticals and SB legs into a single stage.

mmitche:
Verified with internal build

Issue #1111, opened 2025-06-11
[feature bands] Consume source-built packages from 1xx build


mthalman:
Builds in non-1xx branches need to consume the source-built packages produced from the 1xx branch as input (see https://github.com/dotnet/dotnet/issues/1110). In order to consume these packages, it's necessary to know the version to target which will be determined by the package version of any package of a shared component (e.g. runtime) that would be available from the Version.Details.xml of a repo dependent on such a component. Related to https://github.com/dotnet/dotnet/issues/1108.

This requires a way to feed these in as inputs to the build similar to what is done with N-1 artifacts. It's an open question whether to combine N-1 artifacts with these 1xx packages or to have them separated.

PR #1480, opened 2025-07-14
Include runtime tarballs in SB artifacts

mthalman:
Contributes to https://github.com/dotnet/dotnet/issues/1111

In order to consume 1xx artifacts in a 2xx source-only build, the build of sdk repo needs access to the dotnet-runtime and aspnetcore-runtime tarballs so that it can include them as part of its redist.

I've placed these runtime tarballs into the existing artifacts tarball. I feel this is preferable because it doesn't require a new file to be introduced that would otherwise require a way to feed that into build. Instead, it can piggyback on the existing artifacts tarball input method.

The directory structure within the tarball follows the same structure as the `assets` directory. The intention is for the 2xx to build to simply extract this `assets` directory onto the existing `assets` directory of the repo to effectively simulate the output of these runtime tarballs.

Related: https://github.com/dotnet/source-build/issues/5280

### DIFF
--- a/eng/PublishSourceBuild.props
+++ b/eng/PublishSourceBuild.props
@@ -102,6 +102,17 @@
           DestinationFolder="$(SourceBuiltLayoutDir)"
           UseSymbolicLinksIfPossible="true" />
 
+    <ItemGroup>
+      <_RuntimeTarballs Include="$(ArtifactsAssetsDir)Runtime/**/dotnet-runtime-*$(ArchiveExtension)" />
+      <_RuntimeTarballs Include="$(ArtifactsAssetsDir)aspnetcore/Runtime/**/aspnetcore-runtime-*$(ArchiveExtension)"
+                        Exclude="$(ArtifactsAssetsDir)aspnetcore/Runtime/**/aspnetcore-runtime-composite-*$(ArchiveExtension)" />
+    </ItemGroup>
+
+    <!-- Copy runtime tarballs into the tarball preserving directory structure -->
+    <Copy SourceFiles="@(_RuntimeTarballs)"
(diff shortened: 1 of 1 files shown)


jkoritzinsky (approved):

eng/PublishSourceBuild.props
```diff
@@ -102,6 +102,17 @@
...
 
+    <ItemGroup>
+      <_RuntimeTarballs Include="$(ArtifactsAssetsDir)Runtime/**/dotnet-runtime-*$(ArchiveExtension)" />
+      <_RuntimeTarballs Include="$(ArtifactsAssetsDir)aspnetcore/Runtime/**/aspnetcore-runtime-*$(ArchiveExtension)"
+                        Exclude="$(ArtifactsAssetsDir)aspnetcore/Runtime/**/aspnetcore-runtime-composite-*$(ArchiveExtension)" />
+    </ItemGroup>
+
+    <!-- Copy runtime tarballs into the tarball preserving directory structure -->
+    <Copy SourceFiles="@(_RuntimeTarballs)"
+      DestinationFiles="@(_RuntimeTarballs->'$(SourceBuiltLayoutDir)$([MSBuild]::MakeRelative('$(ArtifactsAssetsDir)', '%(FullPath)'))')"
```
MichaelSimons:
```suggestion
          DestinationFiles="@(_RuntimeTarballs->'$(SourceBuiltLayoutDir)$([MSBuild]::MakeRelative('$(ArtifactsAssetsDir)', '%(FullPath)'))')"
```

eng/PublishSourceBuild.props
```diff
@@ -102,6 +102,17 @@
...
+      UseSymbolicLinksIfPossible="true" />
```
MichaelSimons:
```suggestion
          UseSymbolicLinksIfPossible="true" />
```

PR #1497, opened 2025-07-15
officialBuildId argument for build script

mthalman:
Contributes to https://github.com/dotnet/source-build/issues/5109

Setting an official build ID will be a required part of the workflow for distro maintainers. This adds it as an argument to the build script to make it first-class and discoverable. Also updates the pipeline to make use of this argument.

### DIFF
--- a/eng/build.ps1
+++ b/eng/build.ps1
@@ -9,6 +9,7 @@ Param(
   [string][Alias('v')]$verbosity = "minimal",
   [Parameter()][ValidateSet("preview", "rtm", "default")]
   [string]$branding = "default",
+  [string][Alias('obid')]$officialBuildId,
 
   # Actions
   [switch]$clean,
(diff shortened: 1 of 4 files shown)


eng/build.ps1
```diff
@@ -95,6 +97,7 @@ if ($branding) {
         "default" { $arguments += "" }
     }
 }
+if ($officialBuildId) { $arguments += "/p:OfficialBuildId=$officialBuildId" }
```
mmitche, line 101: We should add some validation to this argument. We don't have this validation within the arcade SDK because the use of the property happens before targets execute.

mthalman: @mmitche - Does this look good? https://github.com/dotnet/dotnet/pull/1497/commits/0e22ff47d7639aa98c3990fbda8995bc8944c2c3

Issue #1112, opened 2025-06-11
[feature bands] Poison leak detection


mthalman:
Source build [poison leak detection](https://github.com/dotnet/source-build/blob/main/Documentation/leak-detection.md) will need to exclude all shared component files that originate from packages produced by the 1xx branch. For example, the System.Xml.dll file included in the 2xx SDK originates from the 1xx branch. The intention of poison leak detection is to enforce that all source-built files can be serviceable. In the case of shared components coming from the 1xx branch, it is inherently true that they are serviceable in that branch. And they can then be safely redistributed in the 2xx branch. This is only true for shared component files. Files from tools, such as Roslyn, that originate from the 1xx branch may not be safely redistributed in the 2xx branch because the tools are meant to be produced by the 2xx output; if that were to happen, this should be considered a poison leak.

Whether artifacts provided as input to a 2xx build are separated out between 2xx PSB artifacts and 1xx artifacts or whether they are combined (see https://github.com/dotnet/dotnet/issues/1111) impacts how poisoning will work. If the artifact sets are provided as separate inputs, you know which are from 1xx and can designate those appropriately according to the poisoning requirements described in the previous paragraph. If they are combined, you need a way to know which came from 1xx which may require some sort of manifest file.

When it comes to the packages that come from the 1xx build, we need to ensure that none of the packages from tooling repos get leaked. For example, it's ok for a 1xx runtime package to get redistributed in the 2xx build, but not a 1xx roslyn package. See https://github.com/dotnet/dotnet/pull/934/files#r2140757079

PR #1502, opened 2025-07-15
Consume shared component artifacts for source-only build

### LINKED ISSUE #1111: [feature bands] Consume source-built packages from 1xx build

mthalman:
Fixes #1111

These changes enable support for a source-only build of a 2xx feature band branch to consume the shared component artifacts from a 1xx branch. For context, the shared component artifacts are just the source build artifacts produced by a build of the 1xx branch.

### Inputting the Shared Component Artifacts

There are two scenarios that are accounted for when providing the shared component artifacts as input to the build:

* VMR developer scenario: This involves the use of the `prep-source-build.sh` script. This script will automatically download the required shared component artifacts tarball. The URL for this tarball is computed using the non-stable version of the `Microsoft.NET.Sdk` package defined in the root `Version.Details.xml` file of the VMR. For example:
  ```xml
  <Dependency Name="Microsoft.NET.Sdk" Version="10.0.100-preview.7.25364.110">
    <Sha>afa79da3b68fd0fe48d90ea7fcbffbf7789c6d7f</Sha>
  </Dependency>
  ```
  The file that is downloaded will match the RID of the executing environment.
* Distro maintainer: Distro maintainers won't be using the prep script. Rather, they will be "bringing their own" shared component artifact tarball from the build they had done of the 1xx branch. This will be provided as input to the `build.sh` script via the `--with-shared-components` option.

### Package Consumption

The packages of the shared components need to be treated separately from the "normal" previously source-built artifacts for a few reasons:
* The shared component artifacts tarball includes packages produced from the entire output of the 1xx branch, including tooling repos like roslyn and sdk. These are not to be consumed as input to a build of the 1xx build.
* Poison leak detection needs to allow redistribution of packages from the shared components since that is their whole purpose. But packages from the "normal" previously source-built artifacts are specifically not to be redistributed.

The isolation of these two sets of packages is handled by extracting the shared component packages into a separate `prereqs/packages/previously-source-built-shared-components` directory which is then referenced by the various package source configurations.

The shared component packages need to take priority over reference packages of the same name from the 2xx branch. For example, the build of the 2xx branch should use the lifted `System.Text.Json` package version that comes from shared component packages instead of the `System.Text.Json` package version coming from SBRP of the 2xx branch. However, special validation logic is also added to ensure we don't have something misconfigured that would cause a conflict of some other kind. Except for reference packages, we should never have the version of a shared component package override the version of a package produced from within the 2xx build.

### Runtime Tarball Consumption

The build of the sdk repo requires redistributing the `dotnet-runtime` tarballs and `aspnetcore-runtime` tarballs. In builds of the 1xx branch, the assumption is that the runtime tarball artifact was produced earlier in the build and output to the `artifacts/assets` directory. To make this work for the 2xx branch, we make use of the shared component artifacts tarball which also [includes the runtime tarballs](https://github.com/dotnet/dotnet/pull/1480). These tarballs are extracted out and placed at the expected location within the `artifacts/assets` directory. This allows the rest of the build to proceed as usual allowing those files to be retrieved when needed.

cc @dotnet/distro-maintainers

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -35,6 +35,7 @@ echo "  --binaryLog                         Create MSBuild binary log (short: -b
   echo "  --source-repository <URL>         Source Link repository URL, required when building from tarball"
   echo "  --source-version <SHA>            Source Link revision, required when building from tarball"
   echo "  --with-packages <DIR>             Use the specified directory of previously-built packages"
+  echo "  --with-shared-components <FILE>   Use the specified shared components artifacts tarball file"
   echo "  --with-sdk <DIR>                  Use the SDK in the specified directory for bootstrapping"
   echo "  --prep                            Run prep-source-build.sh to download bootstrap binaries before building"
   echo ""
@@ -84,6 +85,7 @@ sourceRepository=''
 sourceVersion=''
 CUSTOM_PACKAGES_DIR=''
 CUSTOM_SDK_DIR=''
+sharedComponentsArchivePath=''
 packagesDir="$scriptroot/prereqs/packages/"
 packagesArchiveDir="${packagesDir}archive/"
 packagesPreviouslySourceBuiltDir="${packagesDir}previously-source-built/"
@@ -192,9 +194,18 @@ while [[ $# > 0 ]]; do
     -with-packages)
       CUSTOM_PACKAGES_DIR="$(cd -P "$2" && pwd)"
       if [ ! -d "$CUSTOM_PACKAGES_DIR" ]; then
-          echo "Custom prviously built packages directory '$CUSTOM_PACKAGES_DIR' does not exist"
+          echo "Custom previously built packages directory '$CUSTOM_PACKAGES_DIR' does not exist"
           exit 1
       fi
+      shift
+      ;;
+    -with-shared-components)
+      sharedComponentsArchivePath="$2"
+      if [ ! -f "$sharedComponentsArchivePath" ]; then
+          echo "Shared components archive '$sharedComponentsArchivePath' does not exist"
+          exit 1
+      fi
+      properties+=( "/p:PreviouslySourceBuiltSharedComponentsArchivePath=$sharedComponentsArchivePath" )
       shift
       ;;
     -with-sdk)
--- a/eng/VmrLayout.props
+++ b/eng/VmrLayout.props
@@ -36,6 +36,7 @@
     <PreviouslySourceBuiltPackagesPath>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'previously-source-built'))</PreviouslySourceBuiltPackagesPath>
     <PreviouslySourceBuiltPackagesPath Condition="'$(CustomPreviouslySourceBuiltPackagesPath)' != ''">$([MSBuild]::EnsureTrailingSlash('$(CustomPreviouslySourceBuiltPackagesPath)'))</PreviouslySourceBuiltPackagesPath>
     <PreviouslySourceBuiltPackageVersionsPropsFile>$(PreviouslySourceBuiltPackagesPath)PackageVersions.props</PreviouslySourceBuiltPackageVersionsPropsFile>
+    <PreviouslySourceBuiltSharedComponentsPackagesPath>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'previously-source-built-shared-components'))</PreviouslySourceBuiltSharedComponentsPackagesPath>
 
     <SbrpRepoSrcDir>$([MSBuild]::NormalizeDirectory('$(SrcDir)', 'source-build-reference-packages', 'src'))</SbrpRepoSrcDir>
     <ReferencePackagesDir>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'reference'))</ReferencePackagesDir>
--- a/eng/finish-source-only.proj
+++ b/eng/finish-source-only.proj
@@ -43,6 +43,7 @@
 
       <SourceBuiltPackageFile Include="$(ArtifactsPackagesDir)**/*.nupkg" />
       <SourceBuiltPackageFile Include="$(PreviouslySourceBuiltPackagesPath)*.nupkg" />
+      <SourceBuiltPackageFile Include="$(PreviouslySourceBuiltSharedComponentsPackagesPath)*.nupkg" />
       <ReferencePackageFile Include="$(ReferencePackagesDir)**/*.nupkg" />
 
       <!-- Check all RIDs from all restored Microsoft.NETCore.Platforms packages. -->
--- a/eng/init-poison.proj
+++ b/eng/init-poison.proj
@@ -21,6 +21,7 @@
     <ItemGroup>
       <PrebuiltPackages Include="$(PrebuiltPackagesPath)**/*.nupkg" />
       <PreviouslySourceBuiltPackages Include="$(PreviouslySourceBuiltPackagesPath)**/*.nupkg" />
+      <PreviouslySourceBuiltPackages Include="$(PreviouslySourceBuiltSharedComponentsPackagesPath)**/*.nupkg" />
     </ItemGroup>
 
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Poisoning existing packages for leak detection." />
--- a/eng/init-source-only.proj
+++ b/eng/init-source-only.proj
@@ -15,6 +15,7 @@
 
   <PropertyGroup>
     <ExternalTarballsDir>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'archive'))</ExternalTarballsDir>
+    <ExternalSharedComponentsTarballsDir>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'archive-shared-components'))</ExternalSharedComponentsTarballsDir>
   </PropertyGroup>
 
   <Target Name="Build"
@@ -33,6 +34,40 @@
           WorkingDirectory="$(PreviouslySourceBuiltPackagesPath)" />
   </Target>
 
+  <ItemGroup>
+    <SharedComponentsPrereqsArchivePath
+      Include="$(ExternalSharedComponentsTarballsDir)$(SourceBuiltArtifactsTarballName).*$(ArchiveExtension)" />
+  </ItemGroup>
+
+  <!--
+    Unpacks the previously source built archive produced from the 1xx feature band.
+    There are two possible sources for this archive:
+    1. The archive is specified by the user via the with-shared-components option in the build script.
+       This is what distro maintainers would typically use. This maps to the PreviouslySourceBuiltSharedComponentsArchivePath
+       property. This source takes precedence.
+    2. The archive is contained in the prereqs/packages/archive-shared-components directory.
+       This is what the prep script does by default. This provides a seamless experience for devs working in
+       the VMR, not requiring them to pass extra parameters by default when working in a non-1xx branch.
+       This scenario maps to the SharedComponentsPrereqsArchivePath item.
+  -->
+  <Target Name="UnpackSourceBuiltSharedComponentsArtifactsArchive"
+          Condition="'$(PreviouslySourceBuiltSharedComponentsArchivePath)' != '' or @(SharedComponentsPrereqsArchivePath) != ''">
+    <MakeDir Directories="$(PreviouslySourceBuiltSharedComponentsPackagesPath)" />
+
+    <PropertyGroup>
+      <_SharedComponentsArchivePath Condition="'$(PreviouslySourceBuiltSharedComponentsArchivePath)' != ''">$(PreviouslySourceBuiltSharedComponentsArchivePath)</_SharedComponentsArchivePath>
+      <_SharedComponentsArchivePath Condition="'$(_SharedComponentsArchivePath)' == ''">@(SharedComponentsPrereqsArchivePath)</_SharedComponentsArchivePath>
+    </PropertyGroup>
+
+    <!-- Extract everything except the assets directory to the shared components packages path -->
+    <Exec Command="tar -xzf $(_SharedComponentsArchivePath) --exclude='assets'"
+          WorkingDirectory="$(PreviouslySourceBuiltSharedComponentsPackagesPath)" />
+
+    <!-- Extract only the assets directory directly to the artifacts directory -->
+    <Exec Command="tar -xzf $(_SharedComponentsArchivePath) assets"
+          WorkingDirectory="$(ArtifactsDir)" />
+  </Target>
+
   <!-- Check for a prebuilt dependency tarball and extract if exists. If there isn't one, we expect
        the build to be working without prebuilts. -->
   <Target Name="UnpackSourceBuiltPrebuiltsArchive"
@@ -49,6 +84,7 @@
 
   <Target Name="UnpackTarballs"
           DependsOnTargets="UnpackSourceBuiltArtifactsArchive;
+                            UnpackSourceBuiltSharedComponentsArtifactsArchive;
                             UnpackSourceBuiltPrebuiltsArchive" />
 
   <!-- Build the custom msbuild sdk resolver. -->
--- a/eng/tools/Directory.Build.props
+++ b/eng/tools/Directory.Build.props
@@ -4,8 +4,9 @@
 
   <!-- Don't use these feeds when testing. Projects under this directory are referenced by
        test projects. This makes sure that the feeds aren't used. -->
-  <PropertyGroup>
-    <RestoreSources Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(IsTestRun)' != 'true'">$(PrebuiltPackagesPath);$(PreviouslySourceBuiltPackagesPath)</RestoreSources>
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(IsTestRun)' != 'true'">
+    <RestoreSources>$(PrebuiltPackagesPath);$(PreviouslySourceBuiltPackagesPath)</RestoreSources>
+    <RestoreSources Condition="Exists('$(PreviouslySourceBuiltSharedComponentsPackagesPath)')">$(RestoreSources);$(PreviouslySourceBuiltSharedComponentsPackagesPath)</RestoreSources>
   </PropertyGroup>
 
 </Project>
--- a/prep-source-build.sh
+++ b/prep-source-build.sh
@@ -36,12 +36,14 @@ function print_help () {
     sed -n '/^### /,/^$/p' "$source" | cut -b 5-
 }
 
+packagesDir="$REPO_ROOT/prereqs/packages"
+
 # SB prep default arguments
 defaultArtifactsRid='centos.10-x64'
 
 # Binary Tooling default arguments
 defaultDotnetSdk="$REPO_ROOT/.dotnet"
-defaultPackagesDir="$REPO_ROOT/prereqs/packages/previously-source-built"
+defaultPsbDir="$packagesDir/previously-source-built"
 
 # SB prep arguments
 buildBootstrap=true
@@ -56,7 +58,11 @@ runtime_source_feed_key='' # IBM requested these to support s390x scenarios
 
 # Binary Tooling arguments
 dotnetSdk=$defaultDotnetSdk
-packagesDir=$defaultPackagesDir
+psbDir=$defaultPsbDir
+
+artifactsBaseFileName="Private.SourceBuilt.Artifacts"
+prebuiltsBaseFileName="Private.SourceBuilt.Prebuilts"
+artifactsTarballPattern="$artifactsBaseFileName.*.tar.gz"
 
 positional_args=()
 while :; do
@@ -105,7 +111,7 @@ while :; do
       shift
       ;;
     --with-packages)
-      packagesDir=$2
+      psbDir=$2
       shift
       ;;
     *)
@@ -118,7 +124,7 @@ done
 
 # Attempting to bootstrap without an SDK will fail. So either the --no-sdk flag must be passed
 # or a pre-existing .dotnet SDK directory must exist.
-if [ "$buildBootstrap" == true ] && [ "$installDotnet" == false ] && [ ! -d "$REPO_ROOT/.dotnet" ]; then
+if [ "$buildBootstrap" == true ] && [ "$installDotnet" == false ] && [ [ ! -d "$REPO_ROOT/.dotnet" ]; then
   echo "  ERROR: --no-sdk requires --no-bootstrap or a pre-existing .dotnet SDK directory.  Exiting..."
   exit 1
 fi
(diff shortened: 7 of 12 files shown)


test/Microsoft.DotNet.SourceBuild.Tests/SymbolsTests.cs
```diff
@@ -23,7 +23,8 @@ public SymbolsTests(ITestOutputHelper outputHelper) : base(outputHelper) { }
     /// <summary>
     /// Verifies that all symbols have valid sourcelinks.
     /// </summary>
-    [Fact]
+    // Temporarily disabled due until https://github.com/dotnet/dotnet/issues/1465 is fixed
+    //[Fact]
```
MichaelSimons: Can this be left enabled in main?  I would hate to see a regression introduced in main when the failing scenario is only applicable in the 2xx dev branch.

build.sh
```diff
@@ -35,6 +35,7 @@ echo "  --binaryLog                         Create MSBuild binary log (short: -b
   echo "  --source-repository <URL>         Source Link repository URL, required when building from tarball"
   echo "  --source-version <SHA>            Source Link revision, required when building from tarball"
   echo "  --with-packages <DIR>             Use the specified directory of previously-built packages"
+  echo "  --with-shared-components <FILE>   Use the specified shared components artifacts tarball file"
```
MichaelSimons: Would it make sense to call out that these are from a different SDK feature band?

eng/VmrLayout.props
```diff
@@ -36,6 +36,7 @@
     <PreviouslySourceBuiltPackagesPath>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'previously-source-built'))</PreviouslySourceBuiltPackagesPath>
     <PreviouslySourceBuiltPackagesPath Condition="'$(CustomPreviouslySourceBuiltPackagesPath)' != ''">$([MSBuild]::EnsureTrailingSlash('$(CustomPreviouslySourceBuiltPackagesPath)'))</PreviouslySourceBuiltPackagesPath>
     <PreviouslySourceBuiltPackageVersionsPropsFile>$(PreviouslySourceBuiltPackagesPath)PackageVersions.props</PreviouslySourceBuiltPackageVersionsPropsFile>
+    <PreviouslySourceBuiltSharedComponentsPackagesPath>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'previously-source-built-shared-components'))</PreviouslySourceBuiltSharedComponentsPackagesPath>
```
MichaelSimons: I see `previously-source-built-shared-components` is defined in multiple locations, as was `previously-source-built`, can this be defined in a shared location, e.g. the VMRs directory.build.props?

eng/init-poison.proj
```diff
@@ -21,6 +21,7 @@
     <ItemGroup>
       <PrebuiltPackages Include="$(PrebuiltPackagesPath)**/*.nupkg" />
       <PreviouslySourceBuiltPackages Include="$(PreviouslySourceBuiltPackagesPath)**/*.nupkg" />
+      <PreviouslySourceBuiltPackages Include="$(PreviouslySourceBuiltSharedComponentsPackagesPath)**/*.nupkg" />
```
MichaelSimons: This doesn't seem right/complete.  Thoughts on reverting this now to include all required poison changes with https://github.com/dotnet/dotnet/issues/1112?

eng/init-source-only.proj
```diff
@@ -15,6 +15,7 @@
 
   <PropertyGroup>
     <ExternalTarballsDir>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'archive'))</ExternalTarballsDir>
+    <ExternalSharedComponentsTarballsDir>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'archive-shared-components'))</ExternalSharedComponentsTarballsDir>
```
MichaelSimons: Why does this need to be placed in a separate directory?  Can't a common archive directory be used?  In my personal workflow, I like to preserve this dir while cleaning so that I don't have to re-pull it as part of prep.  Having one dir to preserve is nice versus multiple.

eng/init-source-only.proj
```diff
@@ -33,6 +34,40 @@
...
+    1. The archive is specified by the user via the with-shared-components option in the build script.
+       This is what distro maintainers would typically use. This maps to the PreviouslySourceBuiltSharedComponentsArchivePath
+       property. This source takes precedence.
+    2. The archive is contained in the prereqs/packages/archive-shared-components directory.
+       This is what the prep script does by default. This provides a seamless experience for devs working in
+       the VMR, not requiring them to pass extra parameters by default when working in a non-1xx branch.
+       This scenario maps to the SharedComponentsPrereqsArchivePath item.
+  -->
+  <Target Name="UnpackSourceBuiltSharedComponentsArtifactsArchive"
+          Condition="'$(PreviouslySourceBuiltSharedComponentsArchivePath)' != '' or @(SharedComponentsPrereqsArchivePath) != ''">
```
MichaelSimons: This model appears different than the source-built-artifacts.  With the PSB, the --with-packages is a directory where is PSB tarball is extracted.  This difference feels like a potential source of confusion.  Can we unify how these are handled?

build.sh
mthalman: Based on your comment at https://github.com/dotnet/dotnet/pull/1373#issuecomment-3025513896, I thought we didn't want to use that kind of description.

eng/VmrLayout.props
mthalman: You're referring to this spot and this one, I assume: https://github.com/dotnet/dotnet/pull/1502/files#diff-c4c7386806d2707bdf65085bbf28501450a36181d035aa9a306cae52afd6f2f0R175

There is no reason they need to have the same value. One refers to a directory name and the other is a NuGet source name. They don't need to have the same value.

eng/init-source-only.proj
mthalman: That's fair. I'll probably just integrate the prebuilt directory in that as well.

eng/VmrLayout.props
MichaelSimons: I missed that - thanks for the clarification.  No changes needed.

build.sh
MichaelSimons: My view was that this may help the UX for new distro maintainers to understand when to use this option.  I am only asking for the option description to be expanded.  I see this as different than the name of the property used throughout the build logic.

eng/init-source-only.proj
mthalman: Actually, I won't change the `prebuilt` directory; I'll leave it where it is. That's because it doesn't represent input to the build from the user. Rather, it will contain the extracted contents of the prebuilts tarball. In reality, it shouldn't exist at all in source. I'll leave this to https://github.com/dotnet/source-build/issues/5272.

test/Microsoft.DotNet.SourceBuild.Tests/SymbolsTests.cs
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/13162c1841bcce81a38649d19f7f601aa5b6b6ab

build.sh
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/2d77562daa812c5393bbf5b7d6dd6f767cb205c6

eng/init-poison.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/8fdc989ada978f4cf69f429ea6ff062c5c38df6a

eng/init-source-only.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/5965e629ed41cfe7d394e2269e347b8ad2a282b3

eng/init-source-only.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/2d77562daa812c5393bbf5b7d6dd6f767cb205c6

build.sh
```diff
@@ -84,8 +85,9 @@ sourceRepository=''
 sourceVersion=''
 CUSTOM_PACKAGES_DIR=''
 CUSTOM_SDK_DIR=''
+sharedComponentsDir=''
```
MichaelSimons: nit:  consider removing this.  The default value isn't used and the variable is only used in one spot.

eng/init-source-only.proj
```diff
@@ -26,20 +27,74 @@
...
+    <Exec Command="tar -xzf $(CommonArchivesDir)$(SourceBuiltArtifactsTarballName).*$(ArchiveExtension)"
           WorkingDirectory="$(PreviouslySourceBuiltPackagesPath)" />
   </Target>
 
+  <ItemGroup>
+    <SharedComponentsPrereqsArchivePath
+      Include="$(SharedComponentArchivesDir)$(SourceBuiltArtifactsTarballName).*$(ArchiveExtension)" />
+  </ItemGroup>
+
+  <Target Name="ReferenceSharedComponents"
```
MichaelSimons: Consider naming this so that it is tied more directly to CustomSharedComponents.  IMO that would help readability.  e.g. InitCustomSharedComponents, CopyCustomSharedComponents

eng/init-source-only.proj
```diff
@@ -26,20 +27,74 @@
...
+    There are two possible sources for these artifacts:
+    1. The directory to the extracted artifacts is specified by the user via the with-shared-components option in the build script.
+       This is what distro maintainers would typically use. This maps to the CustomSharedComponentsArtifactsPath
+       property. This source takes precedence.
+    2. The archive is contained in the prereqs/packages/archives/shared-components directory.
+       This is what the prep script does by default. This provides a seamless experience for devs working in
+       the VMR, not requiring them to pass extra parameters by default when working in a non-1xx branch.
+       This scenario maps to the SharedComponentsPrereqsArchivePath item.
+  -->
+  <Target Name="PrepareSharedComponents"
```
MichaelSimons, line 98: Is there a way to provide a better UX around validation?  For example, it would be ideal to error if shared components are specified in a 1xx environment.  Additionally, if shared components are not specified in >1xx environment, the build should fail up front.  The primary reasons for this is it will be easier to understand these errors versus the more obscure build errors that may result from this type of misconfiguration.

eng/init-source-only.proj
```diff
@@ -26,20 +27,74 @@
...
+          DestinationFolder="$(ArtifactsDir)/assets/%(RecursiveDir)"
+          UseSymbolicLinksIfPossible="true"
+          Condition="'@(SharedComponentsAssets)' != ''" />
+  
+  </Target>  
+
+  <Target Name="UnpackSharedComponentsPrereqsArchive"
+          Condition="'@(SharedComponentsPrereqsArchivePath)' != '' and '$(CustomSharedComponentsArtifactsPath)' == ''"
+          Inputs="$(SharedComponentArchivesDir)"
+          Outputs="$(SharedComponentsArtifactsPath)">
```
MichaelSimons: From an incremental build perspective, this output doesn't seem complete as it is missing the artifact assets.

eng/init-source-only.proj
```diff
@@ -14,7 +14,8 @@
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
 
   <PropertyGroup>
-    <ExternalTarballsDir>$([MSBuild]::NormalizeDirectory('$(PrereqsPackagesDir)', 'archive'))</ExternalTarballsDir>
+    <CommonArchivesDir>$([MSBuild]::NormalizeDirectory('$(PrereqsPackageArchivesDir)', 'common'))</CommonArchivesDir>
```
MichaelSimons: Did you consider using a single directory?  The prep script could name the files with differing prefixes.  e.g. `Private.SourceBuilt.Artifacts` vs `Private.SourceBuilt.Artifacts.ShareComponents`. A single directory makes more sense specifically in a 1xx build.

mthalman: I think I'll use just `Private.SourceBuilt.SharedComponents`. If I use `Private.SourceBuilt.Artifacts.ShareComponents`, it'll have the same prefix as `Private.SourceBuilt.Artifacts` which will really complicate things since the build uses wildcard search patterns on that.

eng/finish-source-only.proj
```diff
@@ -43,6 +43,7 @@
 
       <SourceBuiltPackageFile Include="$(ArtifactsPackagesDir)**/*.nupkg" />
       <SourceBuiltPackageFile Include="$(PreviouslySourceBuiltPackagesPath)*.nupkg" />
+      <SourceBuiltPackageFile Include="$(SharedComponentsArtifactsPath)*.nupkg" />
```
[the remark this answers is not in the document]
mthalman: Don't need to worry about subdirectories.

eng/tools/Directory.Build.props
```diff
@@ -4,8 +4,9 @@
 
   <!-- Don't use these feeds when testing. Projects under this directory are referenced by
        test projects. This makes sure that the feeds aren't used. -->
-  <PropertyGroup>
-    <RestoreSources Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(IsTestRun)' != 'true'">$(PrebuiltPackagesPath);$(PreviouslySourceBuiltPackagesPath)</RestoreSources>
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(IsTestRun)' != 'true'">
+    <RestoreSources>$(PrebuiltPackagesPath);$(PreviouslySourceBuiltPackagesPath)</RestoreSources>
+    <RestoreSources Condition="Exists('$(SharedComponentsArtifactsPath)')">$(RestoreSources);$(SharedComponentsArtifactsPath)</RestoreSources>
```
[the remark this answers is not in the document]
mthalman: The suggestion looks more complicated than is desired.

build.sh
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/dd7d1bf98c3ea829bc894924219b62314b6f100c

eng/init-source-only.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/70b6c5c72371462943bc9f80fae0f233891705e7

eng/init-source-only.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/b554dfa6b0e92475e14f3b78940609e5c1e159e1

eng/init-source-only.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/32aad0dbfa84a93bfbf2773fbfe2d040c905a14d

eng/init-source-only.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/49d2d97455ade6e86aa6ee49174948369d471401

repo-projects/Directory.Build.targets
```diff
@@ -327,37 +333,82 @@
...
+    <ItemGroup>
+      <_ToolingComponents Include="@(_PreviouslyBuiltSourceBuiltSharedComponentPackages)"
+                          Condition="!$(SharedRepositoryReferenceString.Contains(';%(RepoOrigin);')) or %(RepoOrigin) == 'arcade' or %(RepoOrigin) == 'source-build-reference-packages'" />
+      <_PreviouslyBuiltSourceBuiltSharedComponentPackages Remove="@(_ToolingComponents)" MatchOnMetadata="Version" />
+    </ItemGroup>
+
+    <PropertyGroup>
+      <_UsePackageVersionPropsAggregation>true</_UsePackageVersionPropsAggregation>
+      <_UsePackageVersionPropsAggregation Condition="'$(DotNetSourceOnlyTestOnly)' == 'true'">false</_UsePackageVersionPropsAggregation>
+      <_PreviouslyBuiltSourceBuiltSharedComponentPackagesString>@(_PreviouslyBuiltSourceBuiltSharedComponentPackages)</_PreviouslyBuiltSourceBuiltSharedComponentPackagesString>
```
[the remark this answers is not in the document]
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/b6dfcbb01bbfe4f1b4f86d8aeff1be3d21a7a52d

eng/init-source-only.proj
```diff
@@ -26,20 +27,74 @@
...
+    <Copy SourceFiles="@(SharedComponentsAssets)"
+          DestinationFolder="$(ArtifactsDir)/assets/%(RecursiveDir)"
+          UseSymbolicLinksIfPossible="true"
+          Condition="'@(SharedComponentsAssets)' != ''" />
+  
+  </Target>  
+
+  <Target Name="UnpackSharedComponentsPrereqsArchive"
+          Condition="'@(SharedComponentsPrereqsArchivePath)' != '' and '$(CustomSharedComponentsArtifactsPath)' == ''"
+          Inputs="$(SharedComponentArchivesDir)"
```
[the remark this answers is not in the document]
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/32aad0dbfa84a93bfbf2773fbfe2d040c905a14d

build.sh
```diff
@@ -203,11 +204,20 @@ while [[ $# > 0 ]]; do
     -with-packages)
       CUSTOM_PACKAGES_DIR="$(cd -P "$2" && pwd)"
       if [ ! -d "$CUSTOM_PACKAGES_DIR" ]; then
-          echo "Custom prviously built packages directory '$CUSTOM_PACKAGES_DIR' does not exist"
+          echo "Custom previously built packages directory '$CUSTOM_PACKAGES_DIR' does not exist"
           exit 1
       fi
       shift
       ;;
+    -with-shared-components)
```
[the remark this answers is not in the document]
mthalman: won't fix

prep-source-build.sh
```diff
@@ -240,15 +299,25 @@ if [ "$installDotnet" == true ]; then
...
+    BootstrapArtifacts "$packagesArchiveDir"
   fi
 fi
 
+if [ "$downloadSharedComponentsArtifacts" == true ]; then
+  source $REPO_ROOT/eng/common/native/init-os-and-arch.sh
+  source $REPO_ROOT/eng/common/native/init-distro-rid.sh
+  initDistroRidGlobal "$os" "$arch" ""
+
+  DownloadArchive "shared component artifacts" "MicrosoftNETSdkVersion" false "$__DistroRid" "$packagesArchiveDir" "$sharedComponentsBaseFileName"
```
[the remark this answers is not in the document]
mthalman: This is handled correctly by the `DownloadArchive` function.

eng/init-source-only.proj
```diff
@@ -33,6 +33,73 @@
           WorkingDirectory="$(PreviouslySourceBuiltPackagesPath)" />
   </Target>
 
+  <ItemGroup>
+    <SharedComponentsPrereqsArchivePathItem Include="$(ExternalTarballsDir)$(SourceBuiltSharedComponentsTarballName).*$(ArchiveExtension)" />
+  </ItemGroup>
+  <PropertyGroup>
+    <SharedComponentsPrereqsArchivePath>@(SharedComponentsPrereqsArchivePathItem)</SharedComponentsPrereqsArchivePath>
```
MichaelSimons: Thought on using the file suffix versus path?  This would be more consistent with SourceBuiltPrebuiltsTarballFile.

MichaelSimons (approved):

prep-source-build.sh
```diff
@@ -131,17 +138,23 @@ then
...
-  echo "  Private.SourceBuilt.Artifacts.*.tar.gz exists...it will not be downloaded"
-  downloadArtifacts=false
+  echo "  $artifactsTarballPattern exists in $packagesArchiveDir...it will not be downloaded"
+  downloadPsbArtifacts=false
+fi
+
+# Check if Private.SourceBuilt shared components archive exists
+downloadSharedComponentsArtifacts=$downloadArtifacts
+if [ "$downloadArtifacts" == true ] && [ -f ${packagesArchiveDir}${sharedComponentsBaseFileName}.*.tar.gz ]; then
+  echo "  $sharedComponentsTarballPattern exists in $packagesArchiveDir...it will not be downloaded"
```
ellahathaway, line 155: `sharedComponentsTarballPattern` is not defined.

prep-source-build.sh
```diff
@@ -131,17 +138,23 @@ then
...
+  downloadSharedComponentsArtifacts=false
 fi
 
 # Check if Private.SourceBuilt prebuilts archive exists
-prebuiltsBaseFileName="Private.SourceBuilt.Prebuilts"
 if [ "$downloadPrebuilts" == true ] && [ -f ${packagesArchiveDir}${prebuiltsBaseFileName}.*.tar.gz ]; then
-  echo "  Private.SourceBuilt.Prebuilts.*.tar.gz exists...it will not be downloaded"
+  echo "  $prebuiltsBaseFileName.*.tar.gz exists in $packagesArchiveDir...it will not be downloaded"
```
ellahathaway: Consider maintaining consistency across these types of error messages. The above two messages use `<name>TarballPattern exists in $packagesArchiveDir...it will not be downloaded`

prep-source-build.sh
```diff
@@ -151,75 +164,121 @@ if [ "$installDotnet" == true ] && [ -d "$REPO_ROOT/.dotnet" ]; then
...
-          esac
-        fi
-      done
-    )
-  elif [ "$isRequired" == true ]; then
-    echo "  ERROR: $notFoundMessage"
-    exit 1
+  local archiveUrl
+  if [[ "$propertyName" == "MicrosoftNETSdkVersion" ]]; then
+    archiveUrl="https://ci.dot.net/public/source-build/$artifactsBaseFileName.$archiveVersion.$artifactsRid.tar.gz"
```
ellahathaway, line 234: This is confusing to me. Can you explain why this file uses the artifacts base name and not the shared components base name? Do the shared artifacts contain the PSB artifacts?

build.sh
```diff
@@ -37,6 +37,7 @@ usage()
   echo "  --source-repository <URL>         Source Link repository URL, required when building from tarball"
   echo "  --source-version <SHA>            Source Link revision, required when building from tarball"
   echo "  --with-packages <DIR>             Use the specified directory of previously-built packages"
+  echo "  --with-shared-components <DIR>    Use the specified shared components artifacts (e.g. from a build of a 1xx branch)"
```
ellahathaway, line 40: > Distro maintainers won't be using the prep script. Rather, they will be "bringing their own" shared component artifact tarball from the build they had done of the 1xx branch. This will be provided as input to the build.sh script via the --with-shared-components option.
 
Given this explanation, I don't think that the following scenario is very likely, but I'll mention it anyways: if someone runs `./build.sh --prep --with-shared-components <dir>`, I think prep would unpack the default shared components and then the build would use the custom shared components. This doesn't seem like the desired behavior. IMO, prep should not download the shared components in this case.

prep-source-build.sh
mthalman: All that the shared components are is the artifacts tarball published from a 1xx build. Nothing special about it. It's just our usual artifacts tarball output. So there's no file to download with a "shared components" name.

build.sh
mthalman: Using the `--prep` in the build script is for very basic scenarios where you just want as few steps as possible and for things to just work. I highly doubt you would use `--with-shared-components` in that scenario. If you did, I don't have an issue with the script downloading extra stuff you don't need. Don't use the `--prep` option in that case. The build would use the custom shared components in this case which is the correct behavior. Given all that, I'm not inclined to complicate the scripts to avoid this.

eng/init-source-only.proj
mthalman: I'm not understanding what you're suggesting. This is just converting the item to a property. The item is defined above with:

```xml
<SharedComponentsPrereqsArchivePathItem Include="$(ExternalTarballsDir)$(SourceBuiltSharedComponentsTarballName).*$(ArchiveExtension)" />
```

This is the same pattern that is used for resolve the common artifact tarball a little further above:

```xml
<Exec Command="tar -xzf $(ExternalTarballsDir)$(SourceBuiltArtifactsTarballName).*$(ArchiveExtension)"
```

prep-source-build.sh
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/00da9e6303407d42e9c71ffeb5c8a5909d037810

prep-source-build.sh
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/00da9e6303407d42e9c71ffeb5c8a5909d037810

eng/init-source-only.proj
MichaelSimons: I am simply referring to updating the property name to be `SharedComponentsPrereqsArchiveFile`

mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1502/commits/3c11f4c04e79cb871ac7c7a1d79f58c51066f38e

ellahathaway (approved):

PR #1535, opened 2025-07-18
Consume branding from source-build manifest

mthalman:
Contributes to https://github.com/dotnet/source-build/issues/5109

Part of the distro maintainer UX for configuring official builds is to set the branding option. This should be provided within the source-build release manifest and read from when that manifest is provided as input to the build script.

This updates the build script to read that value and set the `DotNetFinalVersionKind` MSBuild appropriately.

It also refactors the use of OfficialBuildId to follow the same pattern. This includes removing the behavior which checked whether `OfficialBuildId` was explicitly provided as an MSBuild property in the build script arguments. My assertion is that we shouldn't check for these properties when we already have specific build script arguments for such values (in this case `--official-build-id` from https://github.com/dotnet/dotnet/pull/1497).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -58,6 +58,33 @@ usage()
   echo "Arguments can also be passed in with a single hyphen."
 }
 
+function SetBranding()
+{
+  local brandingValue="$1"
+  
+  if [[ "$brandingValue" == "preview" ]]; then
+    properties+=( "/p:DotNetFinalVersionKind=prerelease" )
+  elif [[ "$brandingValue" == "rtm" ]]; then
+    properties+=( "/p:DotNetFinalVersionKind=release" )
+  elif [[ "$brandingValue" == "default" ]]; then
+    # default branding; no extra property needed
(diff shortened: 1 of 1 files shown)


mmitche (approved):

build.sh
```diff
@@ -58,6 +58,33 @@ usage()
   echo "Arguments can also be passed in with a single hyphen."
 }
 
+function SetBranding()
+{
+  local brandingValue="$1"
+  
+  if [[ "$brandingValue" == "preview" ]]; then
```
MichaelSimons, line 65: I know this was existing code.... Where to RCs fit?  They are in a way a middle ground between preview and rtm especially with a golive license.

MichaelSimons (approved):

build.sh
mthalman: @mmitche to comment on this

mmitche: They're still "preview".

Issue #1403, opened 2025-07-07
Signed builds fail running scenario tests


mthalman:
All scenario tests are failing with errors similar to the following:

```
  [FAIL] Microsoft.DotNet.ScenarioTests.SdkTemplateTests.SdkTemplateTests.VerifyConsoleTemplateComplexPortable(language: VB)
  System.TypeInitializationException : The type initializer for 'System.Net.Sockets.SocketsTelemetry' threw an exception.
  ---- System.BadImageFormatException : Could not load file or assembly 'System.Diagnostics.DiagnosticSource, Version=10.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. Reference assemblies cannot be loaded for execution. (0x80131058)
  -------- System.BadImageFormatException : Cannot load a reference assembly for execution.
     at System.Net.Sockets.Socket.ReceiveAsync(SocketAsyncEventArgs e, CancellationToken cancellationToken)
     at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ReceiveAsync(Socket socket, CancellationToken cancellationToken)
     at System.IO.Pipes.PipeStream.ReadAsyncCore(Memory`1 destination, CancellationToken cancellationToken)
     at System.Diagnostics.AsyncStreamReader.ReadBufferAsync()
     at System.Diagnostics.Process.WaitForExitCore(Int32 milliseconds)
     at System.Diagnostics.Process.WaitForExit(Int32 milliseconds)
     at Microsoft.DotNet.ScenarioTests.Common.ExecuteHelper.ExecuteProcess(String fileName, String args, ITestOutputHelper outputHelper, Boolean logOutput, Action`1 configure, Int32 millisecondTimeout) in /_/src/scenario-tests/src/Microsoft.DotNet.ScenarioTests.Common/ExecuteHelper.cs:line 60
     at Microsoft.DotNet.ScenarioTests.SdkTemplateTests.DotNetSdkHelper.ExecuteCmdImpl(String args, String workingDirectory, Action`1 additionalProcessConfigCallback, Int32 expectedExitCode, Int32 millisecondTimeout) in /_/src/scenario-tests/src/Microsoft.DotNet.ScenarioTests.SdkTemplateTests/DotNetSdkHelper.cs:line 46
[...]

PR #1555, opened 2025-07-21
Pre seek entry stream to 0 before computing content hash

mmitche:


mmitche:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2755062&view=results

src/arcade/src/Microsoft.DotNet.SignTool/src/ZipDataEntry.cs
```diff
@@ -37,7 +37,9 @@ public ZipDataEntry(string relativePath, Stream contentStream, long contentSize)
                 contentStream.CopyTo(_stream);
             }
 
-            // compute content hash and reset position back
+            // Compute content hash and reset position back
+            // Reset position to 0 prior to content hash computation.
+            _stream.Position = 0;
```
tarekgh: Nit: Should this be added only in the else case? I'm not sure if there's ever a scenario where `CanSeek` is true, but the stream doesn't start at position `0` and still has the correct read position. Forcing a reset in that case could potentially cause issues.

mmitche: Could be, true...maybe some odd archive type with a header per file. I'll hoist it.

mmitche: Though, if that were the case, resetting to zero would also cause issues, so it probably doesn't help

mmitche:
Resolves https://github.com/dotnet/dotnet/issues/1403

src/arcade/src/Microsoft.DotNet.SignTool/src/ZipDataEntry.cs
```diff
@@ -35,9 +35,11 @@ public ZipDataEntry(string relativePath, Stream contentStream, long contentSize)
                 // if we can't seek we need to copy the stream into a (seekable) MemoryStream so we can compute the content hash
                 _stream = new MemoryStream((int)contentSize);
```
ericstj, line 36: So what's the stream ownership contract with the caller?   Sometimes this object takes on ownership and will dispose it, other times the caller needs to "know" to dispose it?  Seems busted.

I would imagine it needs to always be delegated to this type.  That means this case that copies the stream should dispose the source stream.

Then you need to scrub the callsites to make sure they don't dispose.

mmitche: WIll deal with this scrub in a follow up PR. ZipDataEntry should always own the input stream

ericstj (approved):
Per discussion - the ownership issue can be addressed separately since it seems we're not hitting it here.

mmitche:
https://github.com/dotnet/arcade/issues/15985 for follow up item

akoeplinger:
@mmitche should we port this one to main? or do you want to do it combined with the arcade follow up item?

mmitche:
> @mmitche should we port this one to main? or do you want to do it combined with the arcade follow up item?

Working on an arcade follow up PR.

mmitche:
https://github.com/dotnet/arcade/pull/15989

PR #1556, opened 2025-07-22
Do not publish symbols packages in source-only build

NikolaMilosavljevic:
Fixes: https://github.com/dotnet/source-build/issues/4453

Currently repos are publishing, to disk, in source-only build, all `*.symbols.nupkg`. Additionally, `roslyn` repo is publishing 9 custom named symbols packages in the following format: `<id>.Symbols.<version>.nupkg`.

This PR:
- Updates arcade to bring back the filtering of symbols nupkgs in source-only build. This was originally implemented with https://github.com/dotnet/arcade/pull/14559, but has since regressed.
- Updates `roslyn`'s `Publishing.props` to additionally filter all custom named packages.
- Adds a source-only test to ensure that no symbols packages are included in previously source-build archive.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Publish.proj
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Publish.proj
@@ -138,8 +138,7 @@
 
   <Target Name="BeforePublish" Condition="'@(Artifact)' != ''">
     <ItemGroup>
-      <!-- Exclude all existing *.symbols.nupkg in source-only build - we create a unified symbols archive instead. -->
-      <_ExistingSymbolPackage Include="@(Artifact)" Condition="'$(DotNetBuildSourceOnly)' != 'true' and '%(Artifact.SkipPublish)' != 'true' and $([System.String]::Copy('%(Filename)%(Extension)').EndsWith('.symbols.nupkg'))">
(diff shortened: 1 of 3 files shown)


src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Publish.proj
```diff
@@ -149,6 +148,11 @@
       <_PackageToPublish Include="@(Artifact)" Exclude="@(_ExistingSymbolPackage)" Condition="'%(Artifact.SkipPublish)' != 'true' and '%(Extension)' == '.nupkg'" />
     </ItemGroup>
 
+    <ItemGroup>
+      <!-- Exclude all existing *.symbols.nupkg in source-only build - we create a unified symbols archive instead. -->
+      <_ExistingSymbolPackage Remove="@(_ExistingSymbolPackage)" Condition="'$(DotNetBuildSourceOnly)' == 'true'"/>
```
mthalman, line 153: I don't understand why we add the symbols and then just remove all them if it's source-only. Why not just include the source-only check on the initial adding of the symbols?

NikolaMilosavljevic: `_PackageToPublish` has always contained all packages, including symbol ones. During publishing we filter this further. I'm just following that model and fixing the regression in this infra.

mthalman (approved):

PR #1566, opened 2025-07-22
Switch non-prod comparison pipeline to use the unofficial template

mmitche:


akoeplinger (approved):
LGTM. I assume we could also disable binskim/policheck/tsa now?

PR #1572, opened 2025-07-22
Use RepoOriginalSourceRevisionId in ASP.NET Core

BrennanConroy:
This should reenable us to have repo specific commit diffs for benchmarking regression/improvement analysis.

Will probably want to do the same in dotnet/runtime assuming this works.

Question: Do we only want to do this for nightlies? Should it be disabled for official builds?

Edit: Maybe we want `OriginalRepoCommitUrl` as well?
e.g. Today we have:
```
[assembly: AssemblyMetadata("CommitHash", "67889d9d2f21a890ac29bcd175c0d1937a401781")]
[assembly: AssemblyMetadata("SourceCommitUrl", "https://github.com/dotnet/dotnet/tree/67889d9d2f21a890ac29bcd175c0d1937a401781")]
```

### DIFF
--- a/src/aspnetcore/eng/targets/CSharp.Common.targets
+++ b/src/aspnetcore/eng/targets/CSharp.Common.targets
@@ -94,7 +94,12 @@
         <_Parameter2>$(RepositoryUrl)/tree/$(SourceRevisionId)</_Parameter2>
       </AssemblyAttribute>
 
-      <AssemblyAttribute Include="System.Reflection.AssemblyMetadataAttribute" Condition="'$(Serviceable)' == 'true'">
+      <AssemblyAttribute Include="System.Reflection.AssemblyMetadataAttribute" Condition="'$(RepoOriginalSourceRevisionId)' != ''">
+        <_Parameter1>OriginalRepoCommitHash</_Parameter1>
+        <_Parameter2>$(RepoOriginalSourceRevisionId)</_Parameter2>
+      </AssemblyAttribute>
+
+    <AssemblyAttribute Include="System.Reflection.AssemblyMetadataAttribute" Condition="'$(Serviceable)' == 'true'">
         <_Parameter1>Serviceable</_Parameter1>
         <_Parameter2>True</_Parameter2>
       </AssemblyAttribute>


BrennanConroy:
Verified it worked:
image

wtgodbe (approved):
Nice!

src/aspnetcore/eng/targets/CSharp.Common.targets
```diff
@@ -94,7 +94,12 @@
         <_Parameter2>$(RepositoryUrl)/tree/$(SourceRevisionId)</_Parameter2>
       </AssemblyAttribute>
 
-      <AssemblyAttribute Include="System.Reflection.AssemblyMetadataAttribute" Condition="'$(Serviceable)' == 'true'">
+      <AssemblyAttribute Include="System.Reflection.AssemblyMetadataAttribute" Condition="'$(RepoOriginalSourceRevisionId)' != ''">
+        <_Parameter1>OriginalRepoCommitHash</_Parameter1>
```
akoeplinger, line 98: nit: maybe call it `AspNetCoreRepoCommitHash` so it's a bit clearer instead of "original" ?

akoeplinger (approved):

src/aspnetcore/eng/targets/CSharp.Common.targets
BrennanConroy: We want to do this for multiple repos and have code that grabs this metadata. It'd be very inconvenient to have different names for different repos.

ViktorHofer:
I'm not certain this information should get encoded into shipping artifacts. That SHA doesn't represent the shipping state. This information will be most certainly incorrect during servicing or in case of security fixes that are made directly against the VMR.

IOW this is an intermediate (an implementation detail about how we build the product) that shouldn't ship to customers.

> Question: Do we only want to do this for nightlies? Should it be disabled for official builds?

That would be an option. Or only adding this in Debug / non-Release Configuration builds?

akoeplinger:
@ViktorHofer I don't agree. When we added this feature we discussed this and everyone is aware that it is not 100% correct at all times, but its still very useful to know what the origin is.

BrennanConroy:
> > Question: Do we only want to do this for nightlies? Should it be disabled for official builds?
> 
> That would be an option. Or only adding this in Debug / non-Release Configuration builds?

What we're trying to improve is how easy it is to investigate performance issues. Debug/non-release builds aren't used for benchmarking, nightlies are used. So that could be a good option if we want to avoid the metadata in official builds.

mmitche:
Agreed. I think this is reasonable as an informational only.

Issue #1583, opened 2025-07-23
Why do nuget-client and vstest opt out of deterministic build?


omajid:
https://github.com/dotnet/dotnet/blob/c0e325f90fb79db0da6be5128dc292f2aabb264f/repo-projects/nuget-client.proj#L5

https://github.com/dotnet/dotnet/blob/c0e325f90fb79db0da6be5128dc292f2aabb264f/repo-projects/vstest.proj#L4

Is there a specific reason they opt out?

akoeplinger:
This was added in https://github.com/dotnet/source-build/pull/1168 a long time ago, unfortunately the PR doesn't state exactly why this was needed.

MichaelSimons:
It would be worthwhile to remove the props once and see if they are still needed and if so why.

omajid:
> It would be worthwhile to remove the props once and see if they are still needed and if so why.

I made a change locally and it still built fine, and the SDK was able to re-build itself. Let me do a PR and see if CI/tests can help answer.

PR #1584, opened 2025-07-23
Build nuget-client and vstest in deterministic mode

omajid:
We don't know why they opted out. Local builds seem to be just fine without this.

Closes: https://github.com/dotnet/dotnet/issues/1583


MichaelSimons:
If there are no more uses of DeterministicBuildOptOut, can [this logic](https://github.com/dotnet/dotnet/blob/main/repo-projects/Directory.Build.props#L178) get cleaned up as well?

MichaelSimons (approved):

Issue #1588, opened 2025-07-23
Add rolling Windows-only job on Scouting Queue


wtgodbe:
Will allow us to catch tooling issues before doing a machine rollout. We frequently get build breaks (especially around the summer) when the VS Toolset gets updated on the Windows build machines. This can often lead to multi-day scrambles as we try to figure out the root cause & get a fix - having early builds on the scout queue would give us more runway to prevent such issues.

wtgodbe:
Done: https://dev.azure.com/dnceng-public/public/_build?definitionId=321&_a=summary

PR #1589, opened 2025-07-23
Add rolling CI job running on the Scouting queue

wtgodbe:
https://github.com/dotnet/dotnet/issues/1588

Adds a Win-only job that runs on the scouting queue. Should allow us to catch tooling related build-breaks early

### DIFF
--- a/eng/pipelines/scout-build.yml
+++ b/eng/pipelines/scout-build.yml
@@ -1,0 +1,28 @@
+# This yml is used by these pipelines:
+# - dotnet-unified-build-scout (public)
+
+# Only run daily on schedule to save resources
+trigger: none
+
+schedules:
+  - cron: "0 13 * * *" # run at 013:00 (UTC)
+    branches:
+      include:
+      - main
+      - release/*
+    always: false # run only if there were changes since the last successful scheduled run.
+
+variables:
+- ${{ if eq(variables['System.TeamProject'], 'public') }}:
+  - name: skipComponentGovernanceDetection  # we run CG on internal builds only
+    value: true
+
+  - name: Codeql.Enabled  # we run CodeQL on internal builds only
+    value: false
+
+- template: /eng/common/templates/variables/pool-providers.yml@self
+
+stages:
+- template: /eng/pipelines/templates/stages/vmr-build.yml
+  parameters:
+    scope: scout
--- a/eng/pipelines/templates/stages/source-build-stages.yml
+++ b/eng/pipelines/templates/stages/source-build-stages.yml
@@ -10,6 +10,8 @@ parameters:
 - name: scope
   type: string
   values:
+  # run only windows jobs
+  - scout
   # run several legs e.g. stage1/stage2 for basic testing
   - lite
   # run everything
--- a/eng/pipelines/templates/variables/vmr-build.yml
+++ b/eng/pipelines/templates/variables/vmr-build.yml
@@ -18,6 +18,11 @@ parameters:
   type: string
   default: ''
 
+- name: isScoutingJob
+  displayName: Scope
+  type: boolean
+  default: false
+
 variables:
 - name: _TeamName
   value: DotNetCore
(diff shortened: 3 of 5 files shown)


eng/pipelines/templates/stages/source-build-stages.yml
```diff
@@ -10,6 +10,8 @@ parameters:
 - name: scope
   type: string
   values:
+  # run only windows jobs
```
MichaelSimons: This doesn't seem necessary - this is for source-build, no windows jobs are applicable.

wtgodbe: I was thinking that too, but put it in there for consistency w/ the other files. I'm fine with removing it.

eng/pipelines/templates/variables/vmr-build.yml
```diff
@@ -18,6 +18,11 @@ parameters:
   type: string
   default: ''
 
+- name: isScoutingJob
+  displayName: Scope
```
ViktorHofer: Is that right?

wtgodbe: nope 😆

eng/pipelines/scout-build.yml
```diff
@@ -0,0 +1,27 @@
+# This yml is used by these pipelines:
+# - dotnet-unified-build-scout (public)
+
+# Only run daily on schedule to save resources
```
ViktorHofer: Will this run even if there weren't any changes (new commits)?

wtgodbe: I assumed the default for `always` was `true`, but looks like it's `false`. I'll add it back

wtgodbe: https://github.com/dotnet/dotnet/pull/1589/commits/eec9081ca4d75bc49fde62e333ed369e1fca28c3

eng/pipelines/scout-build.yml
```diff
@@ -0,0 +1,28 @@
...
+trigger: none
+
+schedules:
+  - cron: "0 13 * * *" # run at 013:00 (UTC)
+    branches:
+      include:
+      - main
+      - release/*
+    always: true # run even if there are no new commits, in case the build images changed.
```
ViktorHofer: The downside here is that this will trigger for every release branch now, even super stale ones. This could result in quite a waste of resources.

What if we change the schedule from daily to weekly?

wtgodbe: Does the VMR not clean up its old github release branches?

In any case, .yml supports multiple schedules, so I can have an `always: true` one for `main` and an `always: false` one for release. I'll push a commit

wtgodbe: https://github.com/dotnet/dotnet/pull/1589/commits/37610ee8d049e27cdc3f7421a9dd4069e7a411c0

wtgodbe: Didn't see your edit - does dnceng always do rollouts on the same weekday? I think they might always do wednesdays - @markwilkie @garath @missymessa is that true? If so, I'd be fine with doing a weekly run only on the day after a rollout.

missymessa: We aim to do rollouts each Wednesday, but sometimes we don't for reasons. If you need to do your rollout after ours, I'd check for rollout announcements in case your rollouts depend directly on ours.

wtgodbe: Is it "always Wednesday, but not every Wednesday"? We're using a Cron schedule here, so these builds wouldn't be run manually. If the rollouts _only_ happen on Wednesdays then I think it's fine to run this pipeline every Thursday morning, and accept that some percentage of those weekly runs will be superfluous.

missymessa: I wouldn't even say it's "but not every Wednesday". Sometimes it's Thursday, or Tuesday, or even Monday, but never Friday. But the aim is for Wednesdays.

missymessa: Is there any reason you can't run it every day? Or even every other day?

wtgodbe: If dnceng is fine with us doing it every day, I think the current state of the PR is fine (daily for `main` even if there are no new commits, daily for `release/*` _only_ if there are new commits)

missymessa: If y'all know how long the pipelines take, we can estimate a dollar cost to it and see if that helps any decision making :)

wtgodbe: It'd be 3 jobs of about 90 minutes each, then 1 job of about 30 minutes. So 5 total job-hours per build.

missymessa: 5 machine hours is N.B.D. We're fine if you want to run this daily. If it becomes a problem, we'll see it in the metrics and we'll let you know :)

wtgodbe: Updated to run 3 days a week

PR #1593, opened 2025-07-23
Re-Bootstrap Source Build to .NET 10.0.100-preview.7.25372.107

jkoritzinsky:
Re-Bootstrap Source Build to .NET 10.0.100-preview.7.25372.107

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -2,13 +2,13 @@
   <ProductDependencies>
   </ProductDependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.25322.101">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="10.0.0-beta.25372.107">
       <Uri>https://github.com/dotnet/dotnet</Uri>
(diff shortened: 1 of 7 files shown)


ViktorHofer:
Just to double check, this is a signed build from the weekend?

jkoritzinsky:
Good catch. I'll redo this with a signed build.

MichaelSimons:
> Just to double check, this is a signed build from the weekend?

I know the re-bootstrap pipeline was not used here, but I think it by default should enforce signed builds.  I logged https://github.com/dotnet/source-build/issues/5295 to track this.

jkoritzinsky:
Actually, this build should be signed (as it's from a release branch and those are always signed), so this should work.

lewing:
@jkoritzinsky what is the status here?  it looks like runtime flow has been blocked for a week and we need the base64 fix in aspnetcore

jkoritzinsky:
I'm iterating through the various failures from rebootstrapping (looks primarily to be from smarter Roslyn analyzers for repos that don't build in source-build).

lewing:
> I'm iterating through the various failures from rebootstrapping (looks primarily to be from smarter Roslyn analyzers for repos that don't build in source-build).

Thanks for working on this.

akoeplinger (approved):

ViktorHofer (approved):

PR #1613, opened 2025-07-24
Poisoning support for shared components

### LINKED ISSUE #1112: [feature bands] Poison leak detection

mthalman:
Fixes #1112

This adds source build poisoning support for multi-band SDK scenarios in the VMR. In this scenario, runtime-related packages that are provided as input to a non-1xx branch via a shared components tarball are considered to be acceptable for redistribution by the SDK produced from that branch. That is, of course, the whole intention of providing these packages because they are part of the working product. However, the tooling-related packages that comes from the shared components tarball are not considered acceptable to redistribute because the non-1xx branch is the one that is producing the new tooling-related packages. This means that runtime-related packages are _not_ marked for poisoning but the tooling-related packages are.

Notes on the changes:
* Added extra poisoning checks which validates that at least one package exists to be poisoned from both the previously source built artifacts and the shared component packages.
* Refactored `UpdateNuGetConfigPackageSourcesMappings` so that it can handle an arbitrary number of "fallback" inputs (prebuilts, previously source built, shared components).
* Created a new `GetFilteredSharedComponentPackages` target that can filter the packages from the shared components according to whether they come from runtime- or tooling-related repos. This is shared by both repo-projects (for build input props generation) and poisoning. This is defined in `VmrCommon.targets` and needed to be separately referenced in two different locations because of the need for it to be usable in the case of the `Publish` target. In `Publish`, the `Directory.Build.targets` of the VMR root isn't in scope because the project being built is from Arcade. So it needs to be explicitly imported by the `Publishing.props` file.

### DIFF
--- a/eng/Publishing.props
+++ b/eng/Publishing.props
@@ -1,6 +1,7 @@
 <Project InitialTargets="DiscoverArtifacts">
   <Import Project="$(RepositoryEngineeringDir)RuntimeIdentifier.props" />
   <Import Project="$(RepositoryEngineeringDir)VmrLayout.props" />
+  <Import Project="$(RepositoryEngineeringDir)VmrCommon.targets" />
 
   <PropertyGroup>
     <PublishingVersion>4</PublishingVersion>
--- a/eng/VmrCommon.targets
+++ b/eng/VmrCommon.targets
@@ -1,0 +1,47 @@
+<Project>
+  
+  <UsingTask TaskName="Microsoft.DotNet.UnifiedBuild.Tasks.GetKnownArtifactsFromAssetManifests" AssemblyFile="$(MicrosoftDotNetUnifiedBuildTasksAssembly)" TaskFactory="TaskHostFactory" />
+
+  <Target Name="GetFilteredSharedComponentPackages"
+          Outputs="@(_SharedComponentFilteredPackages)">
+
+    <PropertyGroup>
+      <_PreviouslySourceBuiltSharedComponentAssetManifests>$(SharedComponentsArtifactsPath)VerticalManifest.xml</_PreviouslySourceBuiltSharedComponentAssetManifests>
+      <SharedRepositoryReferenceString>;@(SharedRepositoryReference);</SharedRepositoryReferenceString>
+      <!-- Property to control filtering mode: true = include only tooling components, false = exclude tooling components (default) -->
+      <IncludeOnlyToolingComponents Condition="'$(IncludeOnlyToolingComponents)' == ''">false</IncludeOnlyToolingComponents>
+    </PropertyGroup>
+
+    <GetKnownArtifactsFromAssetManifests AssetManifests="$(_PreviouslySourceBuiltSharedComponentAssetManifests)"
+                                         Condition="Exists($(_PreviouslySourceBuiltSharedComponentAssetManifests))">
+      <Output TaskParameter="KnownPackages" ItemName="_SharedComponentFilteredPackages" />
+    </GetKnownArtifactsFromAssetManifests>
+
+    <!-- Filter tooling components based on the mode -->
+    <ItemGroup Condition="'$(IncludeOnlyToolingComponents)' == 'false'">
+      <!-- Default mode: Remove tooling components from the shared packages -->
+      <_ItemsToRemove Include="@(_SharedComponentFilteredPackages)"
+                      Condition="!$(SharedRepositoryReferenceString.Contains(';%(RepoOrigin);')) or %(RepoOrigin) == 'arcade' or %(RepoOrigin) == 'source-build-reference-packages'" />
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(IncludeOnlyToolingComponents)' == 'true'">
+      <!-- Alternate mode: Keep only tooling components, remove everything else -->
+      <_ItemsToRemove Include="@(_SharedComponentFilteredPackages)"
+                      Condition="$(SharedRepositoryReferenceString.Contains(';%(RepoOrigin);')) and %(RepoOrigin) != 'arcade' and %(RepoOrigin) != 'source-build-reference-packages'" />
+    </ItemGroup>
+
+    <!-- Create a lookup string of items to remove -->
+    <PropertyGroup>
+      <_ItemsToRemoveString>@(_ItemsToRemove->'%(Identity)::%(Version)', ';')</_ItemsToRemoveString>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <_SharedComponentFilteredPackages Remove="@(_SharedComponentFilteredPackages)" 
+                                        Condition="'@(_ItemsToRemove)' != '' and 
+                                                '$(_ItemsToRemoveString)' != '' and
+                                                $(_ItemsToRemoveString.Contains($([System.String]::Concat('%(Identity)', '::', '%(Version)'))))" />
+    </ItemGroup>
+
+  </Target>
+
+</Project>
--- a/eng/init-poison.proj
+++ b/eng/init-poison.proj
@@ -23,6 +23,27 @@
       <PreviouslySourceBuiltPackages Include="$(PreviouslySourceBuiltPackagesPath)**/*.nupkg" />
     </ItemGroup>
 
+    <Error Condition="'@(PreviouslySourceBuiltPackages)' == ''" Text="Expected NuGet packages to be present within the path '$(PreviouslySourceBuiltPackagesPath)'." />
+
+    <!-- Get the packages that come from the tooling repos in the shared components. Only those packages should be marked for poisoning.
+         The packages from the runtime-related components should not be marked for poisoning as it is intentional for them to be
+         redistributed in the SDK. -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
+             Targets="GetFilteredSharedComponentPackages"
+             Condition="'$(RepositoryName)' != 'source-build-reference-packages' and '$(RepositoryName)' != 'arcade'"
+             Properties="IncludeOnlyToolingComponents=true">
+      <Output TaskParameter="TargetOutputs" ItemName="_FilteredSharedComponentPackages" />
+    </MSBuild>
+
+    <Error Condition="'@(_FilteredSharedComponentPackages)' == ''" Text="Expected NuGet packages to be present from shared components artifacts." />
+
+    <ItemGroup>
+      <_SharedComponentFilenames Include="@(_FilteredSharedComponentPackages)">
+        <PackagePath>$(SharedComponentsArtifactsPath)$([System.IO.Path]::GetFileName('%(PipelineArtifactPath)'))</PackagePath>
+      </_SharedComponentFilenames>
+      <PreviouslySourceBuiltPackages Include="@(_SharedComponentFilenames->'%(PackagePath)')" />
+    </ItemGroup>
+
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Poisoning existing packages for leak detection." />
 
     <MarkAndCatalogPackages PackagesToMark="@(PrebuiltPackages)" CatalogOutputFilePath="$(PoisonReportDataFile)" MarkerFileName="$(PoisonMarkerFile)" />
--- a/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
+++ b/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
@@ -21,8 +21,9 @@ namespace Microsoft.DotNet.UnifiedBuild.Tasks
     /*
      * This task updates the package source mappings in the NuGet.Config using the following logic:
      * Add all packages from current source-build sources, i.e. source-built-*, reference-packages.
-     * For previously source-built sources (PSB), add only the packages that do not exist in any of the current source-built sources.
-     * Also add PSB packages if that package version does not exist in current package sources.
+     * For fallback sources (e.g. prebuilts, previously source built), add only the packages that
+     * do not exist in any of the current source-built sources.
+     * Also add fallback packages if that package version does not exist in current package sources.
      * In offline build, remove all existing package source mappings for online sources.
      * In online build, add online source mappings for all discovered packages from local sources.
      * In online build, if NuGet.config didn't originally have any mappings, additionally,
@@ -56,9 +57,7 @@ namespace Microsoft.DotNet.UnifiedBuild.Tasks
 
         public string ReferencePackagesSourceName { get; set; }
 
-        public string PreviouslySourceBuiltSourceName { get; set; }
-
-        public string PrebuiltSourceName { get; set; }
+        public string[] FallbackSourceNames { get; set; }
 
         public string[] CustomSources { get; set; }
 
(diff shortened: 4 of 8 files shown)


eng/Publishing.props
```diff
@@ -1,6 +1,7 @@
 <Project InitialTargets="DiscoverArtifacts">
   <Import Project="$(RepositoryEngineeringDir)RuntimeIdentifier.props" />
   <Import Project="$(RepositoryEngineeringDir)VmrLayout.props" />
+  <Import Project="$(RepositoryEngineeringDir)VmrCommon.targets" />
```
ViktorHofer: Why is this import necessary?

mthalman: I guess it's not. Fixed in https://github.com/dotnet/dotnet/pull/1613/commits/003b10c5cc7cb84c76c460673d31b87466270b75

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
```diff
@@ -261,23 +259,21 @@ private XElement GetPackageMappingsElementForSource(string packageSource)
...
                 }
-                else if (packageSource.Equals(PreviouslySourceBuiltSourceName))
-                {
-                    AddPackageSourceMappingIfPackageVersionsNotInCurrentPackages(pkgSrc, packagePattern, previouslySourceBuiltPackages);
-                }
-                else if (packageSource.Equals(PrebuiltSourceName))
+                else if (isFallbackSource)
                 {
-                    AddPackageSourceMappingIfPackageVersionsNotInCurrentPackages(pkgSrc, packagePattern, prebuiltPackages);
+                    AddPackageSourceMappingIfPackageVersionsNotInCurrentPackages(pkgSrc, packagePattern, fallbackPackages);
```
NikolaMilosavljevic: Would this now treat all fallback sources the same? I.e. they all get the same mapping even if they don't have a package with specific id and version?

mthalman: PSB and prebuilts were already treated the same in this class. I've just generalized it to use a single collection of source names so that I expand it to include shared components as well.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
```diff
@@ -21,8 +21,9 @@ namespace Microsoft.DotNet.UnifiedBuild.Tasks
     /*
      * This task updates the package source mappings in the NuGet.Config using the following logic:
      * Add all packages from current source-build sources, i.e. source-built-*, reference-packages.
-     * For previously source-built sources (PSB), add only the packages that do not exist in any of the current source-built sources.
-     * Also add PSB packages if that package version does not exist in current package sources.
+     * For fallback sources (e.g. prebuilts, previously source built), add only the packages that
+     * do not exist in any of the current source-built sources.
+     * Also add fallback packages if that package version does not exist in current package sources.
```
NikolaMilosavljevic: We should maybe have a new test case - specifically to cover a scenario where only one of the fallback sources have a package with version that does not exist in live sources.

eng/init-poison.proj
```diff
@@ -23,6 +23,27 @@
...
     </ItemGroup>
 
+    <Error Condition="'@(PreviouslySourceBuiltPackages)' == ''" Text="Expected NuGet packages to be present within the path '$(PreviouslySourceBuiltPackagesPath)'." />
+
+    <!-- Get the packages that come from the tooling repos in the shared components. Only those packages should be marked for poisoning.
+         The packages from the runtime-related components should not be marked for poisoning as it is intentional for them to be
+         redistributed in the SDK. -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
+             Targets="GetFilteredSharedComponentPackages"
+             Condition="'$(RepositoryName)' != 'source-build-reference-packages' and '$(RepositoryName)' != 'arcade'"
```
MichaelSimons: The number of places that are conditioning on SBRP and arcade continues to grow.  These repos are the special UseBootStrapArcade repos.  The UseBootStrapArcade property is defined in the repo-project and cannot be utilized here.  Should we define a common itemGroup that defines these repo names that can be utilized in these places rather than hard coding the repos?  I feel like this would help (marginally) the readability of the system.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
mthalman: Why is that an interesting scenario?

eng/init-poison.proj
```diff
@@ -23,6 +23,27 @@
...
+             Properties="IncludeOnlyToolingComponents=true">
+      <Output TaskParameter="TargetOutputs" ItemName="_FilteredSharedComponentPackages" />
+    </MSBuild>
+
+    <Error Condition="'@(_FilteredSharedComponentPackages)' == ''" Text="Expected NuGet packages to be present from shared components artifacts." />
+
+    <ItemGroup>
+      <_SharedComponentFilenames Include="@(_FilteredSharedComponentPackages)">
```
MichaelSimons: Consider renaming to better indicate the contents - e.g. `_SharedToolingComponentFilenames`

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
NikolaMilosavljevic: Isn't `fallbackPackages` representing a union of all fallback packages? Previously we'd only check and potentially add a specific fallback source mapping, for packages that exist in specific fallback source only.

eng/init-poison.proj
```diff
@@ -23,6 +23,27 @@
...
+        <PackagePath>$(SharedComponentsArtifactsPath)$([System.IO.Path]::GetFileName('%(PipelineArtifactPath)'))</PackagePath>
+      </_SharedComponentFilenames>
+      <PreviouslySourceBuiltPackages Include="@(_SharedComponentFilenames->'%(PackagePath)')" />
```
MichaelSimons: From a UX perspective when tracking down leaks, wouldn't it be more helpful to classify the shared components different than PSB?  I feel like that would aid in the speed of tracking down leaks.

repo-projects/Directory.Build.targets
```diff
@@ -255,15 +247,28 @@
...
+    <ItemGroup>
+      <_BuildSources Include="$(PrebuiltNuGetSourceName);
+                              $(PreviouslySourceBuiltNuGetSourceName);
+                              $(PreviouslySourceBuiltSharedComponentsNuGetSourceName);
+                              $(ReferencePackagesNuGetSourceName)"
+                     Condition="'$(AddPrebuiltFeeds)' == 'true'"/>
+      <_BuildSources Include="@(_CommonBuildSources)" />
+      <_BuildSources Include="$(NetSdkSupportingFeedName)"
+                     Condition="'$(AddNetSdkSupportingFeed)' == 'true'" />
+      <_FallbackSources Include="$(PrebuiltNuGetSourceName);
```
MichaelSimons: Is this going to introduce non-determinism as well as potential poison leaks if the same package version exists in multiple feeds?  Don't these sources need to be in priority order?

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
NikolaMilosavljevic: This is related to my other comment. I might be mistaken, but it appears that all fallback sources are now getting a mapping for a package that has a different version than in live sources, even if only one fallback source contains the package with that version.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
mthalman: Fallback sources only consist of prebuilts, PSB, and shared components. Each of them is compared to the "current/live" source. So if package `System.Text.Json.9.0.10` exists in current and in prebuilts (or PSB or shared components), it won't get a mapping for the given fallback source.

MichaelSimons: What if a package version doesn't exist in live but exists in both PSB and shared components.  Is it deterministic on which source the package will get picked up from?

eng/VmrCommon.targets
ViktorHofer, line 1: So I was asking the other question before cause if no one other than D.B.targets imports this file, I would rather inline the target there then create a new file for which it isn't clear when to put stuff into it vs other files.

IOW I would recommend to move this logic into D.B.targets

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
NikolaMilosavljevic: The scenario is the following. PSB source contains `PackageA.1` while live source contains `PackageA.2`. Based on old logic we would add mapping for `PackageA` to PSB source, but not for `prebuilt` source. I believe new logic would add this mapping for all fallback sources. Since mappings are not version-specific, we could end up pulling `PackageA.2` from another fallback source, i.e. `shared-components`.

NikolaMilosavljevic: > What if a package version doesn't exist in live but exists in both PSB and shared components. Is it deterministic on which source the package will get picked up from?

There is no determinism for this case. All sources are valid for a specific package version, if package mappings exist for both sources.

mthalman: > The scenario is the following. PSB source contains `PackageA.1` while live source contains `PackageA.2`. Based on old logic we would add mapping for `PackageA` to PSB source, but not for `prebuilt` source. I believe new logic would add this mapping for all fallback sources. Since mappings are not version-specific, we could end up pulling `PackageA.2` from another fallback source, i.e. `shared-components`.

@NikolaMilosavljevic - I see your point now. Thanks for being patient with me. 😉

mthalman: > What if a package version doesn't exist in live but exists in both PSB and shared components. 

This will require special logic to only provide a source mapping for shared components and not for PSB. Otherwise, it could lead to a poison leak if it restored from the PSB source.

mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1613/commits/ffbd75bce398f6dfa2779281220d40e0e19b9ca4

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1613/commits/ffbd75bce398f6dfa2779281220d40e0e19b9ca4

eng/init-poison.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1613/commits/60183705eae123f4130d14dd35e0c471a5858383

eng/init-poison.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1613/commits/d1754d87ac0cdbf0e80f10aeb803a80eb9938c32

eng/init-poison.proj
mthalman: The paths in the catalog show the difference in where they're coming from. But I've also updated to use a separate item group which will be useful when examining the binlog. Fixed in https://github.com/dotnet/dotnet/pull/1613/commits/95e703979b9818c31ea8a33e3a094999b6c1224a.

repo-projects/Directory.Build.targets
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1613/commits/ffbd75bce398f6dfa2779281220d40e0e19b9ca4

eng/VmrCommon.targets
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1613/commits/bb9f274c8d03d725ddd918d1a0ccb41732d7984a

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/UpdateNuGetConfigPackageSourcesMappings.cs
```diff
@@ -265,19 +269,21 @@ private XElement GetPackageMappingsElementForSource(string packageSource)
...
+                    AddPackageSourceMappingIfPackageVersionsNotInSources(pkgSrc, packagePattern, previouslySourceBuiltPackages, checkSharedComponents: true);
                 }
                 else if (packageSource.Equals(PrebuiltSourceName))
                 {
-                    AddPackageSourceMappingIfPackageVersionsNotInCurrentPackages(pkgSrc, packagePattern, prebuiltPackages);
+                    AddPackageSourceMappingIfPackageVersionsNotInSources(pkgSrc, packagePattern, prebuiltPackages, checkSharedComponents: true);
+                }
+                else if (packageSource.Equals(SharedComponentsSourceName))
+                {
+                    AddPackageSourceMappingIfPackageVersionsNotInSources(pkgSrc, packagePattern, sharedComponentsPackages, checkSharedComponents: false);
```
NikolaMilosavljevic: Nit - maybe we could pass `packageSource` to `AddPackageSourceMappingIfPackageVersionsNotInSources` and decide if we check shared components based on that value, in the method.

MichaelSimons (approved):

PR #1618, opened 2025-07-24
Flow Deterministic=true to repos

omajid:
This *should* be a no-op until Deterministic supported is re-enabled via https://github.com/NuGet/Home/issues/8601.

### DIFF
--- a/repo-projects/Directory.Build.props
+++ b/repo-projects/Directory.Build.props
@@ -175,6 +175,7 @@
     <EnvironmentVariables Include="_InitializeToolset=$(SourceBuiltSdksDir)Microsoft.DotNet.Arcade.Sdk/tools/Build.proj"
                           Condition="'$(UseBootstrapArcade)' != 'true'" />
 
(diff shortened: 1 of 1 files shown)


omajid:
I suppose an alternative implementation is to set this property and this value at each repo level, instead of setting it here in the top-level build.

ViktorHofer:
I don't think there is a need for this. The .NET SDK already defaults Deterministic to true: https://github.com/dotnet/sdk/blob/86e7689b689c8ee7ac76998fa2c00dedf1e0518f/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props#L42

jaredpar:
> I don't think there is a need for this. The .NET SDK already defaults Deterministic to true:

Agree. Rather than set this property globally I think we should add a build / PR check that fails if any shipping project does not have it set at compilation time.

omajid:
Thanks for the feedback!

PR #1635, opened 2025-07-25
Only download specific sets of artifacts

mmitche:
Fixes validation issues in some legs. We were picking up the SB artifacts


mmitche:
Validated here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2758138&view=logs&j=5aa0659d-d34b-5c54-db98-526d154cc18c&t=4d32bbf5-a0c0-546c-5d7a-d95b98fa65f4

PR #1639, opened 2025-07-25
Add support to track external packages in the sbrpUsageReport

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/3405

Additional changes are still to come for https://github.com/dotnet/source-build/issues/3405 to track version numbers.


eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/WriteSbrpUsageReport.cs
```diff
@@ -24,12 +25,22 @@ public class WriteSbrpUsageReport : Task
 
     private readonly Dictionary<string, PackageInfo> _sbrpPackages = [];
 
+    private readonly Regex _packageNameVersionRegex = new Regex(
```
ViktorHofer: We recommend to use source generated Regex instead when targeting modern .NET. But this isn't a performance critical path so should be fine.

mthalman (approved):

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/WriteSbrpUsageReport.cs
MichaelSimons: Fixed with https://github.com/dotnet/dotnet/pull/1639/commits/4798028fe3a61e657b8d96f434326b19ed69a34a.

ViktorHofer (approved):

PR #1658, opened 2025-07-28
Remove runtime/emsdk SignCheckExclusionsFile.txt

akoeplinger:
These repos don't have an official build anymore so only the VMR's copy of the file is actually used. Remove the repo files.


akoeplinger:
Ideally we'd do this for aspnetcore and sdk as well but they still have their official build right now and removing the file would break it.

ViktorHofer (approved):

PR #1660, opened 2025-07-28
Revert "[release/10.0.1xx-preview7] Source code updates from dotnet/msbuild"

mmitche:
Reverts dotnet/dotnet#1655


baronfel:
This justification for this is that the codeflow that [half-reverted the FileWritesShareable logic](https://github.com/dotnet/dotnet/pull/1660/files#diff-6b1e42ff25229f18e20af2761220cd18eb6bb160beae9d7cb76600dd6bdf4179R5792) reverted a fix to the initial implementation, but left the memory increase _and_ the clean-logic bug.  This revert of that codeflow means that p7

* will have slightly higher memory usage
* but will not entirely break incremental cleaning (which happens on every build/publish/pack/etc) for all projects

mmitche:
@baronfel can you approve, if this should go in?

baronfel:
Apologies - I meant to when I left the comment yesterday. Clear to go!

PR #1662, opened 2025-07-28
Sync emscripten workload version with common runtime version

mthalman:
Fixes https://github.com/dotnet/source-build/issues/5298

Allows source-build to override the Emscripten package version. Since all package versions are based off the `OfficialBuildId`, we know that the Emscripten packages will match the package versions of the runtime which are correctly lifted in the `sdk` repo. So I've chosen STJ as the one to match since that is likely to always exist as a dependency of the repo.

### DIFF
--- a/repo-projects/sdk.proj
+++ b/repo-projects/sdk.proj
@@ -58,4 +58,8 @@
     <KeepFeedPrefixes Include="darc-int-dotnet-aspnetcore-" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
+    <ExtraPackageVersionPropsPackageInfo Include="EmscriptenWorkloadsVersion" Version="%24(SystemTextJsonVersion)" />
+  </ItemGroup>
+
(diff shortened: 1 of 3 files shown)


repo-projects/sdk.proj
```diff
@@ -58,4 +58,8 @@
     <KeepFeedPrefixes Include="darc-int-dotnet-aspnetcore-" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
+    <ExtraPackageVersionPropsPackageInfo Include="EmscriptenWorkloadsVersion" Version="%24(SystemTextJsonVersion)" />
```
MichaelSimons, line 62: This isn't ideal, it raises a question for me if we are missing a construct for this type of scenario.  Perhaps it is really a one off.  Regardless of the answer, this is good for P7.

mmitche, line 62: We should avoid the `ExtraPackageVersionPropsPackageInfo` where we can, if possible.

PR #1689, opened 2025-07-30
Flow shared component symbols

### LINKED ISSUE #1465: [feature bands] Include 1xx symbols with 2xx symbol output
A build of the 2xx SDK will include the 1xx runtime. As part of the output for source build, symbols tarballs are output. It's important that the symbols for 1xx runtime are included with this output of a 2xx build.

mthalman:
Fixes #1465

In order to produce an accurate symbols tarball from a source-only build of non-1xx branch of the VMR, we need to flow the symbols of the shared components (runtime, ASP.NET Core, etc) that were produced from the 1xx branch. This is necessary because a build of the non-1xx branch doesn't build any of those shared component repos so it obviously wouldn't output any symbols for them on its own.

This is implemented by manipulating the `obj/Symbols` directories as part of build initialization so that it is pre-populated with the shared component symbols. These are retrieved from the artifacts tarball of the related 1xx build. These symbols are placed in a special sub-directory called `shared-components`. This is necessary specifically because we must also include the `source-build-reference-packages` repo from the shared component symbols even though the non-1xx branch builds its own version of that repo. So having these in the `shared-components` sub-directory avoids file conflicts. Inclusion of these SBRP symbols is important because some of the assemblies from the 1xx build could have references to an external repo assembly from SBRP. For example, the `dotnet-user-jwts` tool produced by `aspnetcore` has a reference to `Microsoft.IdentityModel.Abstractions`, which is built from an external repo assembly of SBRP. We need the exact PDB associated with that IdentityModel assembly to be flowed. But we don't want the build of SBRP in the non-1xx to overwrite that PDB with its own version. That won't produce a matching reference. By having two separate versions of the PDBs within the tarball, it solves this issue.


eng/init-source-only.proj
```diff
@@ -79,6 +79,37 @@
...
+    <!-- Remove all directories except those from shared components. We also want to keep those from the Arcade bootstrap
+         repos, specifically SBRP, because SBRP is built in each feature band branch and so there may be differences between 
+         the packages between those branches. Each SBRP assembly would have its own PDB so these are kept separated. -->
+    <ItemGroup>
+      <_AllSymbolDirs Include="$([System.IO.Directory]::GetDirectories($(_SharedComponentsSymbolsDir)))" />
+      <_RepoSymbolsToKeep Include="@(SharedRepositoryReference);$(BootstrapArcadeRepos)" />
+      <_SymbolDirsToDelete Include="@(_AllSymbolDirs)" Exclude="@(_RepoSymbolsToKeep->'$(_SharedComponentsSymbolsDir)/%(Identity)')" />
+    </ItemGroup>
+
+    <RemoveDir Directories="@(_SymbolDirsToDelete)" Condition="'@(_SymbolDirsToDelete)' != ''" />
```
NikolaMilosavljevic, line 111: When would we have these additional dirs? They would be created later during build of each repo.

mthalman: The symbols coming from the 1xx build consist of all the repos, including the tooling repos like roslyn and sdk. These symbols from the tooling repos are not needed in the context of the non-1xx build and so they are explicitly deleted to avoid bloat. The non-1xx build will produce its _own_ symbols for the tooling repos. _These_ will be included.

NikolaMilosavljevic: Oh, yeah, that makes sense. I misread the code originally - I thought it's deleting dirs in root `Symbols` dir.

PR #1690, opened 2025-07-30
Do not conditionalize EmscriptenWorkloadsVersion on SB

mmitche:


akoeplinger (approved):

repo-projects/sdk.proj
```diff
@@ -58,7 +58,8 @@
     <KeepFeedPrefixes Include="darc-int-dotnet-aspnetcore-" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
+  <ItemGroup>
+    <!-- Temporary. https://github.com/dotnet/dotnet/pull/1686 has the real fix-->
     <ExtraPackageVersionPropsPackageInfo Include="EmscriptenWorkloadsVersion" Version="%24(SystemTextJsonVersion)" />
```
marcpopMSFT, line 63: Probably don't want to use the STJ version since that historically hasn't versioned every release. Something like MicrosoftNETCoreAppRefPackageVersion that versions every release would be better assuming the goal is to have it match the runtime version.

mmitche: Yeah this is just temporary. Though we do ship it every version now. This line should get removed altogether.

PR #1701, opened 2025-07-31
Get the current commit SHA locally

Winniexu01:
Fixes: https://github.com/dotnet/source-build/issues/5199

When running locally, `RepositoryCommit` is assigned the value `0000000000000000000000000000000000000000` when `$(BUILD_SOURCEVERSION)` is empty, making it impossible to correctly retrieve the current commit information. To address this, I have added logic to ensure the actual commit SHA using `git rev-parse HEAD` in local environments. With this change, commit information can be consistently accessed both in CI/CD pipelines and during local development, improving reliability and developer experience.

Pipeline run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2761704 (internal Microsoft link)


ViktorHofer:
Won't this fail the build when git isn't installed?

Winniexu01:
No, if git is not installed, the build will not fail. MSBuild will ignore the failure of this command and continue building.

mthalman (commented):
Can you also update this test so that it would have failed with the prior behavior:
https://github.com/dotnet/dotnet/blob/66d49fe8f010e3af61e6b61b19eab38b11829ec6/test/Microsoft.DotNet.SourceBuild.Tests/SourceBuiltArtifactsTests.cs#L23

eng/finish-source-only.proj
```diff
@@ -177,6 +177,13 @@
   <Target Name="WriteSourceBuiltSdkVersionFile"
           AfterTargets="Build"
           DependsOnTargets="DetermineSourceBuiltSdkNonStableVersion">
+    <!-- Get the current commit SHA locally -->
+    <Exec
+      Command="git rev-parse HEAD 2>&amp;1"
+      Condition="'$(BUILD_SOURCEVERSION)' == ''"
+      StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="RepositoryCommit" />
```
mthalman, line 185: If an error occurs, we don't want the error message to be used as the commit value. Instead, it should replace the error message with a placeholder value. Proposed value: `unknown commit SHA`.

Winniexu01: Update: https://github.com/dotnet/dotnet/pull/1701/commits/6f0ab8fd957cee06321b544bd02041f4406bd93a

eng/finish-source-only.proj
```diff
@@ -177,6 +177,18 @@
...
+    </Exec>
+
+    <PropertyGroup>
+      <RepositoryCommit Condition="$(RepositoryCommit.StartsWith('git:')) Or $(RepositoryCommit.StartsWith('fatal:'))">unknown commit SHA</RepositoryCommit>
```
mthalman: I'd prefer a more explicit approach to see whether it's a valid SHA using a Regex.

Winniexu01: Update: https://github.com/dotnet/dotnet/pull/1701/commits/bc7102d4e8aed4e2c6e9acdd3786dc4e0eaee31f

mthalman (approved):

PR #1707, opened 2025-07-31
Add non-production pipeline and refactor official pipeline

mmitche:
- For production readiness compliance
- Move official pipeline to be named official.yml
- Add support for more signing options (test and dry-run). Refactor variables template to set up these options.
- Add unofficial.yml for non-prod pipeline
    - Unofficial pipeline allows for lite and full options, and test signed, unsigned, or dry run options.
- Introduce parameter to templates indicating that the pipeline is official. Set in official.yml


mmitche:
This PR is currently blocked on:
- Needing a re-bootstrap to pick up https://github.com/dotnet/arcade/commit/2586309f3f8553152e2d1d54612e9199975f9cd4
- Will temporarily revert the ci.yml -> official.yml change to verify official builds.

eng/pipelines/unofficial.yml
```diff
@@ -0,0 +1,156 @@
...
+
+resources:
+  repositories:
+  - repository: 1ESPipelineTemplates
+    type: git
+    name: 1ESPipelineTemplates/1ESPipelineTemplates
+    ref: refs/tags/release
+
+extends:
+  template: v1/1ES.Unofficial.PipelineTemplate.yml@1ESPipelineTemplates
```
mthalman, line 66: It'd be great if we can share all the parameters being passed to this template since they are come between this and `official.yml`. Is it possible to create a wrapper around 1ES templates that can do this? That would require parameterizing the 1ES template to be used (e.g. `v1/1ES.Unofficial.PipelineTemplate.yml@1ESPipelineTemplates`).

mmitche: It's not possible to create a wrapper around 1ESPT in that way..you have to create your own `extends` template-compatible structure to do so which is pretty heavyweight.

There are two things we can do though, I think:
- We don't need to run the same tools..no real reason to do so.
- We might be able to template away some of the verbose bits, like the container variables.

mthalman: We do this in our Docker repos: https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/1es-unofficial.yml

mmitche: Oh interesting..nice. I like it.

mmitche: This doesn't really work. The issue is that we have the variable template that contains common variables like the container names, images, etc. The biggest benefit of abstracting into a template is that we get rid of that `containers:` in official and unofficial yaml.

However, the extends template (like https://github.com/dotnet/docker-tools/blob/main/eng/common/templates/task-prefix-decorator.yml) can't have a variables section, so you can't access the variables to be able to set up the containers.

eng/pipelines/official.yml
```diff
@@ -50,9 +52,6 @@ variables:
 - name: isScheduleTrigger
   value: ${{ eq(variables['Build.Reason'], 'Schedule') }}
 
-- name: isPRTrigger
```
mthalman, line 53: Isn't this still needed? Or is the intention to use `unofficial` for PR builds? Because I see `unofficial` doesn't have any triggers defined right now.

mmitche: There is a pr.yml for PR builds already. Unofficial is intended as a dev-version of the official build pipeline.

mthalman (approved):

mmitche:
Unrelated failure in aspnetcore.

PR #1760, opened 2025-08-05
Fix offline build for 2xx source-only

mthalman:
Offline source-only builds will fail in a non-1xx branch with the following error:

```
  /__w/1/s/src/sdk/src/Layout/redist/redist.csproj : error NU1102: Unable to find package Microsoft.NETCore.Platforms with version (= 10.0.0-preview.7.25377.103) [/__w/1/s/src/sdk/source-build.slnf]
```

The version that's referenced here is the "live" version of `Microsoft.NETCore.Platforms` that comes from the shared components (1xx build artifacts). The problem is that there's no package source mapping defined for `Microsoft.NETCore.Platforms` from the shared components source. There's only a mapping defined for it from SBRP.

This behavior is due to the logic in the `UpdateNuGetConfigPackageSourcesMappings` task. To understand what's happening we first need to look at the package versions that exist in from the shared components:

* 1.1.0
* 5.0.0
* 10.0.0-preview.7.25377.103

And the SBRP repo has these versions:

* 1.1.0
* 5.0.0

Since SBRP is being built in the non-1xx branch, the `UpdateNuGetConfigPackageSourcesMappings` task will classify those package versions as "current" and give them priority. This means that it will not create a mapping for any of the other sources if a package exists in the "current" source. Since one of the versions of `Microsoft.NETCore.Platforms` from the shared components matches the same version that exists in SBRP, it omits out the mapping of `Microsoft.NETCore.Platforms` for the shared component source (a mapping is done on just a package name basis, not version basis). So we miss out on having access to the version of `Microsoft.NETCore.Platforms` that was produced by the 1xx build.

The solution is to strip out the SBRP packages from the shared components feed. They are not needed because the non-1xx branch builds its own SBRP. This is done in a general manner by stripping out all bootstrapped repos which includes arcade and SBRP.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -19,6 +19,8 @@
     <ProjectReference Include="$(RepositoryEngineeringDir)init-detect-binaries.proj"
                       Condition="'$(BuildOS)' != 'windows' and '$(SkipDetectBinaries)' != 'true'"
                       BuildInParallel="true" />
+    <ProjectReference Include="$(RepositoryEngineeringDir)prebuild.proj"
+                      BuildInParallel="true" />
 
     <ProjectReference Include="$(RepoProjectsDir)dotnet.proj" />
 
--- a/eng/init-poison.proj
+++ b/eng/init-poison.proj
@@ -32,7 +32,7 @@
     <MSBuild Projects="$(MSBuildProjectFullPath)"
              Targets="GetFilteredSharedComponentPackages"
              Condition="Exists('$(SharedComponentsArtifactsPath)') and !$([System.String]::new(';$(BootstrapArcadeRepos);').Contains(';$(RepositoryName);'))"
-             Properties="IncludeOnlyToolingComponents=true">
+             Properties="SharedComponentFilterMode=ToolingOnly">
       <Output TaskParameter="TargetOutputs" ItemName="_FilteredSharedComponentPackages" />
     </MSBuild>
 
--- a/eng/prebuild.proj
+++ b/eng/prebuild.proj
@@ -1,0 +1,47 @@
+<Project Sdk="Microsoft.Build.NoTargets">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(TasksDir)Microsoft.DotNet.UnifiedBuild.Tasks\Microsoft.DotNet.UnifiedBuild.Tasks.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <SharedComponentsPrereqsArchivePathItem Include="$(ExternalTarballsDir)$(SourceBuiltSharedComponentsTarballName).*$(ArchiveExtension)" />
+  </ItemGroup>
+  <PropertyGroup>
+    <SharedComponentsPrereqsArchiveFile>@(SharedComponentsPrereqsArchivePathItem)</SharedComponentsPrereqsArchiveFile>
+    <HasSharedComponents>false</HasSharedComponents>
+    <HasSharedComponents Condition="'$(SharedComponentsPrereqsArchiveFile)' != '' or '$(CustomSharedComponentsArtifactsPath)' != ''">true</HasSharedComponents>
+  </PropertyGroup>
+
+  <Target Name="UpdateSharedComponentsArtifacts"
+          BeforeTargets="Build"
+          Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(HasSharedComponents)' == 'true'"
+          Outputs="$(BaseIntermediateOutputPath)UpdateNuGetConfig.complete">
+
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
+             Targets="GetFilteredSharedComponentPackages"
+             Properties="SharedComponentFilterMode=BootstrapOnly">
+      <Output TaskParameter="TargetOutputs" ItemName="_BootstrapPackages" />
+    </MSBuild>
+
+    <ItemGroup>
+      <_BootstrapPackagesFilenames Include="@(_BootstrapPackages)">
+        <PackagePath>$(SharedComponentsArtifactsPath)$([System.IO.Path]::GetFileName('%(PipelineArtifactPath)'))</PackagePath>
+      </_BootstrapPackagesFilenames>
+      <_BootstrapPackagesToDelete Include="@(_BootstrapPackagesFilenames->'%(PackagePath)')" />
+    </ItemGroup>
+    <!-- Delete bootstrap packages that shouldn't be included. For example, the Microsoft.NETCore.Platforms package can
+        exist in the shared components from two different origins: SBRP and runtime. Since non-1xx branches have their
+        own SBRP repo, it can create indeterminism from the NuGet.config as to which package should be loaded since
+        source mappings of that package would need to exist for both the current (non-1xx version) of SBRP and
+        shared-components sources.
+     -->
+    <Delete Files="@(_BootstrapPackagesToDelete)" Condition="'@(_BootstrapPackagesToDelete)' != ''" />
+
+  </Target>
+
+</Project>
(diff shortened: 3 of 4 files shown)


MichaelSimons:
I think this could be solved with https://github.com/dotnet/source-build-reference-packages/issues/1317.  That has other benefits as well.  IMO, we should investigate fixing that in-lieu of this solution.

mthalman:
@MichaelSimons - Friendly ping to review this again after we determined that https://github.com/dotnet/source-build-reference-packages/issues/1317 does not solve this.

eng/prebuild.proj
```diff
@@ -0,0 +1,47 @@
...
+  <PropertyGroup>
+    <SharedComponentsPrereqsArchiveFile>@(SharedComponentsPrereqsArchivePathItem)</SharedComponentsPrereqsArchiveFile>
+    <HasSharedComponents>false</HasSharedComponents>
+    <HasSharedComponents Condition="'$(SharedComponentsPrereqsArchiveFile)' != '' or '$(CustomSharedComponentsArtifactsPath)' != ''">true</HasSharedComponents>
+  </PropertyGroup>
+
+  <Target Name="UpdateSharedComponentsArtifacts"
+          BeforeTargets="Build"
+          Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(HasSharedComponents)' == 'true'"
+          Outputs="$(BaseIntermediateOutputPath)UpdateNuGetConfig.complete">
```
MichaelSimons: Why is this called UpdateNuGetConfig.complete?
```suggestion
          Outputs="$(BaseIntermediateOutputPath)UpdateSharedComponentsArtifacts.complete">
```

MichaelSimons (approved):

eng/prebuild.proj
```diff
@@ -0,0 +1,47 @@
+<Project Sdk="Microsoft.Build.NoTargets">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(TasksDir)Microsoft.DotNet.UnifiedBuild.Tasks\Microsoft.DotNet.UnifiedBuild.Tasks.csproj" />
+  </ItemGroup>
```
ViktorHofer: Why is this P2P necessary? I don't see this project invoking any custom tasks.

build.proj
```diff
@@ -9,16 +9,23 @@
...
   <ItemGroup>
+    <!-- This needs to be executed first as subsequent projects may introduce binaries into the repo
+         which should not be considered as part of the binary scan. -->
+    <ProjectReference Include="$(RepositoryEngineeringDir)init-detect-binaries.proj"
+                      Condition="'$(BuildOS)' != 'windows' and '$(SkipDetectBinaries)' != 'true'" />
+    
+    <!-- This needs to be executed before poisoning because it can end up deleting unneeded package
+         files which we don't want to have poisoned; otherwise it would lead to errors when those files
+         are found to be missing. -->
+    <ProjectReference Include="$(RepositoryEngineeringDir)prebuild.proj" />
```
ViktorHofer: Shouldn't this ProjectReference have some basic condition, i.e. source-build?

eng/prebuild.proj
```diff
@@ -0,0 +1,47 @@
...
+  <Target Name="UpdateSharedComponentsArtifacts"
+          BeforeTargets="Build"
+          Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(HasSharedComponents)' == 'true'"
+          Outputs="$(BaseIntermediateOutputPath)UpdateSharedComponentsArtifacts.complete">
+
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
+             Targets="GetFilteredSharedComponentPackages"
+             Properties="SharedComponentFilterMode=BootstrapOnly">
+      <Output TaskParameter="TargetOutputs" ItemName="_BootstrapPackages" />
+    </MSBuild>
```
ViktorHofer: Why is this recursive project call necessary? Shouldn't a DependsOnTargets="GetFilteredSharedComponentPackages" and setting `SharedComponentFilterMode=BootstrapOnly` before that be sufficient?

build.proj
mthalman: It's meant to be a general purpose project file for either mode.

eng/prebuild.proj
mthalman: It doesn't invoke tasks directly but it does indirectly. It calls the `GetFilteredSharedComponentPackages` which is defined in `Directory.Build.targets` at the root directory which needs to call a custom task.

eng/prebuild.proj
mthalman: I'm not a fan of that pattern. That's like calling a parameter-less `void` method and setting class fields to pass state and get the return state rather than just passing a parameter to the method have it return the value directly.

build.proj
ViktorHofer: But isn't the target in it conditioned on source-build?

eng/prebuild.proj
ViktorHofer: These recursive project calls are problematic in msbuild for multiple reasons:
- They trigger another evaluation (costly)
- When a project is invoked with and without global properties and it doesn't take special care of making sure that output and intermediate output paths are different, it will most likely cause a race condition.
- Don't work with static graph build (which we want to onboard the VMR onto).

There are more but these are the ones top of my head. If not absolutely necessary, I would avoid such project calls.

build.proj
```diff
@@ -9,16 +9,23 @@
...
 
   <ItemGroup>
+    <!-- This needs to be executed first as subsequent projects may introduce binaries into the repo
+         which should not be considered as part of the binary scan. -->
+    <ProjectReference Include="$(RepositoryEngineeringDir)init-detect-binaries.proj"
+                      Condition="'$(BuildOS)' != 'windows' and '$(SkipDetectBinaries)' != 'true'" />
+    
+    <!-- This needs to be executed before poisoning because it can end up deleting unneeded package
+         files which we don't want to have poisoned; otherwise it would lead to errors when those files
+         are found to be missing. -->
```
ViktorHofer: Dependencies in msbuild are usually expressed via ProjectReferences. Consider adding a dependency on prebuild.proj in init-posion.proj so that the ordering is respected when building with static graph and when invoking init-poison.proj directly.

eng/prebuild.proj
mthalman: This recursive pattern is already being used elsewhere for this target: https://github.com/dotnet/dotnet/blob/5c3434d8b831745673c31e186906f463e1036301/repo-projects/Directory.Build.targets#L346-L350

Can you confirm that if I change this to not be recursive, that it won't cause issues with parallel repo builds because each repo project would have its own instance of the `ItemGroup` that would be set by this target, correct?

ViktorHofer: Parallel repo builds invoking the same target is only problematic when those write to disk and to the same path or lock a file. Patterns like that. I only see reads and item changes so there shouldn't be an issue.

build.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1760/commits/28d6b9714e0d016d2c4d08be96f25556095ff70e

build.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1760/commits/1558b855751c0eef5387c79de086f9b2ab3adc59

eng/prebuild.proj
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1760/commits/28277612d7b34ae56d0c0394b28122fc2e87d841

eng/init-poison.proj
```diff
@@ -32,7 +37,7 @@
     <MSBuild Projects="$(MSBuildProjectFullPath)"
              Targets="GetFilteredSharedComponentPackages"
              Condition="Exists('$(SharedComponentsArtifactsPath)') and !$([System.String]::new(';$(BootstrapArcadeRepos);').Contains(';$(RepositoryName);'))"
-             Properties="IncludeOnlyToolingComponents=true">
+             Properties="SharedComponentFilterMode=ToolingOnly">
```
ViktorHofer, line 40: Will you eventually replace this project call as well with a DependsOnTargets? Just wondering.

mthalman: That, and other usages, are going to be more involved because of the `Condition` that needs to be set. Since that's outside the scope of these changes, I created an issue for it: https://github.com/dotnet/dotnet/issues/1953

PR #1762, opened 2025-08-05
Check source-built packages for poison

ellahathaway:
Fixes https://github.com/dotnet/source-build/issues/5297

[Test pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2766280&view=results)

### DIFF
--- a/eng/VmrLayout.props
+++ b/eng/VmrLayout.props
@@ -67,8 +67,10 @@
   <PropertyGroup Condition="'$(EnablePoison)' == 'true'">
     <PrebuiltPoisonMarkerFile>.prebuilt.xml</PrebuiltPoisonMarkerFile>
     <PrebuiltPoisonReportDataFile>$(ArtifactsLogDir)poison-prebuilt-catalog.xml</PrebuiltPoisonReportDataFile>
-    <SourceBuiltPoisonMarkerFile>.source-built.xml</SourceBuiltPoisonMarkerFile>
-    <SourceBuiltPoisonReportDataFile>$(ArtifactsLogDir)poison-source-built-catalog.xml</SourceBuiltPoisonReportDataFile>
+    <PreviouslySourceBuiltPoisonMarkerFile>.previously-source-built.xml</PreviouslySourceBuiltPoisonMarkerFile>
+    <PreviouslySourceBuiltPoisonReportDataFile>$(ArtifactsLogDir)poison-previously-source-built-catalog.xml</PreviouslySourceBuiltPoisonReportDataFile>
+    <SharedComponentPoisonMarkerFile>.shared-component.xml</SharedComponentPoisonMarkerFile>
+    <SharedComponentPoisonReportDataFile>$(ArtifactsLogDir)poison-shared-component-catalog.xml</SharedComponentPoisonReportDataFile>
     <PoisonedReportFile>$(ArtifactsLogDir)poisoned.txt</PoisonedReportFile>
     <PoisonUsageReportFile>$(ArtifactsLogDir)poison-usage.xml</PoisonUsageReportFile>
   </PropertyGroup>
--- a/eng/VmrLayout.props
+++ b/eng/VmrLayout.props
@@ -65,8 +65,10 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(EnablePoison)' == 'true'">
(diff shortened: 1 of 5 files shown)


eng/VmrLayout.props
```diff
@@ -65,8 +65,10 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(EnablePoison)' == 'true'">
-    <PoisonMarkerFile>.prebuilt.xml</PoisonMarkerFile>
-    <PoisonReportDataFile>$(ArtifactsLogDir)poison-catalog.xml</PoisonReportDataFile>
+    <PrebuiltPoisonMarkerFile>.prebuilt.xml</PrebuiltPoisonMarkerFile>
+    <PrebuiltPoisonReportDataFile>$(ArtifactsLogDir)poison-prebuilt-catalog.xml</PrebuiltPoisonReportDataFile>
+    <SourceBuiltPoisonMarkerFile>.source-built.xml</SourceBuiltPoisonMarkerFile>
+    <SourceBuiltPoisonReportDataFile>$(ArtifactsLogDir)poison-source-built-catalog.xml</SourceBuiltPoisonReportDataFile>
```
mthalman: Is there value in having separately named marker and catalog files? Or is it fine to just have the marker file name across all file types and one catalog for everything? cc @MichaelSimons

eng/init-poison.proj
```diff
@@ -48,9 +44,7 @@
 
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Poisoning existing packages for leak detection." />
 
-    <!-- There's an issue where the outputs of these tasks are not being fed into the poison check
-         See https://github.com/dotnet/source-build/issues/5297 -->
-    <MarkAndCatalogPackages PackagesToMark="@(PrebuiltPackages)" CatalogOutputFilePath="$(PoisonReportDataFile)" MarkerFileName="$(PoisonMarkerFile)" />
+    <MarkAndCatalogPackages PackagesToMark="@(PrebuiltPackages)" CatalogOutputFilePath="$(PrebuiltPoisonReportDataFile)" MarkerFileName="$(PrebuiltPoisonMarkerFile)" />
     <MarkAndCatalogPackages PackagesToMark="@(PreviouslySourceBuiltPackages);@(SharedComponentPackagesToPoison)" CatalogOutputFilePath="$(SourceBuiltPoisonReportDataFile)" MarkerFileName="$(SourceBuiltPoisonMarkerFile)" />
```
mthalman: I think `SharedComponentPackagesToPoison` should be split out into a separate call. That then is impacted by my other question of whether we can share marker/catalogs across all these file types.

eng/VmrLayout.props
ellahathaway: I'm don't about the importance of keeping separate catalogs, but I do know that the name of the marker file is irrelevant. Meaning that [we currently don't note the name of the file when we encounter one](https://github.com/dotnet/dotnet/pull/1762/files#diff-f5501759e23c7c042aeb7d720e55ec4c5d4601fb48af30820b98c61448b1677aL425-R428).

ellahathaway: In the event of a nupkg leak, it could be helpful to report the name of the marker file in the poison report so that we have more information about where the leak might've come from.

MichaelSimons:
@ellahathaway, what is your assessment on the new leaks being reported?  At least one looks like it could be false positive because it is a text file that has a matching hash.  Will you be updating the baseline or logging a follow-up issue to address these?

eng/VmrLayout.props
mthalman: That makes sense. We might as well differentiate the leak source as much as we can.

ellahathaway:
> @ellahathaway, what is your assessment on the new leaks being reported? At least one looks like it could be false positive because it is a text file that has a matching hash. Will you be updating the baseline or logging a follow-up issue to address these?

There's a number of leaks, so I am planning on updating the baseline in this PR to not block builds and logging a follow-up issue to address them.

eng/init-poison.proj
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1762/commits/c4f2659eb529b9d13f9f674ebdf385c652d1fd4b

mthalman (approved):

PR #1784, opened 2025-08-07
NuGet-Client should PublicSign on Windows

ViktorHofer:
Remove the unused `MS_PFX_PATH` and `NUGET_PFX_PATH` env vars and the checked-in NuGet.Client.sln file. NuGet already has that key in their source tree.

Switch from DelaySign to PublicSign on Windows. Unblocks insertion PRs from dotnet/dotnet to repos that use the live NuGet packages, i.e. https://github.com/dotnet/sdk/pull/50073

---

Documentation about Delay-signing: https://learn.microsoft.com/en-us/dotnet/standard/assembly/delay-sign

Documentation about Public-signing: https://github.com/dotnet/runtime/blob/main/docs/project/public-signing.md


nkolev92:
I think we'll probably need to tweak this change before it flows in NuGet.Client correct? 

The public sign stuff will break nuget.exe the same way it did pack. 

For pack, we can try to get rid of the ilmerging, but for nuget.exe we can't.

ViktorHofer:
I don't think so. The ilmerge invocation doesn't respect this setting and explicitly passes delays delaysign in.

IOW nothing broke due to this.

nkolev92:
Wasn't there a signing problem that led to an issue on the .NET SDK? 

It was a long conversation, not sure what we ended up needing this for.

ViktorHofer:
> Wasn't there a signing problem that led to an issue on the .NET SDK?

Yes the signing issue was the following: When building nuget-client inside the VMR without real signing, the binaries (from a normal VMR official build), the assemblies were marked as delay signed but never got the real signing information embedded.

.NET Framework relies on that information though in many different cases, so we got a FileNotFoundException when loading the nuget binaries on .NET Framework.

The .NET ecosystem changed from delay sigining to PublicSign by default years ago which doesn't have that pitfall. PublicSign assemblies can be freely loaded (with a few caveats but it mostly works). NuGet-Client wasn't using PublicSign but that never caused an issue cause every build that you promote is real signed.

So for the VMR non-real signed binaries to work on .NET Framework, we had to make this change to default to PublicSign.

PR #1796, opened 2025-08-08
Stop packing ilmerged pack

nkolev92:
An attempt to unblock https://github.com/dotnet/sdk/pull/50073. 

For simplicity, I'm just fixing up pack but not stopping the ilmerge. I'll do that in a follow-up PR on the NuGet side that has tests. 

This is just the faster way to flow the changes.


nkolev92:
In theory https://github.com/NuGet/NuGet.Client/pull/6697 should replace this.

ViktorHofer (approved):

PR #1811, opened 2025-08-10
Sign source-built artifacts

ellahathaway:
Closes https://github.com/dotnet/release/issues/1417

This PR includes the following changes:
- prep script now allows downloading multiple SDKs to dotnet/.dotnet. MicroBuild requires .NET 8, and it must be installed in the sources directory (dotnet/.dotnet)
- VMR signing verification now allows validating files without parsing a manifest first. This is needed because we only sign two source-build artifacts and do not need to parse the entire manifest for the artifacts to validate.
- Implemented a custom MSBuild task for signing SB artifacts
- Source-build artifacts can now be signed and validated when the `allowSigning` parameter is true.


eng/pipelines/templates/stages/source-build-stages.yml
```diff
@@ -67,6 +67,7 @@ stages:
           useMonoRuntime: false              # 🚫
           useSystemLibraries: false          # 🚫
           withPreviousSDK: false             # 🚫
+          sign: true                         # ✅
```
MichaelSimons: Shouldn't this only set to true if we are doing a signed VMR build?  It feels like this template should take in the desiredSigning parameter just like the vmr-build template.

ellahathaway:
Signing passes but signing validation via SignCheck fails. I looked through the signing binlogs, and based on SignTool's post-signing verification & the output from MicroBuild, it appears that things are being signed correctly.

I pulled the SB artifact, ran SignCheck locally, and confirmed that nothing is signed. I also peeked into some of the NuGet packages and assembles and verified the lack of signatures.

Not exactly sure what is causing this, but I'm investigating the following topics to see whether they are affecting the signing of these artifacts:
1. Shipping status
2. Strong naming
3. Repacking of the tars

ellahathaway:
> Not exactly sure what is causing this

The issue was that I needed to sign the artifacts in `$(Build.ArtifactStagingDirectory)`, not `$(Build.SourcesDirectory)/artifacts`.

ellahathaway:
Working through another issue. All artifacts are signed, but not all artifacts are strong named. The issue is related caused by some artifacts using the aspnetcore strong name, which is not imported by arcade's `Sign.props`: https://github.com/dotnet/arcade/blob/86b53945e6b6b239d68fa465e62fcf4323ff3b7b/src/Microsoft.DotNet.Arcade.Sdk/tools/Sign.props#L66-L71.

ellahathaway:
Waiting on a new signed build, then this is ready to be published :)

ellahathaway:
Blocked on https://github.com/dotnet/dotnet/issues/2022

ellahathaway:
[Signed build - stage 1 only](https://dev.azure.com/dnceng/internal/_build/results?buildId=2778005)

ellahathaway:
- Signed build (stage 1 only) passed minus a SignCheck verification failure (will be resolved by https://github.com/dotnet/arcade/pull/16075).
- Signed build (stage 1 signed, stage 2 consuming signed artifacts) is running now: https://dev.azure.com/dnceng/internal/_build/results?buildId=2778111.

Going to publish this PR.

eng/allowed-sb-binaries.txt
```diff
@@ -9,6 +9,7 @@ artifacts/
 .git/
 .packages/
 prereqs/packages/
+MicroBuild/
```
MichaelSimons: I think a note should be added to call out these are not checked in rather something introduced during CI.  It is possible to install this outside the source tree?

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -657,11 +664,26 @@ jobs:
...
+
+          MSBuild -restore eng/sign-source-built-artifacts.proj \
+            /p:SignProperties=$sign_properties \
+            /p:ArtifactsStagingDir=$(Build.ArtifactStagingDirectory) \
+            /p:OfficialBuildId=$(Build.BuildNumber) \
+            /bl:artifacts/log/outer-sign-source-built-artifacts.binlog
+        displayName: Sign Source-Built Artifacts
+        workingDirectory: $(Build.SourcesDirectory)
+
+    - ${{ if and(eq(parameters.sign, 'True'), eq(variables['System.TeamProject'], 'internal')) }}:
```
MichaelSimons: It doesn't seem like `eq(parameters.sign, 'True')` should be necessary to check again.

eng/sign-source-built-artifacts.proj
```diff
@@ -0,0 +1,51 @@
...
+      <SourceBuiltAssetsDir>$(ArtifactsPublishStagingDir)/assets/$(Configuration)</SourceBuiltAssetsDir>
+    </PropertyGroup>
+
+    <PropertyGroup>
+      <SourceBuiltSdkTarballName>dotnet-sdk</SourceBuiltSdkTarballName>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <!-- SDK tarball - prefer root if present; otherwise pick from Sdk/ subtree -->
+      <SignableSourceBuiltAsset Include="$(SourceBuiltAssetsDir)/$(SourceBuiltSdkTarballName)-*$(ArchiveExtension)" />
```
MichaelSimons: Can some validation logic be added to fail the build if the sdk and artifacts tarball are not found?  What is the scenario when the root sdk would not exist?

eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.Sign/Microsoft.DotNet.SourceBuild.Tasks.Sign.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
MichaelSimons: What is the reasoning behind creating a new sign project versus including this in the UB.Tasks project?

prep-source-build.sh
```diff
@@ -293,6 +252,53 @@ function BootstrapArtifacts {
...
+  echo "  $prebuiltsTarballPattern exists in $packagesArchiveDir...it will not be downloaded"
+  downloadPrebuilts=false
+fi
+
+# Check if dotnet is installed
+if [ "$installDotnet" == true ] && [ -d "$REPO_ROOT/.dotnet" ]; then
+  installedVersions=$("$REPO_ROOT/.dotnet/dotnet" --list-sdks | awk '{print $1}')
+  expectedVersion=$(GetXmlPropertyValue "PrivateSourceBuiltSdkVersion" "$REPO_ROOT/eng/Versions.props")
+  if ! grep -qx "$expectedVersion" <<< "$installedVersions"; then
+    echo "  Installing .NET SDK $expectedVersion...to skip this installation, exit and re-run with --no-sdk"
```
MichaelSimons: This message feels weird to me.  The chances of someone seeing and exiting right away before the sdk installation starts is unlikely.  IMO this section should only emit a message if the sdk installation is being skipped.  The install logic below already indicated dotnet is being installed.  If we want to emit the version it should be done there.

prep-source-build.sh
```diff
@@ -293,6 +252,53 @@ function BootstrapArtifacts {
...
+  else
+    echo "  ./.dotnet SDK directory exists for SDK version $expectedVersion...it will not be installed"
```
MichaelSimons:
```suggestion
    echo "  Skipping SDK installation - version $expectedVersion detected"
```

eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.Sign/SignArtifacts.cs
```diff
@@ -0,0 +1,178 @@
...
+            File.Delete(SigningPropsTargetPath);
+        }
+    }
+
+    /// <summary>
+    /// Runs the signing process
+    /// </summary>
+    private void RunSigning()
+    {
+        using (var process = new Process())
```
MichaelSimons: If moved into the UB tasks, you could utilize the ProcessService type being added in https://github.com/dotnet/dotnet/pull/2110

eng/allowed-sb-binaries.txt
ellahathaway: > It is possible to install this outside the source tree?

https://github.com/dotnet/arcade/issues/15731 needs to be addressed first. This issue has come up a few times now, so it's probably worth addressing at this point. I'll start working on it.

eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.Sign/SignArtifacts.cs
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1811/commits/0bf05ea0d8197c3a9d62a93146c47883fcb1bcb3

eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.Sign/Microsoft.DotNet.SourceBuild.Tasks.Sign.csproj
ellahathaway: My logic was that leak detection is in its own project, but you're right, it's better to add this signing task into UB.Tasks project. Addressed in https://github.com/dotnet/dotnet/pull/1811/commits/8dc7759e4640155d08217742dc4e496b53a85e48

eng/pipelines/templates/jobs/vmr-build.yml
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1811/commits/bdbc29c4d82e4d271f4de005f722c6278ddae9f8

eng/allowed-sb-binaries.txt
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1811/commits/2fae31295b448750c6e8d897d9bfe20c54985982

eng/sign-source-built-artifacts.proj
ellahathaway: - The root sdk should always exist I fixed this in https://github.com/dotnet/dotnet/pull/1811/commits/cb801785b9f903c48ecc967bf3fd84200acc1940
- Added validation logic to fail the build if each artifact is not found: https://github.com/dotnet/dotnet/pull/1811/commits/e8ac6f69e5d2384809d6f02b4035d4f82595d72c

prep-source-build.sh
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1811/commits/4136df35dbbaf78cfcf0fec65f410d873aec1800

prep-source-build.sh
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1811/commits/48e64ba7478c5a9d64e134d17cb84bd045938018

ellahathaway:
Waiting on https://github.com/dotnet/dotnet/pull/2232 to be merged, then this will be ready for re-review.

eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.Sign/Microsoft.DotNet.SourceBuild.Tasks.Sign.csproj
MichaelSimons: Not sure the history behind the poisoning infra, it could be merged into the tasks project IMO.

eng/pipelines/templates/stages/source-build-stages.yml
```diff
@@ -73,6 +80,7 @@ stages:
           useMonoRuntime: false              # 🚫
           useSystemLibraries: false          # 🚫
           withPreviousSDK: false             # 🚫
+          allowSigning: true                 # ✅
```
MichaelSimons: `allowSigning` is not self-describing to me.  It raises questions of what does this mean?  After thinking about this, I think negating this parameter would be more helpful - `disableSigning`.  The other legs are overriding the desiredSigning and are never signed.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/static/AspNetCore.snk
MichaelSimons, line 1: Is it possible to utilize the copy checked into arcade?

ellahathaway: This would require updating the Arcade SDK to use the checked-in copy and then waiting until a rebootstrap with those changes. Given that we're in RC1, this process could take a while. That said, it's probably still worth doing in the long run. I will file an issue in arcade and add a comment in the SB signing task linking to that issue.

MichaelSimons: For now, can't you reference the arcade copy or would that not be acceptable?

ellahathaway: Oh, I see. That should work. Will update this

eng/pipelines/templates/stages/source-build-stages.yml
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1811/commits/c4abbcaa71299c52be34360ef13e2b8a0b3e423a

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/static/AspNetCore.snk
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/1811/commits/61d930f14180691c3ee32f6357fb52ff92406f16

ellahathaway:
Needs https://github.com/dotnet/dotnet/pull/2291 to be merged first. Then these changes can be backported to release/10.0.1xx

PR #1827, opened 2025-08-11
Exclude source build artifacts under source-build directory

Winniexu01:
Fixes: https://github.com/dotnet/source-build/issues/5294

Exclude the source build artifacts under source-build directory.


Winniexu01:
Pipeline run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2770021 (internal Microsoft link)

eng/Publishing.props
```diff
@@ -116,7 +116,7 @@
       <Artifact Include="@(SourceBuildArtifact)">
         <IsShipping>true</IsShipping>
         <Kind>Blob</Kind>
-        <RelativeBlobPath>source-build/%(Filename)%(Extension)</RelativeBlobPath>
+        <RelativeBlobPath>%(Filename)%(Extension)</RelativeBlobPath>
```
mthalman: This impacts where the tarball gets published in blob storage. It would change from `https://ci.dot.net/public/source-build/Private.SourceBuilt.Artifacts.<version>.<rid>.tar.gz` to `https://ci.dot.net/public/Private.SourceBuilt.Artifacts.<version>.<rid>.tar.gz`. That's not what we want. It needs to be in the `source-build` location.

Instead, we should change the location of where the tarball at `Release/Private.SourceBuilt.Artifacts.<version>.<rid>.tar.gz` gets written so that it instead gets written at `Release/source-build/Private.SourceBuilt.Artifacts.<version>.<rid>.tar.gz`. Then there will be only one copy of the file.

Winniexu01: Hi @mthalman My title was wrong, but it has been corrected. According to your previous reply, you want to exclude the tarball in the `Release/source-build` path. So I want to confirm with you again, now the `Release/source-build` path is expected to be retained, right?

mthalman: Yes, previously I thought that we could get rid of the file from the `source-build` directory but  as my previous comment states, it makes things simpler if we preserve it and get rid of the other one since it is needed for the actual Arcade publishing logic.

Winniexu01:
Pipeline run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2776926&view=results

eng/PublishSourceBuild.props
```diff
@@ -81,7 +81,7 @@
       <SourceBuiltLayoutDir>$([MSBuild]::NormalizeDirectory('$(BaseIntermediateOutputPath)', 'artifacts-layout'))</SourceBuiltLayoutDir>
 
       <!-- Outputs -->
-      <SourceBuiltTarballName>$(SourceBuiltAssetsDir)/$(SourceBuiltArtifactsTarballName).$(SourceBuiltSdkNonStableVersion).$(TargetRid)$(ArchiveExtension)</SourceBuiltTarballName>
+      <SourceBuiltTarballName>$(SourceBuiltAssetsDir)/source-build/$(SourceBuiltArtifactsTarballName).$(SourceBuiltSdkNonStableVersion).$(TargetRid)$(ArchiveExtension)</SourceBuiltTarballName>
```
mthalman: I suspect that this extra `source-build` dir may not be needed since it's not part of the artifacts dir that ends up getting published.

eng/PublishSourceBuild.props
```diff
@@ -149,7 +149,7 @@
 
     <!-- Tests expect this file to be in ArtifactsAssetsDir. -->
     <Copy SourceFiles="$(SourceBuiltTarballName)"
-          DestinationFolder="$(ArtifactsAssetsDir)"
+          DestinationFolder="$(ArtifactsAssetsDir)source-build"
```
mthalman: A property should be defined for this path so it can be shared both here and in the `Publishing.props` file.

eng/PublishSourceBuild.props
Winniexu01: In fact, the files under the `$(SourceBuiltAssetsDir)/source-build/` path will be published to the Azure artifact: https://dev.azure.com/dnceng/internal/_build/results?buildId=2776377

eng/PublishSourceBuild.props
Winniexu01: Update: https://github.com/dotnet/dotnet/pull/1827/commits/36fdcff8a83adc3d875d2357d270639638a2dc56

eng/PublishSourceBuild.props
mthalman: I think they're only be published to Azure artifacts because these files exist in `ArtifactsAssetSourceBuildDir`, not because they exist in `$(SourceBuiltAssetsDir)/source-build`.

Winniexu01: `$(ArtifactsAssetDir)`: `$(RepoRoot)/artifacts/assets/$(Configuration)`
`$(SourceBuiltAssetsDir)`: `$(Build.ArtifactStagingDirectory)/artifacts/assets/$(Configuration)`
The content under the `$(SourceBuiltAssetsDir)` path will be published to Azure Artifacts: https://github.com/dotnet/dotnet/blob/15307b3e362d30e1e6cfac69becb28160534db63/eng/pipelines/templates/jobs/vmr-build.yml#L336-L341

When executing the `PushToBuildStorage` task, a backup will be left under `$(Build.ArtifactStagingDirectory)/artifacts`:
https://github.com/dotnet/dotnet/blob/15307b3e362d30e1e6cfac69becb28160534db63/src/arcade/src/Microsoft.DotNet.Build.Tasks.Feed/src/PushToBuildStorage.cs#L426

So the extra `source-build` dir is necessary, otherwise, we need to change the target publish path.

Winniexu01: And there is another similar issue, three copies of`dotnet-sdk-*` files are published:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2779775&view=artifacts&pathAsName=false&type=publishedArtifacts (internal Microsoft link)
image


Caused by:
https://github.com/dotnet/dotnet/blob/main/eng/Publishing.props#L87-L111
https://github.com/dotnet/dotnet/blob/main/eng/Publishing.props#L116

These three files are exactly the same, and the original source is at `artifacts\assets\<config>\Sdk\<SdkVersion>\dotnet-sdk-<SdkVersion>-<TargetRid><ArchiveExtension>`. How many copies of these files should be kept? Do we need to open a separate issue to track and document this?

mthalman: Ok, the inclusion of the SDK made me realize we need to consider this to be a breaking change for downstream developers. I think we should avoid that. So I'm going to reverse my position on these changes yet again. We should investigate to see if it's possible to ensure that the artifacts get published to the `source-build` directory in blob storage, but not be left with the `source-build` directory in the pipeline artifacts.

mthalman (approved):

PR #1874, opened 2025-08-13
Detect Git worktrees in build.sh

Winniexu01:
Fixes: https://github.com/dotnet/source-build/issues/5304

Previously, `build.sh` only checked for a .git directory to verify a valid Git repository. In a worktree scenario, .git is a file containing a "gitdir: " reference to the actual git directory.  Add a logic to check if .git is a file and whether it contains a "gitdir:" line (case insensitive).


build.sh
```diff
@@ -395,39 +395,51 @@ if [[ "$sourceOnly" == "true" ]]; then
     }
 
     GIT_DIR="$scriptroot/.git"
-    if [ -f "$GIT_DIR/index" ]; then # We check for index because if outside of git, we create config and HEAD manually
-      if [ -n "$sourceRepository" ] || [ -n "$sourceVersion" ] || [ -n "$releaseManifest" ]; then
-        echo "ERROR: Source Link arguments cannot be used in a git repository"
+    if [ -f "$GIT_DIR" ]; then
+      if ! grep -i '^gitdir: ' "$GIT_DIR"; then
```
mthalman: This will cause `grep` to output the matching line to stdout. That should be suppressed.

build.sh
```diff
@@ -395,39 +395,51 @@ if [[ "$sourceOnly" == "true" ]]; then
...
+        echo "ERROR: $scriptroot is not a git worktree."
```
mthalman: We do not want to return an error if either `-release-manifest` or `--source-repository`/`--source-version` are provided.

build.sh
Winniexu01: Update: https://github.com/dotnet/dotnet/pull/1874/commits/6ab3a9db98735e220e4fd4cc215e78340b83974c, add `-q` to suppresses all output

build.sh
Winniexu01: In fact, if `-release-manifest` or `--source-repository`/`--source-version` are provided:
https://github.com/dotnet/dotnet/blob/a6b8690db968f9c8e7f568bca957b2fd606946d5/eng/pipelines/templates/jobs/vmr-build.yml#L560-L563
`$GIT_DIR` will be automatically deleted in the code: https://github.com/dotnet/dotnet/blob/a6b8690db968f9c8e7f568bca957b2fd606946d5/eng/pipelines/templates/jobs/vmr-build.yml#L376-L382

So, it will not return an error if either `-release-manifest` or `--source-repository`/`--source-version` are provided.

For the sake of rigorous logic, I deleted the error and added an additional condition - `fake_git` to check when `.git` is neither a file nor a directory. In this case, it will obtain `-release-manifest` or `--source-repository`/`--source-version` to create a "fake" .git directory.

build.sh
```diff
@@ -395,39 +395,51 @@ if [[ "$sourceOnly" == "true" ]]; then
...
+        echo '[remote "origin"]' > "$GIT_DIR/config"
+        echo "url=\"$sourceRepository\"" >> "$GIT_DIR/config"
+        echo "$sourceVersion" > "$GIT_DIR/HEAD"
 
-        if [ -z "$sourceRepository" ] || [ -z "$sourceVersion" ]; then
-          echo "ERROR: sourceRepository and sourceVersion must be specified in $releaseManifest"
+        if [ ! -d "$scriptroot/.git" ]; then
+          echo "ERROR: $scriptroot is not a git repository."
           exit 1
         fi
```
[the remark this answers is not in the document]
Winniexu01: Regardless of whether .git is a file or a folder, if it is deleted, then `[ -f "$GIT_DIR" ]` will also be false, so `if [ ! -d "$scriptroot/.git" ]` will be executed to check whether the fake .git folder is created successfully.

build.sh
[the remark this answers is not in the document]
Winniexu01: Update: https://github.com/dotnet/dotnet/commit/6ab3a9db98735e220e4fd4cc215e78340b83974c, add -q to suppresses all output

build.sh
mthalman: Note that this is only true when running from the pipeline. The same conditions wouldn't necessarily be true when running the script outside of the pipeline.

build.sh
```diff
@@ -395,51 +395,60 @@ if [[ "$sourceOnly" == "true" ]]; then
     }
 
     GIT_DIR="$scriptroot/.git"
+    fake_git=false
+    # Check if exist .git directory or file
     if [ -f "$GIT_DIR" ]; then
-      if ! grep -i '^gitdir: ' "$GIT_DIR"; then
-        echo "ERROR: $scriptroot is not a git worktree."
+      if ! grep -iq '^gitdir: ' "$GIT_DIR"; then
+        rm -rf "$GIT_DIR"
```
mthalman: We shouldn't be deleting files from the user's disk. In this case, an error is warranted but only if `--source-repository`/`--source-version` aren't provided.

Winniexu01: Update:
If the `.git file` is corrupted or the `index` file does not exist under the `.git directory`, `$fake_git` will be set to true. In this case, if neither `--release-manifest` nor `--source-repository / --source-version` are not provided, the script will exit with an error:
https://github.com/dotnet/dotnet/blob/3d567ce87de56fbc32cc8b2f40b73ad8b1f3475c/build.sh#L419-L424

If either `--release-manifest` or `--source-repository / --source-version` are provided, the script will attempt to create the .git directory. At this point, if a broken `.git` file exists, the command `mkdir -p "$GIT_DIR"` will fail with the error: `mkdir: cannot create directory '.git': File exists`. Now, instead of removing the file from the user's disk, I rename the `.git` file to `.git.old`:
https://github.com/dotnet/dotnet/blob/3d567ce87de56fbc32cc8b2f40b73ad8b1f3475c/build.sh#L401-L403

mthalman: We don't want to modify it in any way. This may be a user's custom fork of the repo and they may have their own file named `.git` for whatever reason. If it doesn't align with our requirements for the build, a proper error should be returned. But we shouldn't just be modifying/deleting/renaming the file.

Winniexu01: Update:
Revert changes related to creating the `.git directory` for a broken .git file and modifying/deleting/renaming a broken .git file. The current code now only checks whether the `.git file` meets our build requirements. If it does not, the script will check if `--release-manifest` or `--source-repository / --source-version` are provided. If neither is specified, it returns an error; otherwise, it does not output any message.

Winniexu01: Update: https://github.com/dotnet/dotnet/pull/1874/commits/bb464575019d38dd2a143609df822a86605e4923

Check whether the `.git` file conforms to the format requirements and verify the `git_dir path`. If the path does not exist and either `-release-main` or `-source-repository/-source-version` are provided, the corresponding path will be created; otherwise, an error message will be returned.

build.sh
```diff
@@ -545,4 +560,4 @@ if [[ "$sourceOnly" == "true" ]]; then
   echo "Found bootstrap versions: SDK $SDK_VERSION, Arcade $ARCADE_BOOTSTRAP_VERSION, NoTargets $NOTARGETS_BOOTSTRAP_VERSION and Traversal $TRAVERSAL_BOOTSTRAP_VERSION"
 fi
 
-Build
+Build
```
mthalman:
```suggestion
Build

```

build.sh
```diff
@@ -395,39 +395,54 @@ if [[ "$sourceOnly" == "true" ]]; then
     }
 
     GIT_DIR="$scriptroot/.git"
-    if [ -f "$GIT_DIR/index" ]; then # We check for index because if outside of git, we create config and HEAD manually
-      if [ -n "$sourceRepository" ] || [ -n "$sourceVersion" ] || [ -n "$releaseManifest" ]; then
-        echo "ERROR: Source Link arguments cannot be used in a git repository"
+    fake_git=false
```
mthalman: I'm confused about what this variable means. Based on its name I would expect that this means the repo is a fake, meaning "not functional", git repo. But that doesn't seem to be the case since it gets set to true when a valid `.git` file is found or a `.git` directory exists. Can you clarify the variable name? And the use of `_` in the name should be avoided since it doesn't following the camelCasing pattern used by the reset of the script.

build.sh
```diff
@@ -395,39 +395,54 @@ if [[ "$sourceOnly" == "true" ]]; then
...
-          || (echo "ERROR: Failed to find sourceRepository in $releaseManifest" && exit 1)
-        sourceVersion=$(get_property "$releaseManifest" sourceVersion) \
-          || (echo "ERROR: Failed to find sourceVersion in $releaseManifest" && exit 1)
-
-        if [ -z "$sourceRepository" ] || [ -z "$sourceVersion" ]; then
-          echo "ERROR: sourceRepository and sourceVersion must be specified in $releaseManifest"
-          exit 1
-        fi
+        # We need to add "fake" .git/ files when not building from a git repository
+        mkdir -p "$GIT_DIR"
```
mthalman: It doesn't look like this code will ever be reached if the repo has no `.git` file or directory.

build.sh
Winniexu01: After investigation, I found that checking only the `.git` file format is not enough. If the path specified in `.git` does not exist, the build will fail due to missing `RepositoryCommit`: 
https://dev.azure.com/dnceng/internal/_build/results?buildId=2777626&view=logs&j=e7ba9ff4-fe26-5beb-e36c-f5fba7654fac&t=66022860-aeeb-5938-cd86-e3af635d8902 (internal Microsoft link)

Therefore, it’s necessary to ensure both the format and existence of the path, especially the presence of `config` and `HEAD` files.

Would it be more reasonable to check `.git/config` and `.git/HEAD` directly instead of `.git/index`? This would also eliminate the need for the `fake_git` variable.

build.sh
Winniexu01: No, if the `.git` file or path does not exist, this code will still be executed as long as `--release-manifest` or `--source-repository / --source-version` are provided.

build.sh
```diff
@@ -395,7 +395,16 @@ if [[ "$sourceOnly" == "true" ]]; then
...
 
     GIT_DIR="$scriptroot/.git"
-    if [ -f "$GIT_DIR/index" ]; then # We check for index because if outside of git, we create config and HEAD manually
+    # Check if exist .git file
+    if [ -f "$GIT_DIR" ]; then
+      if ! grep -iq '^gitdir: ' "$GIT_DIR"; then
+        echo "ERROR: $GIT_DIR exist, but does not contain a valid 'gitdir:' pointer."
+        exit 1
+      else
+        GIT_DIR=$(grep -oP '^gitdir: \K.*' "$GIT_DIR")
```
mthalman: Perl regex feature is not supported on the BusyBox version of grep which is what is used by Alpine Linux. This script needs to support Alpine.

Winniexu01: Update: https://github.com/dotnet/dotnet/pull/1874/commits/8c0e4c1a3356df896a1af403d547ba4906f8926b

mthalman (approved):

PR #1877, opened 2025-08-13
Update source build poison report to deterministic and decoupled from the environment it was created in

MichaelSimons:
Fixes https://github.com/dotnet/source-build/issues/5309
Fixes https://github.com/dotnet/source-build/issues/5308

These changes do the following:

1. Reverts https://github.com/dotnet/dotnet/pull/1876
2. Replaces PackageVersion values with x.y.z when comparing poison baselines
3. Orders the Matches in poison report
4. Makes the Match paths relative to the repo root
5. Removes all non-assembly files from the poison catalog - only assemblies and nupkgs will be included.


eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.LeakDetection/CheckForPoison.cs
```diff
@@ -266,13 +199,13 @@ private static PoisonedFileEntry CheckSingleFile(IEnumerable<CatalogPackageEntry
             {
                 // This hash can match either the original hash (we couldn't poison the file, or redownloaded it) or
```
mthalman: The comment should be updated.

eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.LeakDetection/CheckForPoison.cs
```diff
@@ -388,7 +321,7 @@ private static PoisonedFileEntry ExtractAndCheckZipFileOnly(IEnumerable<CatalogP
             // first check for a matching poisoned or non-poisoned hash match:
```
mthalman: The comment should be updated

MichaelSimons: Fixed with https://github.com/dotnet/dotnet/pull/1877/commits/4510ec03ead05ea5c8ccb0767c6ab4185586ded3

eng/tools/tasks/Microsoft.DotNet.SourceBuild.Tasks.LeakDetection/CheckForPoison.cs
MichaelSimons: Fixed with https://github.com/dotnet/dotnet/pull/1877/commits/4510ec03ead05ea5c8ccb0767c6ab4185586ded3

PR #1879, opened 2025-08-13
Update binskim scan paths to hit artifact dirs

mmitche:
Update the binskim scan paths to a pattern similar to sdk's. Hits binaries properly, rather than relying on artifact upload.

### DIFF
--- a/eng/pipelines/official.yml
+++ b/eng/pipelines/official.yml
@@ -93,6 +93,7 @@ extends:
         tsaEnabled: true
       binskim:
         enabled: true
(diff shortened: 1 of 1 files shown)


mmitche:
Hits 80k files in Windows x64 build, similar in other arches

eng/pipelines/official.yml
```diff
@@ -93,6 +93,7 @@ extends:
         tsaEnabled: true
       binskim:
         enabled: true
+        analyzeTargetGlob: +:f|eng\**\*.props;+:f|**\artifacts\bin\**\*.dll;+:f|**\artifacts\bin\**\*.exe;-:f|**\artifacts\bin\**\msdia140.dll;-:f|**\artifacts\bin\**\pgort140.dll;-:f|**\artifacts\bin\*Tests\**;-:f|**\Microsoft.NET.Runtime.Emscripten**\tools\**;-:f|**\CodeCoverage\**;-:f|**\artifacts\bin\**\capstone.dll;
```
wtgodbe: Will we ever have to change this? If so, would be nice to add a comment explaining how to update it

mmitche: We may, to exclude various files we don't want scanned, at least until binskim gets proper artifact unpacking support

wtgodbe (approved):

PR #1897, opened 2025-08-14
Exclude runtime-dependent jobs on non-1xx branches

### LINKED ISSUE #1113: [feature bands] Exclude certain UB legs from pipeline for non-1xx branches
There are several build legs which are only relevant in the 1xx branch (where shared components are produced from). For other release branches, these branches should be excluded from running in the pipeline. Examples include short stack and cross-OS DAC builds that only build the runtime repo.

mthalman:
In non-1xx branches, the runtime repo never gets built. So the pipeline jobs that are dependent on building that repo need to be excluded from running. Specifically, those are the ShortStack and cross-OS DAC jobs.

Other jobs which are dependent on these jobs retrieve those assets via online package feed instead of from the current pipeline run's artifacts.

Closes #1113


eng/pipelines/templates/stages/vmr-verticals.yml
```diff
@@ -658,21 +675,22 @@ stages:
...
+            - Android_Shortstack_x64
+            - Android_Shortstack_x86
+            - iOS_Shortstack_arm64
+            - iOSSimulator_Shortstack_arm64
+            - iOSSimulator_Shortstack_x64
+            - tvOS_Shortstack_arm64
+            - tvOSSimulator_Shortstack_arm64
+            - tvOSSimulator_Shortstack_x64
+            - MacCatalyst_Shortstack_arm64
+            - MacCatalyst_Shortstack_x64
```
jkoritzinsky, line 693: To confirm: the sdk will build workload packs for the 2xx band with the 1xx packages from emsdk, runtime, etc? Just want to make sure that's the intended behavior.

mthalman: Correct

eng/pipelines/templates/stages/vmr-build.yml
```diff
@@ -100,6 +100,8 @@ stages:
     isBuiltFromVmr: ${{ parameters.isBuiltFromVmr }}
     scope: ${{ parameters.scope }}
     verifications: ${{ parameters.verifications }}
+    ${{ if and(ne(variables['Build.SourceBranch'], 'refs/head/main'), not(contains(variables['Build.SourceBranch'], '.1xx'))) }}:
```
MichaelSimons: Should we be concerned about preserving the ability to queue a run on a feature/dev branch?  e.g. run the full pipeline to validate changes?  Should a pipeline parameter be exposed to control this for those scenarios?

mthalman: Ah, good point. That should be a requirement.

mthalman: The only decent solution I can think of is to expose a pipeline parameter.

I'm thinking of a tri-state parameter like:

```yaml
- name: featureBand
  type: string
  default: detectByBranchName
  values:
  - detectByBranchName
  - 1xx
  - not1xx
```

I'm curious to get @mmitche's thoughts on this as well.

MichaelSimons: I was thinking a tri-state parameter as well.

Do you think it would help to add an ends with xx condition

`(not main) and (not ends with 1xx) and (ends with xx)`

It feels like this would get more feature/dev branches to work with the default (detectByBranchName) assuming the majority of these are based on 1xx

mthalman: > assuming the majority of these are based on 1xx

I feel like we should bias toward making it work by default for main and non-1xx branches since that will be where the majority of the development work will occur. But this approach doesn't accomplish that.

Perhaps we need a different approach. What if we based the conditions on the name of the pipeline?

- `dotnet-unified-build` -> main and 1xx
- `dotnet-unified-build-feature-bands` -> 2xx, 3xx, etc.

Then it wouldn't matter what your branch is named. You just need to use the right pipeline.

mthalman: Fixed in https://github.com/dotnet/dotnet/pull/1897/commits/1197962d11e98a5c94b15c09f7119894bf34e9ee

MichaelSimons (approved):

PR #1909, opened 2025-08-15
Remove whitespace from property name

am11:
Apparently, msbuild evaluates it with whitespaces as a different property.


am11:
cc @akoeplinger it was a typo and the effect was we were missing CG2 :(
image
Please let me know if this is the right place (so it can make in rc1).

akoeplinger (approved):

akoeplinger:
Iiiinteresting, didn't know this was possible :)

am11:
Lol, yes it's surprising especially when we are not allowed to _define_ such property `< Foo >`. At the very least, it should be an error 😅

akoeplinger:
weird, this should have been detected by https://github.com/dotnet/msbuild/pull/5672

am11:
Quick repro:

```sh
$ cat > foo.proj <<EOF
<Project>
  <PropertyGroup>
    <Foo>test</Foo>
  </PropertyGroup>

  <Target Name="Build">
    <Message Text="Foo: '\$(Foo)' and with whitespace: '\$( Foo )'" Importance="High" />
  </Target>
</Project>
EOF

$ dotnet build foo.proj -v:d
Restore complete (0,0s)
  foo succeeded (0,0s)
    Foo: 'test' and with whitespace: ''
```

shall upstream it.

am11:
https://github.com/dotnet/msbuild/issues/12352

Issue #1011, opened 2021-12-17
Support multiple feature bands


MichaelSimons:
**Scenario:**
* Multiple feature bands (e.g. 6.0.1xx and 6.0.2xx) are released on the same day.  
* Source-build today only supports building each release tarball separately which each produce a copy of the runtime (logically identical).
* Source-build tooling should support building and releasing both but with a shared runtime just like the Microsoft release.

ghost:
I couldn't figure out the best area label to add to this issue. If you have write-permissions please help me learn by adding exactly one area label.

tmds:
Every feature band involves some work.
If there is enough compatibility, having the latest would be preferable.

I guess one of the driving reasons for feature bands is to ensure compatibility for visual studio.
I think this is less important for source-built .NET.

cc @omajid

omajid:
> I guess one of the driving reasons for feature bands is to ensure compatibility for visual studio. I think this is less important for source-built .NET.

Agreed. I think we care more about *backwards compat* for source-build users. We would be okay with new features, but it's backwards compat that have a large impact.

Red Hat ships any feature band upgrades as monthly updates. That means customers will upgrade to new feature bands (without any choice) to get CVE fixes. Any backward incompatibility in the feature band has a risk of breaking users building .NET applications, and is almost too risky for the users/customers Red Hat cares about.

mmitche:
@mthalman Is there work left here?

mthalman:
Yes, the remaining 3 open sub-issues are still relevant.

PR #1910, opened 2025-08-15
Use SB artifacts RID

mthalman:
Contributes to https://github.com/dotnet/dotnet/issues/1011

To support non-1xx feature bands for source-only builds, the prep script downloads the SB artifacts that came were produced from the corresponding 1xx build. This is known as the shared components artifacts tarball.

However, the logic to determine the URL from which to download this tarball is incorrect. The filename contains the RID value. The script is using Arcade scripts to get the RID but that RID doesn't align with the RID defined in the tarball. The RID for the tarball is actually defined by the pipeline ([example](https://github.com/dotnet/dotnet/blob/a7b5a075945764f5e1775604300e3a8b7c12b24e/eng/pipelines/templates/stages/source-build-and-validate.yml#L90)). To fix this, we need to align with the RID being used by the pipeline.

The prep script already happens to have an `artifactsRid` option. But that applies to the downloading of the PSB artifacts, which today is only ever valid to use the `centos.<version>-x64` RID since that is the only artifacts tarball we publish from where this downloads from. So, without any other changes, setting `artifactsRid` wouldn't work in cases like `almalinux.8-x64`.

Instead, we can consolidate the manner in which we retrieve SB artifacts between PSB and shared components. Today these are referenced in different ways. PSB artifacts get published as part of the SB release process and referenced from `https://builds.dotnet.microsoft.com/source-built-artifacts/assets`. And shared components artifacts are published via Arcade from VMR builds and are referenced from `https://ci.dot.net/public/source-build`.

These changes align the reference of the artifacts so that we always consume the artifacts published by Arcade at `https://ci.dot.net/public/source-build`. This allows `artifactsRid` to be set in all cases and be able to retrieve the corresponding tarball for both PSB and shared components.

Furthermore, this means that SB artifact publishing can be removed from the SB release process as these won't be consumed anymore.

### DIFF
--- a/eng/pipelines/templates/stages/source-build-and-validate.yml
+++ b/eng/pipelines/templates/stages/source-build-and-validate.yml
@@ -39,17 +39,17 @@ stages:
           container:
             name: ${{ variables.centOSStreamContainerName }}
             image: ${{ variables.centOSStreamContainerImage }}
+          artifactsRid: ${{ variables.centOSStreamX64Rid }}
           ${{ if ne(leg.reuseBuildArtifactsFrom, '') }}:
             reuseBuildArtifactsFrom:
             - ${{ format(leg.reuseBuildArtifactsFrom, variables.centOSStreamName) }}
-          ${{ if and(contains(leg.buildNameFormat, 'Previous'), eq(leg.targetArchitecture, 'x64')) }}:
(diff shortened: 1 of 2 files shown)


mthalman:
> Can you make sure an issue is opened to track the release pipeline publishing cleanup?

https://github.com/dotnet/source-build/issues/5317

PR #1911, opened 2025-08-15
Package installer test support for non-1xx branches

mthalman:
Contributes to https://github.com/dotnet/dotnet/issues/1011

The package installer tests in the VMR pipeline require the ability to install the `dotnet-sdk` deb/rpm packages. But these packages are dependent on the runtime packages (e.g. `dotnet-runtime`) which are not produced in a non-1xx branch of the VMR.

To solve this, these dependent packages are downloaded from ci.dot.net where they've been published by Arcade for each VMR build. The version of the 1xx build is determined by the VMR's configuration where it has a `MicrosoftNETSdkVersion` property that indicates this dependency. The condition for knowing whether this is in the context of a 1xx branch or not is determined by the `DotNetBuildSharedComponents` property which is `true` only in the context of a 1xx branch.

### DIFF
--- a/test/Microsoft.DotNet.Installer.Tests/Microsoft.DotNet.Installer.Tests.csproj
+++ b/test/Microsoft.DotNet.Installer.Tests/Microsoft.DotNet.Installer.Tests.csproj
@@ -44,6 +44,12 @@
       <RuntimeHostConfigurationOption Include="$(MSBuildProjectName).KeepDockerImages">
         <Value>$(KeepDockerImages)</Value>
       </RuntimeHostConfigurationOption>
+      <RuntimeHostConfigurationOption Include="$(MSBuildProjectName).DotNetBuildSharedComponents">
+        <Value>$(DotNetBuildSharedComponents)</Value>
+      </RuntimeHostConfigurationOption>
+      <RuntimeHostConfigurationOption Include="$(MSBuildProjectName).Sdk1xxVersion">
+        <Value>$(MicrosoftNETSdkVersion)</Value>
+      </RuntimeHostConfigurationOption>
     </ItemGroup>
   </Target>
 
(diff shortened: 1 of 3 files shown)


test/Microsoft.DotNet.Installer.Tests/Microsoft.DotNet.Installer.Tests.csproj
```diff
@@ -44,6 +44,12 @@
       <RuntimeHostConfigurationOption Include="$(MSBuildProjectName).KeepDockerImages">
         <Value>$(KeepDockerImages)</Value>
       </RuntimeHostConfigurationOption>
+      <RuntimeHostConfigurationOption Include="$(MSBuildProjectName).DotNetBuildSharedComponents">
+        <Value>$(DotNetBuildSharedComponents)</Value>
+      </RuntimeHostConfigurationOption>
+      <RuntimeHostConfigurationOption Include="$(MSBuildProjectName).Sdk1xxVersion">
+        <Value>$(MicrosoftNETSdkVersion)</Value>
```
NikolaMilosavljevic, line 51: I understand that this property is specifying the 1xx SDK version. Do you know why the property name isn't versioned, to avoid confusion with which version it specifies, especially in non-1xx builds?

mthalman: Because it's a property of a darc dependency just like any other property. Those are inherently not versioned in the name. They are based off the package name.

NikolaMilosavljevic: I see, so `MicrosoftNETSdkVersion` isn't defined in 1xx builds. Is it not going to fail here? https://github.com/dotnet/dotnet/blob/ad3682555c96aa9e03371a6264f2d450c37a47d7/test/Microsoft.DotNet.Installer.Tests/Config.cs#L46

mthalman: It should be empty. That config value won't be used if `DotNetBuildSharedComponents` is `false`.

Issue #1758, opened 2025-08-05
Enable test signing in VMR build


mmitche:
The VMR pipeline templates have support for test signing, but we need a re-bootstrap to pick up the required template changes to avoid using the ESRP service connection....

After a rebootstrap that picks up https://github.com/dotnet/arcade/commit/2586309f3f8553152e2d1d54612e9199975f9cd4, uncommment the lines in eng\pipelines\templates\jobs\vmr-build.yml around `install-microbuild.yml` step.

mmitche:
Test signing is enabled, but not perfectly functional, as the timestamps on the test signed nuget packages are not valid. This results in downstream restore failures.

mmitche:
Closing this, as #2360 covers the failing scenario

PR #1948, opened 2025-08-18
`SignType` variables must be lowercase

ellahathaway:
Arcade expects these values to be lowercase, when they are not, conditions that check the `DotNetSignType` value evaluate to false.

As explained in https://github.com/dotnet/dotnet/issues/1949, this doesn't affect real signing, but it does affect test signing.

### DIFF
--- a/eng/pipelines/templates/variables/vmr-build.yml
+++ b/eng/pipelines/templates/variables/vmr-build.yml
@@ -82,7 +82,7 @@ variables:
   - name: signEnabled
     value: true
   - name: signType
-    value: Real
+    value: real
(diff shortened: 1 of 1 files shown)


wtgodbe:
Do we need this in the 10.0 branches?

ellahathaway:
> Do we need this in the 10.0 branches?

Yeah, we should fix this in 10.0. I'll open a backport PR

wtgodbe:
Do we need RC1 as well?

ellahathaway:
> Do we need RC1 as well?

This only affects our ability to run test signing (https://github.com/dotnet/dotnet/issues/1758) and the signing of the SB artifacts (https://github.com/dotnet/dotnet/pull/1811) when the `Signed (Real)` pipeline parameter is passed for a manual pipeline run. I don't expect either to go into RC1. I think that if there's a need later, we can backport at that time.

NikolaMilosavljevic:
Presumably `DryRun` case was not affected - should that be updated to all lowercase for consistency?

PR #1957, opened 2025-08-18
Account for linux-musl in community platforms

am11:
https://github.com/dotnet/source-build/issues/5316

cc @mthalman

### DIFF
--- a/src/sdk/src/Layout/Directory.Build.props
+++ b/src/sdk/src/Layout/Directory.Build.props
@@ -56,9 +56,7 @@
     <!-- Crossgen2 is not bundled by default on platforms where Microsoft provides a package on nuget.org, 
          because it is large (100MB+). -->
     <_IsCommunityPlatform Condition="'$(OSName)' != 'win' and '$(OSName)' != 'osx'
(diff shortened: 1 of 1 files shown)


src/sdk/src/Layout/Directory.Build.props
```diff
@@ -55,8 +55,8 @@
 
     <!-- Crossgen2 is not bundled by default on platforms where Microsoft provides a package on nuget.org, 
          because it is large (100MB+). -->
-    <_IsCommunityPlatform Condition="'$(OSName)' != 'win' and '$(OSName)' != 'osx' 
-      and !('$(OSName)' == 'linux' and ('$(TargetArchitecture)' == 'x64' or '$(TargetArchitecture)' == 'x86' or '$(TargetArchitecture)' == 'arm' or '$(TargetArchitecture)' == 'arm64'))">true</_IsCommunityPlatform>
+    <_IsCommunityPlatform Condition="'$(OSName)' != 'win' and '$(OSName)' != 'osx'
+      and !($(OSName.StartsWith('linux')) and ('$(TargetArchitecture)' == 'x64' or '$(TargetArchitecture)' == 'x86' or '$(TargetArchitecture)' == 'arm' or '$(TargetArchitecture)' == 'arm64'))">true</_IsCommunityPlatform>
```
mthalman: Should this instead explicitly check for `linux` and `linux-musl` without a substring check? That way we avoid potentially matching on some unknown future name that starts with `linux` but is a community platform.

PR #1983, opened 2025-08-19
Include shared component packages in SB artifacts

mthalman:
Contributes to #1011

In a non-1xx build, the stage 2 source-only build fails in the build of `source-build-reference-packages` with errors like the following:
`/__w/1/s/src/source-build-reference-packages/src/targetPacks/ILsrc/microsoft.netcore.app.ref/6.0.0/Microsoft.NETCore.App.Ref.6.0.0.csproj error NU1603: Microsoft.NETCore.App.Ref.6.0.0 depends on runtime.centos.10-x64.Microsoft.NETCore.ILAsm (>= 10.0.0-preview.7.25377.103) but runtime.centos.10-x64.Microsoft.NETCore.ILAsm 10.0.0-preview.7.25377.103 was not found. runtime.centos.10-x64.Microsoft.NETCore.ILAsm 10.0.0-rc.1.25405.108 was resolved instead. [/__w/1/s/.dotnet/sdk/10.0.200-rc.1.25408.102/NuGet.targets]`

The issue here is that a dependency to `runtime.centos.10-x64.Microsoft.NETCore.ILAsm` is being defined targeting the version that is defined directly in the repo's [`MicrosoftNETCoreILAsmVersion` property](https://github.com/dotnet/source-build-reference-packages/blob/299471506deb40dc03945ecc920932f95027fec8/eng/Version.Details.props#L14) (in this case `10.0.0-preview.7.25377.103`. It's not being overridden with the version that was produced by the stage 1 build (in this case `10.0.0-rc.1.25405.108`). It doesn't get overridden because that package is not listed in the versions props file that exists in the SB artifacts tarball produced by the stage 1 build. In fact, _none_ of the packages which originate from the shared components are listed in there.

The fix is to ensure that those properties are written. This also applies to the VersionManifest.xml file which also needs to accurately reflect this state as there are dependencies on that file as well.

### DIFF
--- a/eng/PublishSourceBuild.props
+++ b/eng/PublishSourceBuild.props
@@ -1,4 +1,6 @@
 <Project>
+  <Import Project="$(RepositoryEngineeringDir)shared-source-only.targets" />
+
   <!-- Add targets here that run logic that depends on the merged asset manifest and published packages. -->
   <UsingTask TaskName="Microsoft.DotNet.SourceBuild.Tasks.LeakDetection.CheckForPoison" AssemblyFile="$(MicrosoftDotNetSourceBuildTasksLeakDetectionAssembly)" TaskFactory="TaskHostFactory" Condition="'$(EnablePoison)' == 'true'" />
   <Target Name="ReportPoisonUsage"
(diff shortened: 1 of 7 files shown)


eng/PublishSourceBuild.props
```diff
@@ -80,14 +82,21 @@
     </PropertyGroup>
   </Target>
 
+  <PropertyGroup>
+    <SharedComponentFilterMode>NonToolingOnly</SharedComponentFilterMode>
```
NikolaMilosavljevic, line 86: The default filter mode is changing from `NonToolingAndBootstrapOnly`. Presumably this is expected?

mthalman: Yes, it's intentional to use `NonToolingOnly` here.

PR #2013, opened 2025-08-20
Free up disk space by deleting simulator runtimes on macOS

akoeplinger:
We don't use these iOS/tvOS simulators in the VMR build and they take up a significant amount of disk space.

Available disk space increased from 44GB to 160GB and allows us to turn off clean-while-building for macOS jobs as well.

### DIFF
--- a/eng/pipelines/templates/jobs/vmr-build.yml
+++ b/eng/pipelines/templates/jobs/vmr-build.yml
@@ -224,17 +224,6 @@ jobs:
       value: '--'
 
   ### Basic arguments
-  - name: cleanArgument
-    # Don't clean-while-building on OSs that have enough disk space to keep
-    # intermediates for compliance scanners.
-    ${{ if and(eq(variables['System.TeamProject'], 'internal'), ne(parameters.targetOS, 'osx')) }}:
-      value: ''
-    ${{ else }}:
-      ${{ if eq(parameters.targetOS, 'windows') }}:
-        value: $(commandPrefix)cleanWhileBuilding
-      ${{ else }}:
-        value: $(commandPrefix)clean-while-building
-
   - name: brandingArgument
     value: $(commandPrefix)branding $(brandingType)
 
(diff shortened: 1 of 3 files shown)


eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -224,17 +224,6 @@ jobs:
...
-  - name: cleanArgument
-    # Don't clean-while-building on OSs that have enough disk space to keep
-    # intermediates for compliance scanners.
-    ${{ if and(eq(variables['System.TeamProject'], 'internal'), ne(parameters.targetOS, 'osx')) }}:
-      value: ''
-    ${{ else }}:
-      ${{ if eq(parameters.targetOS, 'windows') }}:
-        value: $(commandPrefix)cleanWhileBuilding
-      ${{ else }}:
-        value: $(commandPrefix)clean-while-building
```
MichaelSimons: distro maintainers rely on this option.  I would prefer that SB still runs at least one configuration with this setting to provide some coverage with it.

akoeplinger: sure. let's keep it in public and just remove it from internal where all the compliance scanners run

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -741,3 +743,5 @@ jobs:
           PathtoPublish: $(Build.ArtifactStagingDirectory)/artifacts/manifests/${{ parameters.configuration }}/$(Agent.JobName).xml
           ArtifactName: VerticalManifests
         displayName: Publish Vertical Manifest
+
```
MichaelSimons: Are these two added lines intentional/necessary?

akoeplinger: ah no, that's just because I used the GitHub web editor... Fixed

PR #2046, opened 2025-08-21
Add differentiator to SB CI job names

MichaelSimons:
The differentiator prefix makes it easier to distinguish the SB artifacts and test results from the others.

[Test Build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2777154) (Microsoft internal link)

### DIFF
--- a/eng/pipelines/templates/stages/source-build-sdk-diff-tests.yml
+++ b/eng/pipelines/templates/stages/source-build-sdk-diff-tests.yml
@@ -13,7 +13,7 @@ stages:
   jobs:
   - template: /eng/pipelines/templates/jobs/sdk-diff-tests.yml@self
     parameters:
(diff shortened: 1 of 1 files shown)


MichaelSimons:
> This file needs to reflect these changes: https://github.com/dotnet/dotnet/blob/main/eng/pipelines/templates/stages/source-build-sdk-diff-tests.yml
> 
> The release pipeline will also need updating.

Thanks - I fixed the sdk diff pipeline in https://github.com/dotnet/dotnet/pull/2046/commits/b1d9fb2078d99467346a071d90d5e6481aacd058.  I will log a follow-up issue to update the release pipeline.  That will have to be coordinated with the RC1 release as well as backporting this to the 10 release branch.

PR #2107, opened 2025-08-25
Update the testhost root rather than the installer root to avoid impa…

wtgodbe:
…cting the installer msi

https://github.com/dotnet/sdk/pull/50473


MichaelSimons:
Merging ahead of validation to unblock other work per request.

PR #2110, opened 2025-08-25
Create and publish source tarball and zipball

ellahathaway:
Related to https://github.com/dotnet/release/issues/1416

This PR enables creating and publishing the source as a build artifact.

[Test run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2780028&view=results)

### DIFF
--- a/eng/finish-source-only.proj
+++ b/eng/finish-source-only.proj
@@ -264,6 +264,7 @@
         SourceCommit="$(RepositoryCommit)"
         ArtifactName="$(SourceBuiltSourceArtifactName)"
         ArtifactVersion="$(SourceBuiltSdkNonStableVersion)"
+        RepoRoot="$(RepoRoot)"
         OutputDirectory="$(ArtifactsAssetsDir)" />
   </Target>
 
--- a/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
+++ b/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
@@ -33,6 +33,12 @@ public class CreateSourceArtifacts : BuildTask
     public required string ArtifactVersion { get; init; }
 
     /// <summary>
+    /// Path to the root of the repo
+    /// </summary>
+    [Required]
+    public required string RepoRoot { get; init; }
+
+    /// <summary>
     /// Path to the directory to place the source artifacts in
     /// </summary>
     [Required]
@@ -75,7 +81,10 @@ public class CreateSourceArtifacts : BuildTask
 
         try
         {
-            ProcessResult result = await processService.RunProcessAsync("git", $"archive -o {artifactFilePath} {SourceCommit}");
+            ProcessResult result = await processService.RunProcessAsync(
+                "git",
+                $"archive -o {artifactFilePath} {SourceCommit}",
+                workingDirectory: RepoRoot);
 
             if (result.ExitCode != 0)
             {
--- a/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
+++ b/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
@@ -19,7 +19,7 @@ public class ProcessService(TaskLoggingHelper log, int timeout = 10 * 1000)
     private TaskLoggingHelper Log { get; } = log;
     private TimeSpan Timeout { get; } = TimeSpan.FromMilliseconds(timeout);
 
-    public async Task<ProcessResult> RunProcessAsync(string command, string arguments, bool printOutput = false)
+    public async Task<ProcessResult> RunProcessAsync(string command, string arguments, string workingDirectory = "", bool printOutput = false)
     {
         var processInfo = new ProcessStartInfo
         {
@@ -28,7 +28,8 @@ public class ProcessService(TaskLoggingHelper log, int timeout = 10 * 1000)
             RedirectStandardOutput = true,
             RedirectStandardError = true,
             UseShellExecute = false,
-            CreateNoWindow = true
+            CreateNoWindow = true,
+            WorkingDirectory = workingDirectory
         };
 
         using var process = new Process { StartInfo = processInfo, EnableRaisingEvents = true };


eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
```diff
@@ -0,0 +1,95 @@
...
+    [Required]
+    public required string ArtifactVersion { get; init; }
+
+    /// <summary>
+    /// Path to the directory to place the source artifacts in
+    /// </summary>
+    [Required]
+    public required string OutputDirectory { get; init; }
+
+    public const string ArtifactRootName = "dotnet-source";
```
MichaelSimons: Consider defining this value in an msbuild property, pass the value in as a parameter, and then share the value with publishing logic.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
```diff
@@ -0,0 +1,95 @@
...
+        string artifactFileName = $"{ArtifactRootName}-{ArtifactVersion}{artifactType.GetArtifactExtension()}";
+        string artifactFilePath = Path.Combine(OutputDirectory, artifactFileName);
+
+        Log.LogMessage(MessageImportance.High, $"Creating {artifactType} source artifact at: {artifactFilePath}");
+
+        ProcessResult result = await processService.RunProcessAsync("git", $"archive -o {artifactFilePath} {SourceCommit}");
+
+        if (result.ExitCode != 0)
+        {
+            throw new InvalidOperationException($"Failed to create {artifactType} source artifact: {result.Error}");
```
MichaelSimons: Consider not throwing here given the calling method is catching all exceptions.  It would be more efficient to pass in the errors collection. Alternatively, could you just log the error right here?

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    public const string ArtifactRootName = "dotnet-source";
+
+    public override bool Execute()
+    {
+        try
+        {
+            return ExecuteAsync().GetAwaiter().GetResult();
+        }
+        catch (Exception ex)
```
MichaelSimons: Curious why the catch is necessary here when ExecuteAsync is also catching?

MichaelSimons (commented):
I like this approach!

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/2110/commits/cc94125f3bd7a2fcd02f9ffad6929f00e1376ef9

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/2110/commits/e9541625d6745dbfebb764aec8f746b8ec6f9e97. I moved exception handling into CreateArtifactAsync. Errors are now collected in the shared errors queue during artifact creation, rather than being thrown in the function and caught at a higher level.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
ellahathaway: I removed this in https://github.com/dotnet/dotnet/pull/2110/commits/e9541625d6745dbfebb764aec8f746b8ec6f9e97

eng/Publishing.props
```diff
@@ -114,6 +119,7 @@
     <ItemGroup Condition="'$(IsPublishPass2)' == 'true' and '$(DotNetBuildSourceOnly)' == 'true'">
       <SourceBuildArtifact Include="$(ArtifactsAssetsDir)/$(SourceBuiltArtifactsTarballName).*$(ArchiveExtension)" />
       <SourceBuildArtifact Include="$(ArtifactsAssetsDir)/**/dotnet-sdk-*$(ArchiveExtension)" />
+      <SourceBuildArtifact Include="$(ArtifactsAssetsDir)/$(SourceBuiltSourceArtifactName)-*" />
```
MichaelSimons, line 122: Should this match the pattern used above?  It feels inconsistent with the way it is implemented.

MichaelSimons (approved):

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Threading.Tasks;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.DotNet.UnifiedBuild.Tasks.Services;
+
+public record ProcessResult(string Output, string Error, int ExitCode);
+
+public class ProcessService(TaskLoggingHelper log, int timeout = 10 * 1000)
```
NikolaMilosavljevic, line 17: Wondering if 10 sec for a default timeout is too short. `CreateSourceArtifacts` task doesn't override this value.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        if (!process.Start())
+        {
+            throw new InvalidOperationException($"Failed to start process: {command} {arguments}");
+        }
+
+        process.BeginOutputReadLine();
+        process.BeginErrorReadLine();
+
+        using var timeoutCts = new CancellationTokenSource(Timeout);
```
mthalman, line 81: Using the ctor overload that takes a `TimeSpan` instead of `int` would be better for readability, IMO.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
ellahathaway: I think 10 seconds is more than enough for most processes. Processes that take longer can choose to override the value.

NikolaMilosavljevic: My local tests of the `git archive` command are taking more than a minute. Granted, it is in a docker container. I presume this took much shorter in your tests.

eng/Publishing.props
ellahathaway: Updated the above pattern in https://github.com/dotnet/dotnet/pull/2110/commits/cb365663f446dfd55ec41c033e7d53ae07132f61. The pattern should be `$(ArtifactsAssetsDir)$(SourceBuiltSourceArtifactName)-*` in order to match both `.tar.gz` and `.zip` extension (`ArchiveExtension` equals `.tar.gz`).

As for the difference in path separators between the patterns, `ArtifactsAssetsDir` ends with a path separator, so I can remove it from these patterns if we'd like, but the binlogs show no difference in value with and without it.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/2110/commits/1cea03f155aa1358a8e49e36b6a83e8ed2a2eb43

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
ellahathaway: I see. Running this is CI is very quick so the 10 second timeout is enough there. But we should still support local scenarios, so I've increased the timeout to 5 minutes: 8f9c2136c6f8e3cf2b58875bb9a6ba7a1d53a229

eng/finish-source-only.proj
```diff
@@ -253,6 +253,20 @@
...
+  <Target Name="CreateSourceArtifacts"
+          BeforeTargets="Build"
+          DependsOnTargets="WriteSourceBuiltSdkVersionFile"
+          Condition="'$(CreateSourceArtifacts)' == 'true'" >
+
+    <CreateSourceArtifacts
+        SourceCommit="$(RepositoryCommit)"
+        ArtifactName="$(SourceBuiltSourceArtifactName)"
+        ArtifactVersion="$(SourceBuiltSdkNonStableVersion)"
+        OutputDirectory="$(ArtifactsAssetsDir)" />
```
NikolaMilosavljevic, line 268: I think you need to pass the working dir here, otherwise `git archive` would use the current working dir.

ellahathaway: You're right, this was only archiving the eng directory. Thank you for catching this. I've fixed this in https://github.com/dotnet/dotnet/pull/2110/commits/3af6eac7063381081510ee5e356a7a226c7562ef

NikolaMilosavljevic (approved):

PR #2132, opened 2025-08-26
Fix NuGetConfig in incremental VMR builds

ViktorHofer:
Incremental VMR builds were failing because the updated NuGet.config file with the local feeds was being overwritten.

This was due to the `CopyNuGetConfig` target always running (no Inputs & Outputs) but the `UpdateNuGetConfig` target only ran when the input was newer than the output.

Fix this by consolidating the two targets.

### DIFF
--- a/repo-projects/Directory.Build.targets
+++ b/repo-projects/Directory.Build.targets
@@ -145,8 +145,14 @@
     </ItemGroup>
   </Target>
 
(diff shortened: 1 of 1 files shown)


ViktorHofer:
@ericstj this fixes the failure that you noticed weeks ago that you showed me. I just ran into this as well.

akoeplinger (approved):

PR #2139, opened 2025-08-26
Add `succeeded()` to conditions where it is missing

ericstj:
I'm certain this is not the right way to do this, but I wanted to share the results of what I found and determine if we need to take some of these changes and how to go about that.

@ViktorHofer @jkoritzinsky @hoyosjs @mmitche - would appreciate your first look.  Let me know if you think there is value in this and how to go about it.  Maybe start with arcade, then let it flow, then apply to repo-specific pipelines?

These changes were all made by a script that I worked on generating with copilot -- https://gist.github.com/ericstj/ebc1823b3032da809c44e7497d588d79

### DIFF
--- a/src/cecil/eng/azure-pipelines.yml
+++ b/src/cecil/eng/azure-pipelines.yml
@@ -81,7 +81,6 @@ stages:
 
       - ${{ if eq(variables.officialBuild, 'false') }}:
         - job: Linux
-          condition: and(succeeded(), eq(variables.officialBuild, 'false'))
           pool:
             vmImage: ubuntu-latest
           steps:
--- a/src/cecil/eng/official.yml
+++ b/src/cecil/eng/official.yml
@@ -92,7 +92,6 @@ extends:
                 displayName: Build and publish Mono.Cecil.sln $(_BuildConfig)
           - ${{ if eq(variables.officialBuild, 'false') }}:
             - job: Linux
-              condition: and(succeeded(), eq(variables.officialBuild, 'false'))
               pool:
                 vmImage: ubuntu-latest
               steps:


akoeplinger:
I think it makes sense but it's very hard to see the actual changes when the eng/common findings are duplicated across all repos. can you remove those except for src/arcade?

ericstj:
Good suggestion.  I reverted everything that was under eng\common outside of arcade.

eng/pipelines/pr.yml
```diff
@@ -51,6 +51,7 @@ stages:
   displayName: Validate user changes in VMR
   dependsOn: []
   condition: and(
+              succeeded(),
```
akoeplinger: this one isn't needed since this stage doesn't depend on anything

src/arcade/eng/common/core-templates/post-build/post-build.yml
```diff
@@ -159,7 +159,7 @@ stages:
 
     - job:
       displayName: Signing Validation
-      condition: and( eq( ${{ parameters.enableSigningValidation }}, 'true'), ne( variables['PostBuildSign'], 'true'))
+      condition: and(succeeded(),  eq( ${{ parameters.enableSigningValidation }}, 'true'), ne( variables['PostBuildSign'], 'true'))
```
akoeplinger: this doesn't look correct since it'd make this job fail if the previous one (`NuGet Validation`) failed but these jobs are independent

src/arcade/eng/common/core-templates/post-build/post-build.yml
```diff
@@ -220,7 +220,7 @@ stages:
 
     - job:
       displayName: SourceLink Validation
-      condition: eq( ${{ parameters.enableSourceLinkValidation }}, 'true')
+      condition: and(succeeded(), eq( ${{ parameters.enableSourceLinkValidation }}, 'true'))
```
akoeplinger: same here

src/arcade/eng/common/core-templates/post-build/post-build.yml
akoeplinger: that said, I don't know why the `NuGet Validation` job uses `succeededOrFailed` since it's the first job in the stage so this should be a no-op

akoeplinger: ah I guess this catches the case where the build is canceled: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops&preserve-view=true#succeededorfailed

> You may want to use `not(canceled())` instead when there are previous skipped jobs in the dependency graph.

src/cecil/eng/azure-pipelines.yml
```diff
@@ -81,7 +81,7 @@ stages:
 
       - ${{ if eq(variables.officialBuild, 'false') }}:
         - job: Linux
-          condition: eq(variables.officialBuild, 'false')
+          condition: and(succeeded(), eq(variables.officialBuild, 'false'))
```
akoeplinger: this whole condition can be removed since we check it already two lines above

src/cecil/eng/official.yml
```diff
@@ -92,7 +92,7 @@ extends:
                 displayName: Build and publish Mono.Cecil.sln $(_BuildConfig)
           - ${{ if eq(variables.officialBuild, 'false') }}:
             - job: Linux
-              condition: eq(variables.officialBuild, 'false')
+              condition: and(succeeded(), eq(variables.officialBuild, 'false'))
```
akoeplinger: same here

src/fsharp/azure-pipelines-PR.yml
```diff
@@ -339,7 +339,7 @@ stages:
             continueOnError: true
           - task: PublishBuildArtifacts@1
             displayName: Publish Build BinLog
-            condition: eq(variables['System.JobPositionInPhase'], 1)
+            condition: and(succeeded(), eq(variables['System.JobPositionInPhase'], 1))
```
akoeplinger: I think this would change the behavior, right now we still publish build logs if the PublishTestResults fails (which has `continueOnError: true`)

src/fsharp/azure-pipelines-PR.yml
```diff
@@ -580,7 +580,7 @@ stages:
 
           - task: PublishBuildArtifacts@1
             displayName: Publish BinLog
-            condition: eq(variables['System.JobPositionInPhase'], 1)
+            condition: and(succeeded(), eq(variables['System.JobPositionInPhase'], 1))
```
akoeplinger: same here

src/fsharp/azure-pipelines.yml
```diff
@@ -154,15 +154,15 @@ extends:
                 testResultsFiles: '*.xml'
                 searchFolder: '$(Build.SourcesDirectory)/artifacts/TestResults/$(_BuildConfig)'
               continueOnError: true
-              condition: ne(variables['SkipTests'], 'true')
+              condition: and(succeeded(), ne(variables['SkipTests'], 'true'))
```
akoeplinger: this also sounds like we'd not publish test results if the previous step fails (which I assume it does if there are test failures)

src/msbuild/.vsts-dotnet-ci.yml
```diff
@@ -98,7 +98,7 @@ jobs:
       publishRunAttachments: true
       mergeTestResults: true
     continueOnError: true
-    condition: eq(variables.onlyDocChanged, 0)
+    condition: and(succeeded(), eq(variables.onlyDocChanged, 0))
```
akoeplinger: another case of test results publishing behavior change (applies to the rest of the changes in this file)

src/razor/azure-pipelines-official.yml
```diff
@@ -256,7 +256,7 @@ extends:
                 targetPath: artifacts/TestResults/$(_BuildConfig)
                 artifact: $(Agent.Os)_$(Agent.JobName) Attempt $(System.JobAttempt) TestResults
               displayName: Publish Test Artifacts
-              condition: in(variables['Build.Reason'], 'PullRequest')
+              condition: and(succeeded(), in(variables['Build.Reason'], 'PullRequest'))
```
akoeplinger: another case of test results publishing behavior change

src/razor/azure-pipelines.yml
```diff
@@ -173,7 +173,7 @@ stages:
         - publish: artifacts/TestResults/$(_BuildConfig)
           artifact: $(Agent.Os)_$(Agent.JobName) Attempt $(System.JobAttempt) TestResults
           displayName: Publish Test Artifacts
-          condition: in(variables['Build.Reason'], 'PullRequest')
+          condition: and(succeeded(), in(variables['Build.Reason'], 'PullRequest'))
```
akoeplinger: another case of test results publishing behavior change

src/runtime/eng/pipelines/coreclr/templates/jit-replay-pipeline.yml
```diff
@@ -43,7 +43,7 @@ extends:
                   archiveExtension: $(archiveExtension)
                   artifactName: BuildArtifacts_$(osGroup)$(osSubgroup)_$(archType)_$(_BuildConfig)
                   displayName: Build Assets
-            condition: not(eq(stageDependencies.EvaluatePaths.evaluate_paths.outputs['SetPathVars_jiteeversionguid.containsChange'], true))
+            condition: and(succeeded(), not(eq(stageDependencies.EvaluatePaths.evaluate_paths.outputs['SetPathVars_jiteeversionguid.containsChange'], true)))
```
akoeplinger, line 46: I'm not sure this is correct, would need to follow the template expansion. I'd assume it's a no-op since this job probably doesn't depend on anything

akoeplinger (commented):
did a quick pass, there are some changes that would change behavior around test publishing

src/fsharp/azure-pipelines-PR.yml
ericstj: I think for cases where we want them to run always we could be explicit with either `succeededOrFailed()` or `always()` what do you think?

ericstj:
> did a quick pass, there are some changes that would change behavior around test publishing

Thanks for doing this, I wasn't expecting a line by line review.  I was more looking for categorical review for patterns and a value assessment.  You found some good patterns though, and I can adjust for those.

Do you think this is adding value or just noise?  Do you see enough real fixes to warrant deeper analysis?

Agree that many of the changes will touch some steps that we want to run all the time.  Does it make sense to be explicit with those conditions?  Do you see patterns we could add to a scanning tool to distinguish these (for instance, steps that mention "log" or "test" could be made to run `always()`)?

src/arcade/eng/common/core-templates/post-build/post-build.yml
ericstj: Maybe for the `job` conditions we can just leave them alone.  It's less obvious what their behavior should be.

src/runtime/eng/pipelines/coreclr/templates/jit-replay-pipeline.yml
ericstj: I'll remove the changes made to jobs and templates - I think they were lower confidence anyhow.

src/fsharp/azure-pipelines-PR.yml
akoeplinger: yep

akoeplinger:
I think it's good to uncover these, even though I haven't seen a case where the previous behavior was necessarily causing something bad to happen.

PR #2140, opened 2025-08-26
Adjust non-portable RID handling for tool packs to handle cross builds between non-portable RIDs

jkoritzinsky:
Try using the non-portable cg2/ilc when targeting any non-portable RID, not just the non-portable RID matching the SDK RID.

Fixes dotnet/source-build#5225

### DIFF
--- a/src/sdk/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
+++ b/src/sdk/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
@@ -813,12 +813,20 @@ namespace Microsoft.NET.Build.Tasks
 
                 var runtimeGraph = new RuntimeGraphCache(this).GetRuntimeGraph(RuntimeGraphPath);
 
-                // If the provided RID is compatible with a non-portable RID in the support list, then it's considered non-portable.
-                NuGetUtils.GetBestMatchingRidWithExclusion(runtimeGraph, RuntimeIdentifier, packSupportedPortableRuntimeIdentifiers, packSupportedRuntimeIdentifiers, out bool targetsNonPortableSdkRid);
+                // Prefer portable when the "supported RID" for the tool pack is the same RID as the "supported portable RID".
+                // This makes non-portable SDKs behave the same as portable SDKs except for the specific cases added to "supported", such as targeting the non-portable RID.
+                // This also ensures that targeting common RIDs doesn't require any non-portable assets that aren't packaged in the SDK by default.
+                // Due to size concerns, the non-portable ILCompiler and Crossgen2 aren't included by default in non-portable SDK distributions.
+                string supportedTargetRid = NuGetUtils.GetBestMatchingRid(runtimeGraph, RuntimeIdentifier, packSupportedRuntimeIdentifiers, out _);
+                string supportedPortableTargetRid = NuGetUtils.GetBestMatchingRid(runtimeGraph, RuntimeIdentifier, packSupportedPortableRuntimeIdentifiers, out _);
 
-                string hostRuntimeIdentifier = targetsNonPortableSdkRid ? NETCoreSdkRuntimeIdentifier : portableSdkRid;
+                bool usePortable = !string.IsNullOrEmpty(NETCoreSdkPortableRuntimeIdentifier) && supportedTargetRid == supportedPortableTargetRid;
+
                 // Get the best RID for the host machine, which will be used to validate that we can run crossgen for the target platform and architecture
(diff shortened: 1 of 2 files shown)


jkoritzinsky:
Validated the new experience:

1. Building a SB SDK for the same RID as the bootstrapping SB SDK will use the "live" crossgen2 as it does today.
2. Building a SB SDK for a different RID as the bootstrapping SB SDK will try to use the portable crossgen2 when CG2 is not bundled with the bootstrapping SDK (and as such won't work offline only).
3. When CG2 is bundled in the bootstrapping SB SDK, it will be used when building the new SB SDK for the new non-portable RID (matching the rules in the linked issue).

jkoritzinsky:
cc: @tmds please review and validate

src/sdk/src/Tasks/Microsoft.NET.Build.Tasks/ProcessFrameworkReferences.cs
```diff
@@ -802,17 +802,22 @@ private ToolPackSupport AddToolPack(
...
+                // This makes non-portable SDKs behave the same as portable SDKs except for the specific case of targetting a non-portable RID.
+                // This ensures that targeting portable RIDs doesn't require any non-portable assets that aren't packaged in the SDK.
+                // Due to size concerns, the non-portable ILCompiler and Crossgen2 aren't included by default in non-portable SDK distributions.
                 string portableSdkRid = !string.IsNullOrEmpty(NETCoreSdkPortableRuntimeIdentifier) ? NETCoreSdkPortableRuntimeIdentifier : NETCoreSdkRuntimeIdentifier;
-                bool targetsNonPortableSdkRid = RuntimeIdentifier == NETCoreSdkRuntimeIdentifier && NETCoreSdkRuntimeIdentifier != portableSdkRid;
+
+                var runtimeGraph = new RuntimeGraphCache(this).GetRuntimeGraph(RuntimeGraphPath);
+
+                // If the provided RID is compatible with a non-portable RID in the support list, then it's considered non-portable.
+                NuGetUtils.GetBestMatchingRidWithExclusion(runtimeGraph, RuntimeIdentifier, packSupportedPortableRuntimeIdentifiers, packSupportedRuntimeIdentifiers, out bool targetsNonPortableSdkRid);
```
tmds: This is meant to assign `targetsNonPortableSdkRid` by taking into account `packSupportedPortableRuntimeIdentifiers`.

Looking at [GetBestMatchingRidWithExclusion](https://github.com/dotnet/sdk/blob/6d31ccb7052fd098e5c0b732599b5cf82f1e6cd9/src/Common/NuGetUtils.NuGet.cs#L86), it seems `packSupportedPortableRuntimeIdentifiers` won't affect `targetsNonPortableSdkRid`.

tmds: Taking a shot at this:
```cs
// Prefer portable when "supported" has the same match as "portable supported".
// This makes non-portable SDKs behave the same as portable SDKs except for the specific cases added to "supported", such as targeting the non-portable RID.
// This also ensures that targeting common RIDs doesn't require any non-portable assets that aren't packaged in the SDK.
// Due to size concerns, the non-portable ILCompiler and Crossgen2 aren't included by default in non-portable SDK distributions.
string? supportedTargetRid = NuGetUtils.GetBestMatchingRid(runtimeGraph, RuntimeIdentifier, packSupportedRuntimeIdentifiers, out _);
string? supportedPortableTargetRid = NuGetUtils.GetBestMatchingRid(runtimeGraph, RuntimeIdentifier, packSupportedPortableRuntimeIdentifiers, out _);
bool usePortable = !string.IsNullOrEmpty(NETCoreSdkPortableRuntimeIdentifier) && supportedTargetRid == supportedPortableTargetRid;
hostRuntimeIdentifier = usePortable ? NuGetUtils.GetBestMatchingRid(runtimeGraph, NETCoreSdkPortableRuntimeIdentifier, packSupportedPortableRuntimeIdentifiers, out _)
                                    : NuGetUtils.GetBestMatchingRid(runtimeGraph, NETCoreSdkRuntimeIdentifier, packSupportedRuntimeIdentifiers, out _);
```

@jkoritzinsky wdyt?

jkoritzinsky: That looks good. I'll apply the changes and validate tomorrow.

jkoritzinsky: Yeah it seems like it's working better locally. Pushed with these changes.

mthalman (approved):

PR #2221, opened 2025-09-01
Installer tests for Linux package dependencies

NikolaMilosavljevic:
Contributes to https://github.com/dotnet/arcade/issues/16042

We've seen regressions in package dependencies in several builds. It affected both DEB and RPM packages, specifically the SDK one. Missing dependencies cause issues with installation of .NET SDK via Linux package managers.

These new tests are utilizing various classes in arcade for obtaining dependencies from installer packages. DEB classes were usable today as their protection level was `public`. Some RPM classes were `internal` - that was fixed with https://github.com/dotnet/arcade/pull/16079

This PR does the following:
- Enables dependency tests for Debian packages
- Adds a currently-skipped RPM test

## RPM test

Once https://github.com/dotnet/arcade/pull/16079 is available in bootstrap arcade: 1) remove the `Skip` attribute for rpm test, 2) update the following `throw` statement with a code block below in `GetPackageDependencies` method - this has been tested locally.
```
            throw new NotImplementedException("https://github.com/dotnet/arcade/pull/16079 is required for getting RPM package dependencies.");
```

Replace with:
```
            try
            {
                using FileStream rpmStream = File.OpenRead(packagePath);
                using RpmPackage rpmPackage = RpmPackage.Read(rpmStream);

                string[] requireNames = (string[])rpmPackage.Header.Entries.FirstOrDefault(e => e.Tag == RpmHeaderTag.RequireName).Value;
                if (requireNames == null || requireNames.Length == 0)
                {
                    return [];
                }

                var result = new HashSet<string>(StringComparer.Ordinal);

                for (int i = 0; i < requireNames.Length; i++)
                {
                    string name = requireNames[i];
                    if (string.IsNullOrWhiteSpace(name) || name.StartsWith("rpmlib(", StringComparison.Ordinal))
                    {
                        continue; // Skip internal rpm capabilities
                    }

                    result.Add(name);
                }

                return result.ToList();
            }
            catch (Exception ex)
            {
                _outputHelper.WriteLine($"Error parsing RPM package '{package}': {ex}");
                return [];
            }
```

### DIFF
--- a/test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
+++ b/test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
@@ -696,44 +696,44 @@ public partial class LinuxInstallerTests : IDisposable
 
     private static List<string> ParseDebControlDependencies(string contents)
     {
-            Match match = DependsLineRegex.Match(contents);
-            if (!match.Success)
+        Match match = DependsLineRegex.Match(contents);
+        if (!match.Success)
+        {
+            return [];
+        }
+
+        string dependsLine = match.Groups[1].Value.Trim();
+        if (dependsLine.Length == 0)
+        {
+            return [];
+        }
+
+        var results = new List<string>();
+
+        foreach (string segment in dependsLine.Split(','))
+        {
+            string part = segment.Trim();
+            if (part.Length == 0)
             {
-                return [];
+                continue;
             }
 
-            string dependsLine = match.Groups[1].Value.Trim();
-            if (dependsLine.Length == 0)
+            // If there are alternates (pkgA | pkgB), keep only the first one as the dependency
+            int pipeIndex = part.IndexOf('|');
+            if (pipeIndex >= 0)
             {
-                return [];
+                part = part.Substring(0, pipeIndex).Trim();
             }
 
-            var results = new List<string>();
+            part = RemoveVersionConstraintRegex.Replace(part, "").Trim();
 
-            foreach (string segment in dependsLine.Split(','))
+            // Skip native lib packages
+            if (part.Length > 0 && !part.StartsWith("lib"))
             {
-                string part = segment.Trim();
-                if (part.Length == 0)
-                {
-                    continue;
-                }
-
-                // If there are alternates (pkgA | pkgB), keep only the first one as the dependency
-                int pipeIndex = part.IndexOf('|');
-                if (pipeIndex >= 0)
-                {
-                    part = part.Substring(0, pipeIndex).Trim();
-                }
-
-                part = RemoveVersionConstraintRegex.Replace(part, "").Trim();
-
-                // Skip native lib packages
-                if (part.Length > 0 && !part.StartsWith("lib"))
-                {
-                    results.Add(part);
-                }
+                results.Add(part);
             }
+        }
 
-            return results;
+        return results;
     }
 }


test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -16,6 +16,8 @@
 using TestUtilities;
 using Xunit;
 using Xunit.Abstractions;
+using System.Formats.Tar;
+using Microsoft.DotNet.Build.Tasks.Installers;
```
mthalman: nit: sort these

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -496,4 +542,203 @@ private async Task DownloadFileAsync(Uri url, string filePath)
...
+    {
+        foreach (string package in list)
+        {
+            // Skip netstandard and runtime-deps packages as they are not expected to have .NET dependencies
+            if (package.StartsWith(NetStandardTargetingPackPrefix) || package.StartsWith(DotnetRuntimeDepsPrefix))
+            {
+                continue;
+            }
+
+            Assert.True(PackageContainsExpectedDependencies(package, packageType), $"Package {package} does not contain the expected dependencies.");
```
mthalman: Rather than asserting `True`, it'd be nicer if `PackageContainsExpectedDependencies` was changed to `AssertPackageContainsExpectedDependencies` and did the assertion it self. Then it can use the property `XUnit` comparison of the two sets. I think you'd want to use `Assert.Equal(expectedDependencies, dependencies)`. That gives you more informative output when it fails.

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -496,4 +542,203 @@ private async Task DownloadFileAsync(Uri url, string filePath)
...
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    private string GetPackagePrefixFromPackageName(string packageName)
+    {
+        Regex regex = new Regex(@"^(.*?-)(?=\d)");
```
mthalman: Use a source-generated regex field instead: `GeneratedRegexAttribute`.

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -496,4 +542,203 @@ private async Task DownloadFileAsync(Uri url, string filePath)
...
+                                using MemoryStream controlFileData = new MemoryStream();
+                                entry.DataStream?.CopyTo(controlFileData);
+                                string controlContent = Encoding.UTF8.GetString(controlFileData.ToArray());
+                                return ParseDebControlDependencies(controlContent);
+                            }
+                        }
+                    }
+                }
+
+                _outputHelper.WriteLine($"No control.tar.* entry found in {package}");
```
mthalman: Split this into a separate method.

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -496,4 +542,203 @@ private async Task DownloadFileAsync(Uri url, string filePath)
...
+                return [];
```
mthalman: Why is everything in this method always returning an empty list?

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -496,4 +542,203 @@ private async Task DownloadFileAsync(Uri url, string filePath)
...
+        }
+        else if (packageType == PackageType.Rpm)
+        {
+            throw new NotImplementedException("https://github.com/dotnet/arcade/pull/16079 is required for getting RPM package dependencies.");
+        }
+
+        return [];
+    }
+
+    private static List<string> ParseDebControlDependencies(string controlContent)
```
mthalman: Can a regex handle this instead?

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/2221/commits/8e5636ba1fbbfcd4cc946fbd14e5102de8acc18d

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/2221/commits/53236a4b63bc31a06e1088a01f0e62625e9b144c

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/2221/commits/53236a4b63bc31a06e1088a01f0e62625e9b144c

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
NikolaMilosavljevic: I improved this with 2 regexes in https://github.com/dotnet/dotnet/pull/2221/commits/4f938cdadde1dc128876b6b245fc0d0da43b11ec

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/2221/commits/4f938cdadde1dc128876b6b245fc0d0da43b11ec

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -496,4 +559,181 @@ private async Task DownloadFileAsync(Uri url, string filePath)
...
+            return [];
+        }
+        catch (Exception ex)
+        {
+            _outputHelper.WriteLine($"Error parsing DEB package '{packagePath}': {ex}");
+            return [];
+        }
+    }
+
+    private static List<string> ParseDebControlDependencies(string contents)
```
NikolaMilosavljevic, line 697: Fixed indentation of this method with https://github.com/dotnet/dotnet/pull/2221/commits/4ed8bb06b40117a6ea51f4bb4de7b9dfdff360f0

mthalman (approved):

PR #2270, opened 2025-09-04
[release/10.0.1xx] Add roslyn and msbuild dependencies to efcore

AndriySvyryd:


mmitche:
@AndriySvyryd What's the motivation for this change?

AndriySvyryd:
@mmitche We need to depend on the live versions of msbuild and roslyn packages to get the newest fixes and support net10 projects, as well as aligning our support policies with dependencies

dibarbet:
> /cc @dibarbet We might have possibility of a broken package graph here.
> 
> Some of efcore's shipping packages depend on roslyn's/msbuild's packages built in the VMR. They're also the "live" version after this change. This is in contrast to 9.0, where the dependencies were on older released versions (at least msbuild).
> 
> This is going to cause issues in a few ways:
> 
> * Roslyn doesn't stabilize in the VMR, since it doesn't ship packages from the VMR. When we build stable, this will cause the efcore package builds to fail, due to building stable packages with non-stable dependencies. See https://dev.azure.com/dnceng/internal/_build/results?buildId=2789004&view=logs&j=9050e078-31bf-5111-d8ec-8b6fa95caf9c&t=ed25c71b-10cf-587f-5b90-d13c9f84858f&l=1435
> * The non-stable roslyn packages built will not be released to nuget.org on the .NET release day (at least not automatically), since roslyn controls their own publishing. This we could resolve by manually publishing if need be, though.
> * If efcore needs to have dependencies on the 5.0.0 final stable packages, then we're in a real pickle because they won't be built in the VMR (since roslyn doesn't stabilize), and they also won't and probably **can't** be available on release day. Roslyn's packages depend on VS packages that likely won't have stable versions available until a few days after GA, so roslyn will need to wait around for these, update its dependencies, and then finally publish 5.0.0.
> 
> For 9.0, some of these dependencies were updated to stable, released versions (17.8.3?) prior to release day, so we didn't have this problem. For 10.0. 5.0.0 roslyn packages aren't published from the VMR (so those versions that your packages depend on won't be automatically published), and stable versions likely will not even be available on release day, due to some of roslyn's dependencies not being available on release day.

Commented here as well - https://github.com/dotnet/efcore/pull/36707#issuecomment-3276805289

Yes I think this is going to be a problem.  As you mentioned Roslyn is not part of SDK stablization, and we currently publish packages ourselves with separate stable versions.  When the SDK releases with these efcore packages, they're going to be referencing Roslyn versions that will never get published anywhere (in the current model).  So we hit issues like https://github.com/dotnet/sdk/issues/50697

While we do want to [publish Roslyn packages out of the VMR](https://github.com/dotnet/roslyn/issues/78298), due to the aforementioned VS dependencies we likely will not participate in the full stabilization process either, so it would not be safe to reference the very tip packages here.

One potential option is to split Roslyn's packages into two set, some that can participate in VMR stabilization and others that cannot.  Still investigating that, but these are larger changes.

AndriySvyryd:
Closing (for now) in favor of https://github.com/dotnet/efcore/pull/36742

PR #2271, opened 2025-09-05
Update branding to .NET 11

mmitche:


wtgodbe (approved):

src/sdk/eng/Versions.props
```diff
@@ -20,7 +20,7 @@
     <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
     <DotNetFinalVersionKind Condition="'$(StabilizePackageVersion)' == 'true'">release</DotNetFinalVersionKind>
     <!-- Calculate prerelease label -->
-    <PreReleaseVersionLabel Condition="'$(StabilizePackageVersion)' != 'true'">rc</PreReleaseVersionLabel>
+  <PreReleaseVersionLabel Condition="'$(StabilizePackageVersion)' != 'true'">alpha</PreReleaseVersionLabel>
```
akoeplinger: wrong indentation

mmitche: Bad AI

src/command-line-api/eng/Versions.props
```diff
@@ -2,8 +2,8 @@
   <Import Project="Version.Details.props" Condition="Exists('Version.Details.props')" />
   <!-- Repo Version Information -->
   <PropertyGroup>
-    <VersionPrefix>2.0.0</VersionPrefix>
-    <PreReleaseVersionLabel>rc</PreReleaseVersionLabel>
+    <VersionPrefix>3.0.0</VersionPrefix>
+    <PreReleaseVersionLabel>alpha</PreReleaseVersionLabel>
     <PreReleaseVersionIteration>1</PreReleaseVersionIteration>
```
ViktorHofer, line 8: @jonsequitur @adamsitnik @jeffhandley with this change, SCL will get updated from 2.0.0 (.NET 10) to 3.0.0 (.NET 11). Are you OK with that?

mmitche: I made an educated guess here. I think we're shipping 2.0.0 with GA, so that made some sense to me at least.

mmitche:
APICompat in the runtime build is unhappy about the version update, so more work to do for sure

ViktorHofer:
Yes, phase 1 needs to pin the assembly version in runtime. See https://github.com/dotnet/runtime/issues/118583. Sending a commit into this branch.

src/command-line-api/eng/Versions.props
jeffhandley: Yes, bumping to 3.0.0 works. Thanks for confirming!

mmitche:
This PR will need additional attention. The packages created after updating to net11 change their names, so lots of updates are required in V.D.xml

wtgodbe:
There are a couple extra changes required in aspnetcore (the changes to ApiExplorerTest.cs and TemplatePackageInstaller.cs from https://github.com/dotnet/aspnetcore/pull/57340/files, but updated for 11). If you want I can push to your branch

src/command-line-api/eng/Versions.props
adamsitnik: > @jonsequitur @adamsitnik @jeffhandley with this change, SCL will get updated from 2.0.0 (.NET 10) to 3.0.0 (.NET 11). Are you OK with that?

I am happy with that! (I would not be happy with jumping from 2 to 11 ;p )

mmitche:
10.0.0 PackageOverride.txt missing:

```
  D:\a\_work\1\s\src\sdk\src\Layout\redist\targets\GeneratePackagePruneData.targets(50,5): error MSB3030: Could not copy the file "D:\a\_work\1\s\src\sdk\artifacts\.packages\microsoft.netcore.app.ref\10.0.0\data\PackageOverrides.txt" because it was not found. [D:\a\_work\1\s\src\sdk\src\Layout\redist\redist.csproj]
##[error]src\sdk\src\Layout\redist\targets\GeneratePackagePruneData.targets(50,5): error MSB3030: (NETCORE_ENGINEERING_TELEMETRY=Build) Could not copy the file "D:\a\_work\1\s\src\sdk\artifacts\.packages\microsoft.netcore.app.ref\10.0.0\data\PackageOverrides.txt" because it was not found.
    dotnet-outputsframeworkversion-netcoreapp1.0 -> D:\a\_work\1\s\src\sdk\test\TestAssets\TestPackages\dotnet-outputsframeworkversion\dotnet-outputsframeworkversion-netcoreapp1.0\bin\Release\netcoreapp1.0\dotnet-outputsframeworkversion-netcoreapp1.0.dll
```

am11:
> Yes, phase 1 needs to pin the assembly version in runtime. See [dotnet/runtime#118583](https://github.com/dotnet/runtime/issues/118583). Sending a commit into this branch.

Can we initiate major version update from VMR and flow the changes to downstream repos all at once (without going through multiple passes)?

mmitche:
> > Yes, phase 1 needs to pin the assembly version in runtime. See [dotnet/runtime#118583](https://github.com/dotnet/runtime/issues/118583). Sending a commit into this branch.
> 
> Can we initiate major version update from VMR and flow the changes to downstream repos all at once (without going through multiple passes)?

We think so? But not sure...

ViktorHofer:
> Can we initiate major version update from VMR and flow the changes to downstream repos all at once (without going through multiple passes)?

Yes but not for the assembly version in runtime. We need to pin that until we update to the net11.0 TFM there. That's a much larger change and won't happen centralized in the VMR.

mmitche:
> There are a couple extra changes required in aspnetcore (the changes to ApiExplorerTest.cs and TemplatePackageInstaller.cs from https://github.com/dotnet/aspnetcore/pull/57340/files, but updated for 11). If you want I can push to your branch

That would be MUCH appreciated.

mmitche:
@ViktorHofer I updated the Package pruning target here. I thiiink this the right thing to do. Most likely this is going to fail, and i'll need to add a public RTM feed. https://github.com/dotnet/dotnet/pull/2271/commits/c9da95ad1203406aadeb9a73b2b268805616bba2

mmitche:
Actually, maybe it's not the right thing to do. Is the package download collation supposed to feed into the restore step? https://github.com/dotnet/dotnet/pull/2271/files#diff-f252ee6876261239bbd1e6fb70b5443fdede7e98479c5ba584f348e3295a94f9R34-R41

Basically, when are the 10.0 ref packs supposed to get downloaded.

mmitche:
I see what's going on...the target was correct before. But the package deduplication during restore is removing all the 10.0.0 packs:

```
<Target Name="CollectPackageDownloads" Returns="@(PackageDownload)">
    <!-- NOTE for design-time builds we need to ensure that we continue on error. -->
    <PropertyGroup>
      <CollectPackageDownloadsContinueOnError>$(ContinueOnError)</CollectPackageDownloadsContinueOnError>
      <CollectPackageDownloadsContinueOnError Condition="'$(ContinueOnError)' == '' ">false</CollectPackageDownloadsContinueOnError>
    </PropertyGroup>

    <CheckForDuplicateNuGetItemsTask
      Condition="'$(DisableCheckingDuplicateNuGetItems)' != 'true' "
      Items="@(PackageDownload)"
      ItemName="PackageDownload"
      LogCode="NU1505"
      MSBuildProjectFullPath="$(MSBuildProjectFullPath)"
      TreatWarningsAsErrors="$(TreatWarningsAsErrors)"
      WarningsAsErrors="$(WarningsAsErrors)"
      WarningsNotAsErrors="$(WarningsNotAsErrors)"
      NoWarn="$(NoWarn)"
      ContinueOnError="$(CollectPackageDownloadsContinueOnError)"
      >
      <Output TaskParameter="DeduplicatedItems" ItemName="DeduplicatedPackageDownloads" />
    </CheckForDuplicateNuGetItemsTask>

    <ItemGroup Condition="'@(DeduplicatedPackageDownloads)' != ''">
      <PackageDownload Remove="@(PackageDownload)" />
      <PackageDownload Include="@(DeduplicatedPackageDownloads)" />
    </ItemGroup>
  </Target>
```

mmitche:
This is because package download deduplication only deduplicates based on the item spec. It ignores all metadata.

src/aspnetcore/eng/Versions.props
```diff
@@ -7,7 +7,7 @@
 <Project>
   <Import Project="Version.Details.props" Condition="Exists('Version.Details.props')" />
   <PropertyGroup Label="Version settings">
-    <AspNetCoreMajorVersion>10</AspNetCoreMajorVersion>
+    <AspNetCoreMajorVersion>11</AspNetCoreMajorVersion>
     <AspNetCoreMinorVersion>0</AspNetCoreMinorVersion>
     <AspNetCorePatchVersion>0</AspNetCorePatchVersion>
     <PreReleaseVersionIteration>3</PreReleaseVersionIteration>
```
akoeplinger:
```suggestion
    <PreReleaseVersionIteration>1</PreReleaseVersionIteration>
```

mmitche:
https://github.com/NuGet/Home/issues/14627

src/sdk/eng/Versions.props
```diff
@@ -20,7 +20,7 @@
     <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
     <DotNetFinalVersionKind Condition="'$(StabilizePackageVersion)' == 'true'">release</DotNetFinalVersionKind>
     <!-- Calculate prerelease label -->
-    <PreReleaseVersionLabel Condition="'$(StabilizePackageVersion)' != 'true'">rc</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel Condition="'$(StabilizePackageVersion)' != 'true'">alpha</PreReleaseVersionLabel>
     <PreReleaseVersionLabel Condition="'$(StabilizePackageVersion)' == 'true' and $(VersionPrefix.EndsWith('00'))">rtm</PreReleaseVersionLabel>
     <PreReleaseVersionLabel Condition="'$(StabilizePackageVersion)' == 'true' and !$(VersionPrefix.EndsWith('00'))">servicing</PreReleaseVersionLabel>
     <PreReleaseVersionIteration Condition="'$(StabilizePackageVersion)' != 'true'">3</PreReleaseVersionIteration>
```
akoeplinger:
```suggestion
    <PreReleaseVersionIteration Condition="'$(StabilizePackageVersion)' != 'true'">1</PreReleaseVersionIteration>
```

src/sdk/src/Layout/redist/targets/GeneratePackagePruneData.targets
```diff
@@ -32,15 +32,15 @@
 
   <UsingTask TaskName="Microsoft.NET.Build.Tasks.CollatePackageDownloads" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
   <Target Name="AddTargetingPacksForPruneDataAsPackageDownloads"
-          BeforeTargets="CollectPackageDownloads"
+          AfterTargets="CollectPackageDownloads"
```
ViktorHofer: The change in L43 makes sense. I don't understand this line change though. If this is AfterTargets then I would assume that NuGet Restore which calls the target won't necessarily get the added PackageDownload items. cc @nkolev92

mmitche: Then I think we need to fix the issue with the package download Dedupe.

ViktorHofer: Should be fixed now with the latest commit

ViktorHofer:
Yeaaaah my fix worked :)

> src/sdk/src/Layout/redist/redist.csproj(0,0): error NU1102: (NETCORE_ENGINEERING_TELEMETRY=Restore) Unable to find package Microsoft.NETCore.App.Ref with version (= 10.0.0)
> src/sdk/src/Layout/redist/redist.csproj(0,0): error NU1102: (NETCORE_ENGINEERING_TELEMETRY=Restore) Unable to find package Microsoft.AspNetCore.App.Ref with version (= 10.0.0)
> src/sdk/src/Layout/redist/redist.csproj(0,0): error NU1102: (NETCORE_ENGINEERING_TELEMETRY=Restore) Unable to find package Microsoft.WindowsDesktop.App.Ref with version (= 10.0.0)

I'm not sure how source-build is supposed to have access to these prebuilt targeting packs (my guess: it must come via the SBRP repo). @dsplaisted the package pruning targeting pack data generation (https://github.com/dotnet/dotnet/blob/main/src/sdk/src/Layout/redist/targets/GeneratePackagePruneData.targets) creates an interesting new dependency. This is a bit problematic as it shows up during rebranding - not the ideal time.

cc @dotnet/source-build

MichaelSimons:
I can make an attempt at adding the targeting packs needed by source-build.

ViktorHofer:
Following this approach also means that source-build will need to store targeting packs for every .NET version >= 10. So imagine we are rebranding to a far ahead future version like .NET 18. For that, the .NET 10 up to .NET 17 targeting packs would still need to be available via SBRP. I think we really want to avoid that.

A harvesting approach from N-1 might be a better fit here. cc @ericstj

MichaelSimons:
> Following this approach also means that source-build will need to store targeting packs for every .NET version >= 10. So imagine we are rebranding to a far ahead future version like .NET 18. For that, the .NET 10 up to .NET 17 targeting packs would still need to be available via SBRP. I think we really want to avoid that.

Yes, completely agree.

mmitche:
> > Following this approach also means that source-build will need to store targeting packs for every .NET version >= 10. So imagine we are rebranding to a far ahead future version like .NET 18. For that, the .NET 10 up to .NET 17 targeting packs would still need to be available via SBRP. I think we really want to avoid that.
> 
> Yes, completely agree.

Double agree

dsplaisted:
> This is a bit problematic as it shows up during rebranding - not the ideal time.

The SDK needs prune package data from the current SDK and previous SDKs.  When the current version changes, then the old "current" version is added to the list of previous SDKs.  For source build, I'm not sure if there's any way around having this be a manual process that will have to be done when updating the current version.  Either we could add the previous package versions to the prebuilts, or we could check in the new data (is that what you mean by harvesting?).  Any other ideas?

@ViktorHofer @ericstj @mmitche @MichaelSimons

ViktorHofer:
Couple of questions:
1. I see that https://github.com/dotnet/dotnet/blob/bb6bd81348665b4f664f0148d1e32f74fadc1ca2/src/sdk/src/Layout/redist/targets/GeneratePackagePruneData.targets#L23 only ever downloads the `X.0.0` version of the targeting packs even though the PackageOverrides.txt file changes during servicing. Is that a bug or doesn't it matter because the pruning algorithm doesn't consider the patch versions from that data source?
2. If we already only ever restore the `X.0.0` version, why don't we just check the data in directly for already released targeting pack versions? That would unblock source-build and avoid the strong dependency on previous targeting packs during the product composition.

> Is that what you mean by harvesting?

Here I meant copying the `PackagePruneData` folder from the N-1 SDK that is used to build and only restoring targeting packs that are missing from the data. What this implies though is to fix one of those package prune entries for past frameworks you would need to service the past SDKs first. This might be quite impractical. I would probably go with checking the data in, instead.

dsplaisted:
> Is that a bug or doesn't it matter because the pruning algorithm doesn't consider the patch versions from that data source?

For .NET Core we ignore the patch versions for pruning (by setting the patch version to 32767 in the pruning data we return).  So we only need the .0 targeting packs, and whatever manual process we need to do to make the data available to source build only needs to happen once a year, not once a month.

> If we already only ever restore the X.0.0 version, why don't we just check the data in directly for already released targeting pack versions? That would unblock source-build and avoid the strong dependency on previous targeting packs during the product composition.

We could do that.  When I was originally implementing this, I tried to come up with a system that would require no manual changes each major or minor release.  However, it looks like that won't work for source build.  So we could either add the .0 targeting packs to the prebuilts, or we could check in the data, whichever seems best / easier.

ViktorHofer:
> So we could either add the .0 targeting packs to the prebuilts, or we could check in the data, whichever seems best / easier.

I think we need to check in the data directly into the sdk source, otherwise we would accumulate that large number of pre-builts which is not viable.

dsplaisted:
> > So we could either add the .0 targeting packs to the prebuilts, or we could check in the data, whichever seems best / easier.
> 
> I think we need to check in the data directly into the sdk source, otherwise we would accumulate that large number of pre-builts which is not viable.

I get that it grows over time, but I think it's 3 targeting packs per year.  Is that really an issue?

ViktorHofer:
We can't have an unbounded growth of prebuilts in source-build. Targeting packs get deleted from SBRP when all consumers upgraded to newer TFMs. Checking them into SBRP is very likely not what you want. You would either need to check in the entire targeting packs (all content) or create a special transport package which you then would need to restore in sdk but just for source-build. Given that, I believe it makes sense and is lower cost to do this in sdk and not special case for source-build.

ViktorHofer:
@dsplaisted as this blocking the .NET 11 re-branding effort, can you please take a look and work on this in sdk?

I think we could move all the existing data (currently .cs files) to a PackagePruneData folder in redist.csproj as .json/.txt which would then get copied to the sdk output directory folder during the build.

When we rebrand and are missing data files for the N-1 TFM (i.e. net10.0 in this case), we would get an error and then include that data in the checked-in files. This would be manual by invoking a target that does that for us. Does that make sense? I didn't see anything in the existing .cs files that would prevent moving them into static data files.

dsplaisted:
@ViktorHofer @mmitche I've created a PR in the SDK which creates a process for checking in the prune package data: https://github.com/dotnet/sdk/pull/51992

mmitche:
@dsplaisted That change will require that https://github.com/dotnet/dotnet/pull/3123 go in. May just want to make the change here.

dsplaisted:
@mmitche Is there a process to "cherry-pick" changes from a repo into the VMR?  Or is it just to manually apply the changes?

dsplaisted:
Ugh, I accidentally clicked the close button on this, and that caused a bot to move it to Done on some board.

am11:
@dsplaisted, we can apply upstream PR patches with:

```sh
$ curl -sSL https://github.com/dotnet/sdk/pull/51992.diff | git apply --directory=src/sdk -
```

mmitche:
> @mmitche Is there a process to "cherry-pick" changes from a repo into the VMR? Or is it just to manually apply the changes?

darc vmr cherry-pick :)

dsplaisted:
I've cherry-picked the changes from https://github.com/dotnet/sdk/pull/51992 into this PR and added the prune package data.

It required a little bit of hacking since I didn't want to try to build the full VMR locally, just the SDK repo.  I hardcoded the MaxNetVersion to 11 in GeneratePackagePruneData.targets, and I had to disable some template packaging which was failing.

ViktorHofer:
scenario-tests grrrrr

ViktorHofer:
> D:\a\_work\1\s\artifacts\obj\extracted-dotnet-sdk\sdk\11.0.100-ci\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(66,5): error NETSDK1226: Prune Package data not found .NETCoreApp 10.0 Microsoft.AspNetCore.App.  To ignore this error, set the AllowMissingPrunePackageData to true. [D:\a\_work\1\s\src\scenario-tests\artifacts\obj\generatedtests\SdkTemplateTest_Mvc_CSharp\SdkTemplateTest_Mvc_CSharp.csproj] [D:\a\_work\1\s\src\scenario-tests\src\Microsoft.DotNet.ScenarioTests.SdkTemplateTests\Microsoft.DotNet.ScenarioTests.SdkTemplateTests.csproj] [D:\a\_work\1\s\repo-projects\scenario-tests.proj]

Looks like the aspnetcore net10.0 pruning data isn't there or can't be resolved.

image

It is there though:

image

ViktorHofer:
Looks like this hardcode is wrong: https://github.com/dotnet/sdk/blob/0d31ba892465268a030b3563df0bddef39855f6e/src/Tasks/Microsoft.NET.Build.Tasks/GetPackagesToPrune.cs#L247

ViktorHofer (approved):

dsplaisted:
🎉

PR #2289, opened 2025-09-05
Create roslyn apphosts

jjonescz:
Part of https://github.com/dotnet/roslyn/issues/80151.
Based on https://github.com/dotnet/dotnet/pull/2284.

### DIFF
--- a/src/sdk/eng/Signing.props
+++ b/src/sdk/eng/Signing.props
@@ -86,12 +86,7 @@
     <FileSignInfo Include="StreamJsonRpc.dll" CertificateName="MicrosoftSHA2" />
 
     <!-- Roslyn apphosts -->
-    <FileSignInfo Include="csc" CertificateName="$(DotNetCertificateName)" />
-    <FileSignInfo Include="vbc" CertificateName="$(DotNetCertificateName)" />
-    <FileSignInfo Include="VBCSCompiler" CertificateName="$(DotNetCertificateName)" />
-    <FileSignInfo Include="csc.exe" CertificateName="$(DotNetCertificateName)" />
-    <FileSignInfo Include="vbc.exe" CertificateName="$(DotNetCertificateName)" />
-    <FileSignInfo Include="VBCSCompiler.exe" CertificateName="$(DotNetCertificateName)" />
+    <FileSignInfo Condition="'$(TargetsOSX)' == 'true'" Include="csc;vbc;VBCSCompiler" CertificateName="MacDeveloperHarden" />
   </ItemGroup>
 
   <!-- Filter out any test packages from ItemsToSign -->
--- a/src/sdk/eng/Signing.props
+++ b/src/sdk/eng/Signing.props
@@ -86,7 +86,7 @@
     <FileSignInfo Include="StreamJsonRpc.dll" CertificateName="MicrosoftSHA2" />
 
     <!-- Roslyn apphosts -->
-    <FileSignInfo Condition="'$(TargetsOSX)' == 'true'" Include="csc;vbc;VBCSCompiler" CertificateName="MacDeveloperHarden" />
+    <FileSignInfo Condition="'$(OSName)' == 'osx'" Include="csc;vbc;VBCSCompiler" CertificateName="MacDeveloperHarden" />
   </ItemGroup>
 
   <!-- Filter out any test packages from ItemsToSign -->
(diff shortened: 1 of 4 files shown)


src/sdk/src/Layout/redist/targets/Crossgen.targets
```diff
@@ -197,6 +197,7 @@
 
     <Exec Command="find $(InstallerOutputDirectory) -type d -exec chmod 755 {} \;" Condition="'$(OSName)' != 'win'" />
     <Exec Command="find $(InstallerOutputDirectory) -type f -exec chmod 644 {} \;" Condition="'$(OSName)' != 'win'" />
+    <Exec Command="chmod 755 $(RedistInstallerLayoutPath)sdk/$(Version)/Roslyn/bincore/%(_RoslynAppHost.Filename)$(ExeExtension)" Condition="'$(OSName)' != 'win'" />
```
akoeplinger: shouldn't this use InstallerOutputDirectory as well? and I'm not sure we want to rely on _RoslynAppHost being set by GenerateLayout.targets here

src/sdk/src/Layout/redist/targets/GenerateLayout.targets
```diff
@@ -59,6 +59,17 @@
...
+    <ItemGroup>
+      <_RoslynAppHost Include="$(OutputPath)Roslyn\bincore\csc.dll" />
+      <_RoslynAppHost Include="$(OutputPath)Roslyn\bincore\vbc.dll" />
+      <_RoslynAppHost Include="$(OutputPath)Roslyn\bincore\VBCSCompiler.dll" />
+    </ItemGroup>
+    <CreateAppHost AppHostSourcePath="$(NuGetPackageRoot)\microsoft.netcore.app.host.$(SharedFrameworkRid)\$(MicrosoftNETCoreAppHostPackageVersion)\runtimes\$(SharedFrameworkRid)\native\apphost$(ExeExtension)"
+                   AppHostDestinationPath="%(_RoslynAppHost.RootDir)%(_RoslynAppHost.Directory)%(_RoslynAppHost.Filename)$(ExeExtension)"
+                   AppBinaryName="%(_RoslynAppHost.Filename)%(_RoslynAppHost.Extension)"
+                   IntermediateAssembly="%(_RoslynAppHost.FullPath)"
+                   EnableMacOSCodeSign="$(RuntimeIdentifier.StartsWith('osx'))" />
```
akoeplinger, line 73: I think we'll need to add the apphost binaries to eng/Signing.props to make sure they're correctly signed

src/sdk/src/Layout/redist/targets/Crossgen.targets
jjonescz: > shouldn't this use InstallerOutputDirectory as well?

You're right, I have no idea why I used RedistInstallerLayoutPath. Thanks.

> I'm not sure we want to rely on _RoslynAppHost being set by GenerateLayout.targets here

Good point. I can move the `_RoslynAppHost` item group outside the GenerateLayout target so it's guaranteed to be visible across all targets.

src/sdk/eng/Signing.props
```diff
@@ -84,6 +84,14 @@
   
   <ItemGroup>
     <FileSignInfo Include="StreamJsonRpc.dll" CertificateName="MicrosoftSHA2" />
+
+    <!-- Roslyn apphosts -->
+    <FileSignInfo Include="csc" CertificateName="$(DotNetCertificateName)" />
+    <FileSignInfo Include="vbc" CertificateName="$(DotNetCertificateName)" />
+    <FileSignInfo Include="VBCSCompiler" CertificateName="$(DotNetCertificateName)" />
```
akoeplinger: Mac needs a different cert: https://github.com/dotnet/runtime/blob/0e3562e97c6db531f26a2ffe3e8084cf67ba8a93/eng/Signing.props#L42-L46

you can just update this one since we don't sign Linux executables

src/sdk/eng/Signing.props
```diff
@@ -84,6 +84,14 @@
...
+    <FileSignInfo Include="csc.exe" CertificateName="$(DotNetCertificateName)" />
+    <FileSignInfo Include="vbc.exe" CertificateName="$(DotNetCertificateName)" />
+    <FileSignInfo Include="VBCSCompiler.exe" CertificateName="$(DotNetCertificateName)" />
```
akoeplinger: .exe should be signed by default due to arcade's defaults, I think you can remove this.

akoeplinger (changes requested):

src/sdk/eng/Signing.props
akoeplinger: hm I think it'd make more sense to move this to Roslyn's Signing.props

jjonescz:
Reopened at https://github.com/dotnet/dotnet/pull/2425.

PR #2295, opened 2025-09-05
Ensure that signed SB artifacts are published

ellahathaway:
The signed SB artifacts were not being published via Arcade. This PR resolves that problem with the following changes:

- Adds checks to ensure that SB artifacts are always published when signing is enabled
- Moves the signed build leg from the CentOS online leg to the CentOS offline leg which is already publishing the artifacts via Arcade.

[Test build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2787275&view=results)


eng/pipelines/templates/stages/source-build-stages.yml
```diff
@@ -153,14 +153,14 @@ stages:
           - buildNameFormat: 'SB_{0}_Offline_MsftSdk'
             distro: centos
             targetArchitecture: x64
-            buildFromArchive: true             # ✅
+            buildFromArchive: false            # 🚫
```
ellahathaway, line 156: In order to sign, we need the source artifacts. We cannot generate the source artifacts and build from archive. We should determine if we want to build from archive in a different leg.

MichaelSimons: We only need one leg that covers this scenario just to ensure it does not regress.  I see there are three others that do so it is fine to remove it from this leg.

ellahathaway:
Test pipeline run was successful. Signing and signing validation were green, and the merged manifest includes the source artifacts, PSB tarball, and SB SDK tarball as `DotNetReleaseShipping` artifacts. This PR is now ready for review :)

eng/Publishing.props
```diff
@@ -9,6 +9,7 @@
     <ArtifactsStagingDir Condition="'$(ArtifactsStagingDir)' == ''">$(ArtifactsDir)/staging/</ArtifactsStagingDir>
     <ArtifactsPublishStagingDir>$(ArtifactsStagingDir)/artifacts</ArtifactsPublishStagingDir>
     <SourceBuiltAssetsDir>$(ArtifactsPublishStagingDir)/assets/$(Configuration)</SourceBuiltAssetsDir>
+    <SourceBuiltPublishAssetsDir>$(SourceBuiltAssetsDir)/source-build</SourceBuiltPublishAssetsDir>
```
MichaelSimons, line 12: Is this path used?  It looks like the only place it is used, grab the last segment.  Should this be defined differently?

ellahathaway: Yes, it is used in [`sign-source-built-artifacts.proj`](https://github.com/dotnet/dotnet/blob/58101cbac6fc8379eabd92b586e38fdaba18d614/eng/sign-source-built-artifacts.proj#L44-L53). I'd rather not move it per the note [here](https://github.com/dotnet/dotnet/blob/58101cbac6fc8379eabd92b586e38fdaba18d614/eng/sign-source-built-artifacts.proj#L29-L30).

PR #2338, opened 2025-09-09
Retry on all errors for PSB download

mthalman:
A network error recently caused the download of the PSB artifacts to fail in the pipeline. The use of `--retry` doesn't cause retries on all types of errors. The use of `--retry-all-errors` will make it more robust.

### DIFF
--- a/prep-source-build.sh
+++ b/prep-source-build.sh
@@ -141,30 +141,6 @@ function GetXmlPropertyValue {
   echo "$value"
 }
 
-# Helper to download a file with retries
-function DownloadWithRetries {
-  local url="$1"
-  local targetDir="$2"
-  (
-    cd "$targetDir" &&
-    for i in {1..5}; do
-      if curl -fL --retry 5 -O "$url"; then
-        return 0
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -424,7 +424,7 @@ jobs:
           downloadDir="$(sourcesPath)/prereqs/packages/archive/"
 
           echo "Downloading source-built SDK from $archiveUrl..."
-          (cd "$downloadDir" && curl --retry 5 -O "$archiveUrl")
+          (cd "$downloadDir" && curl --retry 5 --retry-all-errors -O "$archiveUrl")
```
MichaelSimons, line 427: prep-source-build.sh has the same logic.  Should both be doing the same?

akoeplinger:
This doesn't work on older curl versions and breaks the official build on AlmaLinux8, see https://github.com/dotnet/dotnet/pull/1185

Reverting.

PR #2297, opened 2025-09-06
Add Winforms PackageOverrides to WindowsDesktop

ericstj:
Fixes https://github.com/dotnet/windowsdesktop/issues/5244

PR #2357, opened 2025-09-10
[release/10.0.1xx] Add Winforms PackageOverrides to WindowsDesktop

ericstj:
Backporting https://github.com/dotnet/dotnet/pull/2297

## Customer Impact
-	Internally found
System.Drawing.Common is part of WindowsDesktop shared framework but is not trimmed.  The package will be downloaded even though it will not be used on .NET 10.0

## Regression
No, missing functionality in a feature new to 10.0

## Testing
Discovered this as part of comprehensive testing of https://github.com/dotnet/dotnet/pull/2254
Manually verified that this one addition fixes the remaining missing packages from pruning data for net10.

## Risk
Low.  Functionally this will not change anything since the package was removed by conflict resolution in build.  
This change adds to pruning.  There is a small chance that folks who don’t multitarget and have a direct reference to the package on net10.0 will see a pruning warning, but this is an actionable warning and consistent with all other packages that are part of shared framework (and introduced in .NET 10).  
For those moving from a previous framework this will not be a new warning but will be part of the rest of the warnings they see.


akoeplinger:
there's a merge conflict with the V.D.xml

ericstj:
Yeah, that dang thing gets conflicts with every single commit.

PR #2359, opened 2025-09-10
System.CommandLine - Change Terminating API to be callable from C# 7

jonsequitur:
## Description
Fixes dotnet/command-line-api#2213 - [`init` accessor of CliAction.Terminating cannot be called in C# 7](https://github.com/dotnet/command-line-api/issues/2213)

## Impact
The `init` accessor cannot be directly called in C# 7; it requires at least C# 9.  That makes the property difficult to initialize in these cases:
* When targeting .NET Framework, where [Microsoft does not support C# 8](https://github.com/dotnet/docs/blob/c802a0c56bccdc87908365c0788b3c5ecacea2b3/docs/csharp/language-reference/configure-language-version.md?plain=1#L12).
* When the application must remain buildable in Visual Studio 2017 ([whose support ends in April 2027](https://learn.microsoft.com/lifecycle/products/visual-studio-2017)).

This issue was originally reported in 2023, but it had fallen off our radar. A customer commented on the issue last week bringing it back to our attention.

## Details
This fix updates the `CommandLineAction.Terminating` API to be virtual rather than relying on the `protected init` language feature. This is a breaking change, and we are therefore applying to change via the dotnet/dotnet repo so the compensating changes to the SDK can be applied at the same time.

## Breaking Change?

**Yes** - but usage of this property is relatively low (SDK is the only impacted component in our product), and we want to get the API correct before stable release. We confirmed that neither the NuGet Client nor the new Microsoft/MCP repos utilize this API. NuGet does not have any subclasses of `CommandLineAction`.

* [NuGet.Client - .SetAction](https://github.com/search?q=repo%3ANuGet%2FNuGet.Client%20.SetAction&type=code) | [NuGet.Client - Terminating](https://github.com/search?q=repo%3ANuGet%2FNuGet.Client+Terminating+language%3AC%23&type=code&l=C%23)
* [Microsoft/MCP - .SetAction](https://github.com/search?q=repo%3Amicrosoft%2Fmcp+.SetAction&type=code) | [Microsoft/MCP - Terminating](https://github.com/search?q=repo%3Amicrosoft%2Fmcp+Terminating&type=code)

## Testing

Automated tests continue to preserve the existing functionality - this is mostly an API surface area change.

### DIFF
--- a/src/sdk/src/Cli/dotnet/Commands/Workload/WorkloadCommandParser.cs
+++ b/src/sdk/src/Cli/dotnet/Commands/Workload/WorkloadCommandParser.cs
@@ -174,10 +174,7 @@ internal static class WorkloadCommandParser
 
     private class ShowWorkloadsInfoAction : SynchronousCommandLineAction
     {
-        public ShowWorkloadsInfoAction()
-        {
-            Terminating = true;
-        }
+        public override bool Terminating => true;
 
         public override int Invoke(ParseResult parseResult)
         {
@@ -189,10 +186,7 @@ internal static class WorkloadCommandParser
 
     private class ShowWorkloadsVersionOption : SynchronousCommandLineAction
     {
-        public ShowWorkloadsVersionOption()
-        {
-            Terminating = true;
-        }
+        public override bool Terminating => true;
 
         public override int Invoke(ParseResult parseResult)
         {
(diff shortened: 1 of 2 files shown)


artl93:
Are there any compensating changes needed elsewhere in the dotnet repo? (Based on this change, I assume not.)

jeffhandley:
> LGTM, thank you @jonsequitur ! Is it the only `init` public property we have?

This is the only `init` property we had.

> Are there any compensating changes needed elsewhere in the dotnet repo?

The SDK changes included here are the only occurrences within dotnet/dotnet or that could be found across the dotnet org.

PR #2361, opened 2025-09-10
Add feature flag for binskim scanning of packages

ericstj:
This should enable scanning of published outputs - including archives and packages, instead of the current hack which is scanning our artifacts folder.


eng/pipelines/official.yml
```diff
@@ -95,13 +96,6 @@ extends:
         tsaEnabled: true
       binskim:
         enabled: true
-        # This may need to be updated to exclude files we don't want scanned.
-        # This pattern was pulled from dotnet/sdk, with the ESRPClient added.
-        # When binskim supports scanning of archive contents, this can be removed altogether.
-        # This is of the form: <pattern>;<pattern>
-        # +:f| - Include glob
-        # -:f| - Exclude glob
-        analyzeTargetGlob: +:f|eng\**\*.props;+:f|**\artifacts\bin\**\*.dll;+:f|**\artifacts\bin\**\*.exe;-:f|**\artifacts\bin\**\msdia140.dll;-:f|**\artifacts\bin\**\pgort140.dll;-:f|**\artifacts\bin\*Tests\**;-:f|**\Microsoft.NET.Runtime.Emscripten**\tools\**;-:f|**\CodeCoverage\**;-:f|**\artifacts\bin\**\capstone.dll;-:f|**\Microsoft.EsrpClient\**;-:f|**\artifacts\bin\**\SuperFileCheck\**\llvm-mca.exe;-:f|**\artifacts\bin\**\SuperFileCheck\**\FileCheck.exe;
```
akoeplinger, line 104: what about the files we excluded from binskim?

ericstj: Presumably those are non-shipping files that won't be captured with the real logic that only scans our published shipping artifacts.  If they are shipping files, then we shouldn't be excluding them and the issues are real and folks who own those will need to get exceptions.

akoeplinger: I was thinking of https://github.com/dotnet/dotnet/pull/2339 but I'm not sure if the new binskim setup will still catch that, i.e. is it configured to _only_ scan the shipping packages?

eng/pipelines/official.yml
```diff
@@ -52,6 +52,12 @@ variables:
 - name: isScheduleTrigger
   value: ${{ eq(variables['Build.Reason'], 'Schedule') }}
 
+- name: GDN_EXTRACT_TOOLS
+  value: 'binskim,bandit,roslynanalyzers'
+ 
+- name: GDN_EXTRACT_FILTER
+  value: 'f|**/*.zip;f|**/*.nupkg;f|**/*.vsix;f|**/*.cspkg;f|**/*.sfpkg;f|**/*.package'
```
akoeplinger, line 59: can we make sure this only scans shipping artifacts?

eng/pipelines/official.yml
ericstj: All the 1es injected stuff is meant to run on only the artifacts we publish (instead of everything on the machine).  So if we publish it that will run.

If we're redistributing it in our production bits, there may be a different way to exclude it.

There are some things we classify as "nonshipping" but still publish - a lot of the arcade infra, for example.  This stuff will still be scanned, and it probably should since it will be used in production build pipelines.

eng/pipelines/official.yml
ericstj: We don't even want to have to set this one but were instructed to do so.  I tried to omit it, but the feature doesn't work when omitted.
https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-mohanb/security-integration/guardian-wiki/microsoft-guardian/general/features/target-extraction
We are setting it as a filter to apply to the files already selected to be passed.  My understanding is that this does not include any new files, but identifies which files of those included that should be considered for extraction.

@akoeplinger If you have any files you are worried about spot check the results in the test builds I've been doing to see if they are scanned or not.

ericstj:
Marked this as NO-MERGE.  We probably don't want this all set at the top-level of the pipeline if it can be avoided.  If possible we should have arcade configure this for all dotnet repos.

ericstj:
Previous build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2796043&view=results (failing due to R2R composite images)

Ported @jkoritzinsky change https://github.com/dotnet/aspnetcore/pull/63724
https://dev.azure.com/dnceng/internal/_build/results?buildId=2796723&view=results

ericstj:
@mmitche do you want to take this as is, or would you prefer changes go to yml templates so that they might be reused by other repos?

mmitche:
As-is. There's not an obvious way this to get that binSkimScanAllExtensions bit into every arcade pipeline without some general redesign of the arcade templates. The variables would work though, I think. In addition, I think we should let this ride for a little while before attempting to roll it out to everyone

PR #2371, opened 2025-09-11
Opt out of Guardian Roslyn analyzers

akoeplinger:
We hit a bug: "ErrorLog property didn't match expected pattern"


akoeplinger:
Fast merge to unblock the official build.

PR #2375, opened 2025-09-11
Use SDK's .version file in PSB archive

NikolaMilosavljevic:
Fixes https://github.com/dotnet/source-build/issues/5068

- Instead of creating source-build version of `.version` file, we are now using SDK's `.version` file, as is.
- Tests are updated to ensure the `.version` files are identical and that the `.version` file structure was not modified.

### DIFF
--- a/test/Microsoft.DotNet.SourceBuild.Tests/SourceBuiltArtifactsTests.cs
+++ b/test/Microsoft.DotNet.SourceBuild.Tests/SourceBuiltArtifactsTests.cs
@@ -46,10 +46,8 @@ public class SourceBuiltArtifactsTests : SdkTests
             }
 
             // Verify that .version file is the same as the SDK's .version file.
-
             string sbVersionFileContent = File.ReadAllText(sbVersionFilePath);
 
-            // Find the expected SDK version by getting it from the source built SDK
             DirectoryInfo sdkDir = new DirectoryInfo(Path.Combine(Config.DotNetDirectory, "sdk"));
             string sdkVersionPath = sdkDir.GetFiles(".version", SearchOption.AllDirectories).Single().FullName;
             string sdkVersionFileContent = File.ReadAllText(Path.Combine(outputDir, sdkVersionPath));


test/Microsoft.DotNet.SourceBuild.Tests/SourceBuiltArtifactsTests.cs
```diff
@@ -30,9 +30,13 @@ public void VerifyVersionFile()
...
             Utilities.ExtractTarball(Config.SourceBuiltArtifactsPath, outputDir, ".version");
+            string sbVersionFilePath = Path.Combine(outputDir, ".version");
 
-            string[] versionLines = File.ReadAllLines(Path.Combine(outputDir, ".version"));
-            Assert.Equal(2, versionLines.Length);
+            string[] versionLines = File.ReadAllLines(sbVersionFilePath);
+
+            // Verify the expected number of lines in .version file.
+            // Ensures that the test fails if SDK changes the .version file format.
+            Assert.Equal(5, versionLines.Length);
```
mthalman: Why are we validating the format of the .version file in the SDK? That seems outside the scope of this test.

NikolaMilosavljevic: In the past we were validating the number of lines in SB's `.version` file.

Now that we copy/reuse the SDK's `.version` file, we can either:
1) assume that there are no changes in that file going forward that would regress the user experience - i.e. ability to obtain the unstable version, or
2) do some simple checks of the SDK file (now SB file too) - this is where my check for exact number of lines come in.

I went with option 2, but if others are OK I will remove this check.

@MichaelSimons

MichaelSimons: I think a smoke test makes sense as we need some way to ensure that the file exists and an contains some form of the correct content.  Maybe simply comparing the content against the SDK version would be sufficient (e.g compare shas)...

mthalman: I don't think we need to do anything more than verify that the contents match between two files. This test is for validating the behavior of the VMR orchestrator and that's all that it's doing. It's the SDK which is defining the content of that file. If we want a test in the sdk repo that verifies the format, that's fine. But I do not think that the testing of that format should be duplicated.

NikolaMilosavljevic: OK - will remove the lines-count test.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/2375/commits/7e97eee06529a184b6c503c9b72ab61080bfc830

PR #2403, opened 2025-09-12
Download shared components outside of prep script

mthalman:
Fixes dotnet/source-build#5332

Moves the logic for downloading the shared component artifacts out of the prep-source-build script so that it can be explicitly initiated by the VMR pipeline only. This avoids the scenario described by dotnet/source-build#5332.

To implement this, I refactored the prep script so that the archive downloading is in a separate standalone script. This allows the prep script and the pipeline logic to share a common implementation for this download logic.


eng/download-source-built-archive.sh
```diff
@@ -0,0 +1,108 @@
...
+
+  local archiveUrl
+  local artifactsBaseFileName="Private.SourceBuilt.Artifacts"
+  local prebuiltsBaseFileName="Private.SourceBuilt.Prebuilts"
+  local defaultArtifactsRid='centos.10-x64'
+
+  if [[ "$propertyName" == "MicrosoftNETSdkVersion" ]]; then
+    archiveUrl="https://ci.dot.net/public/source-build/$artifactsBaseFileName.$archiveVersion.$artifactsRid.tar.gz"
+  elif [[ "$propertyName" == *Prebuilts* ]]; then
+    archiveUrl="https://builds.dotnet.microsoft.com/source-built-artifacts/assets/$prebuiltsBaseFileName.$archiveVersion.$defaultArtifactsRid.tar.gz"
```
MichaelSimons, line 80: Will this need to account for internal builds?  Should/can that be accounted for now?

mthalman: It will, yes. I will log a follow-up issue. There's not really a way to test this right now.

MichaelSimons (approved):

eng/download-source-built-archive.sh
mthalman: https://github.com/dotnet/source-build/issues/5342

PR #2425, opened 2025-09-15
Create roslyn apphosts

jjonescz:
Part of https://github.com/dotnet/roslyn/issues/80151.
Continuation of https://github.com/dotnet/dotnet/pull/2289.
Official build validation run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2793026&view=results

PR #2464, opened 2025-09-17
Fix macOS apphost sign condition

jjonescz:
Follow up on https://github.com/dotnet/dotnet/pull/2425.
Related to https://github.com/dotnet/roslyn/issues/80151.

### DIFF
--- a/src/sdk/src/Layout/redist/targets/GenerateLayout.targets
+++ b/src/sdk/src/Layout/redist/targets/GenerateLayout.targets
@@ -64,7 +64,7 @@
                    AppHostDestinationPath="%(_RoslynAppHost.RootDir)%(_RoslynAppHost.Directory)%(_RoslynAppHost.Filename)$(ExeExtension)"
                    AppBinaryName="%(_RoslynAppHost.Filename)%(_RoslynAppHost.Extension)"
                    IntermediateAssembly="%(_RoslynAppHost.FullPath)"
-                   EnableMacOSCodeSign="$(RuntimeIdentifier.StartsWith('osx'))" />
+                   EnableMacOSCodeSign="$(SharedFrameworkRid.StartsWith('osx'))" />
   </Target>
 
   <Target Name="PublishNETAnalyzers">


src/sdk/src/Layout/redist/targets/GenerateLayout.targets
```diff
@@ -64,7 +64,7 @@
                    AppHostDestinationPath="%(_RoslynAppHost.RootDir)%(_RoslynAppHost.Directory)%(_RoslynAppHost.Filename)$(ExeExtension)"
                    AppBinaryName="%(_RoslynAppHost.Filename)%(_RoslynAppHost.Extension)"
                    IntermediateAssembly="%(_RoslynAppHost.FullPath)"
-                   EnableMacOSCodeSign="$(RuntimeIdentifier.StartsWith('osx'))" />
+                   EnableMacOSCodeSign="$(SharedFrameworkRid.StartsWith('osx'))" />
```
akoeplinger, line 67: I think I'd prefer this to align with how similar checks are done in other parts of the sdk

```suggestion
                   EnableMacOSCodeSign="$([MSBuild]::IsOSPlatform('OSX'))" />
```

jjonescz: Hm, I chose `SharedFrameworkRid` because that's what is used to find the apphost template, i.e., we set `EnableMacOSCodeSign=true` if and only if we are using macOS apphost, consistently. Similarly, in theory what if I'm cross-building the sdk on linux machine but want to target macOS? I guess that might not be possible anyway (i.e., `EnableMacOSCodeSign` might not work on non-macOS machines). So I'm fine with accepting your suggestion, just wanted to state my reasoning.

akoeplinger: ah yeah given we're already using SharedFrameworkRid here like you said I'm fine with this

akoeplinger:
~~port to main?~~ should be done together with https://github.com/dotnet/dotnet/pull/2425

Issue #2009, opened 2025-07-20
Microsoft.Build.NoTargets missing in prep-source-build on 10.0.0-preview.6


corngood:
### Describe the Bug

When building 10.0.0-preview.6 in a sandbox without networking, I get these errors:

```
/build/dotnet-10.0.0-preview.6.25358.103/eng/init-detect-binaries.proj : error : Could not resolve SDK "Microsoft.Build.NoTargets". Exactly one of the probing messages below indicates why we could not resolve the SDK. Investigate and resolve that message to correctly specify the SDK.
/build/dotnet-10.0.0-preview.6.25358.103/eng/init-detect-binaries.proj : error :   SDK resolver "Microsoft.DotNet.MSBuildWorkloadSdkResolver" returned null.
/build/dotnet-10.0.0-preview.6.25358.103/eng/init-detect-binaries.proj : error :   Unable to find package Microsoft.Build.NoTargets. No packages exist with this id in source(s): dotnet-eng, dotnet-libraries, dotnet-public
/build/dotnet-10.0.0-preview.6.25358.103/eng/init-detect-binaries.proj : warning : Unable to load the service index for source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json.
/build/dotnet-10.0.0-preview.6.25358.103/eng/init-detect-binaries.proj : warning : Unable to load the service index for source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json.
/build/dotnet-10.0.0-preview.6.25358.103/eng/init-detect-binaries.proj : warning : Unable to load the service index for source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json.
/build/dotnet-10.0.0-preview.6.25358.103/eng/init-detect-binaries.proj : error MSB4236: The SDK 'Microsoft.Build.NoTargets' specified could not be found.
```

This worked okay on preview 5.

### Steps to Reproduce

Build `dotnetCorePackages.dotnet_10.stage0.vmr` from https://github.com/NixOS/nixpkgs/pull/426786

### Other Information

I believe this is something to do with init-detect-binaries.proj being called from prep-source-build.sh without the bootstrap packages being available.

corngood:
I can work around this by adding both 

[...]

PR #2469, opened 2025-09-17
Add support to run prep binary removal offline

MichaelSimons:
Fixes https://github.com/dotnet/dotnet/issues/2009

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -115,7 +115,7 @@ releaseManifest=''
 sourceRepository=''
 sourceVersion=''
 CUSTOM_PACKAGES_DIR=''
-CUSTOM_SDK_DIR=''
+customSdkDir=''
 packagesDir="$scriptroot/prereqs/packages/"
 packagesArchiveDir="${packagesDir}archive/"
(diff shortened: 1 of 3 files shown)


mthalman (approved):

build.sh
```diff
@@ -115,7 +115,7 @@ releaseManifest=''
 sourceRepository=''
 sourceVersion=''
 CUSTOM_PACKAGES_DIR=''
-CUSTOM_SDK_DIR=''
+customSdkDir=''
```
ellahathaway, line 118: Nit: I prefer `CUSTOM_SDK_DIR` over `customSdkDir` because it matches the style of `CUSTOM_PACKAGES_DIR`. If we change the style of the sdk dir variable, we should change the style of the packages dir variable too.

MichaelSimons: The upper casing of these two variable do not match anything else which is why I changed the sdkDir since I was refactoring all usages of it.  I updated CUSTOM_PACKAGES_DIR as well in https://github.com/dotnet/dotnet/pull/2469/commits/b00333b3f74b4258fd5a58c5d69ac09feaa2243d.

PR #2488, opened 2025-09-18
Add telemetry for the task host and task factory usage in MSBuild.

AR-May:
Fixes https://github.com/dotnet/msbuild/issues/12507

Changes:
In MSBuild: Adding MSBuild project level telemetry collecting task host and task factories usage 
In SDK: Adding new telemetry events processing to SDK

Telemetry events are sent per project build in MSBuild and then aggregated on SDK side.

Testing:
SDK part - unit tests
MSBuild part - manual testing, compared reported numbers to what is available in binlog, they seem to be correct


AR-May:
After internal conversation decided that we will aggregate the new events on the SDK side and fire only one event per build.

baronfel:
cc @dsplaisted because of SDK telemetry

surayya-MS (approved):

baronfel (approved):
This LGTM from a functional perspective - but it should be targeted against release/10.0.1xx.

PR #2513, opened 2025-09-19
Bump timeout for internal CodeQL and public OSX runs

akoeplinger:
I noticed that internally when CodeQL is active the AzureLinux jobs timed out after 4hours, extend the current increased timeout from Windows to all jobs.

Also increase the timeout for public OSX runs since we use older macOS hardware there which is slower and finishes in just under 4 hours (we only build OSX in dotnet-unified-build-full so we don't see it all the time)

### DIFF
--- a/eng/pipelines/templates/jobs/vmr-build.yml
+++ b/eng/pipelines/templates/jobs/vmr-build.yml
@@ -327,8 +327,10 @@ jobs:
     timeoutInMinutes: 720
   ## Currently, CodeQL slows the build down too much
   ## https://github.com/dotnet/source-build/issues/4276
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -603,13 +605,13 @@ jobs:
...
-          # Update the owner of the staging directory to the current user
-          sudo chown -R $(whoami) $(Build.ArtifactStagingDirectory)
-        displayName: Update owner of artifacts staging directory
-
+    - ${{ if ne(parameters.runOnline, 'True' )}}:
+      - script: |
+          set -ex
+          # Update the owner of the staging directory to the current user
+          sudo chown -R $(whoami) $(Build.ArtifactStagingDirectory)
+        displayName: Update owner of artifacts staging directory
```
akoeplinger, line 613: this is unrelated but GitHub told me there are mixed line endings in the file and it would fix it

mmitche (approved):

PR #2541, opened 2025-09-22
[release/10.0.1xx] arcade: Pass Platform variable as WiX input

PranavSenthilnathan:
The Platform variable was not being correctly assigned during wixpack creation. This PR just moves it outside to be an input to wix instead of creating a variable in wxi.


PranavSenthilnathan:
This fix was applied directly to https://github.com/dotnet/dotnet/pull/2538.

PR #2542, opened 2025-09-22
API diffing test for source-built SDK

NikolaMilosavljevic:
Fixes https://github.com/dotnet/source-build/issues/4795

This test is diffing APIs between MSFT and source-built SDKs.

It uses `Microsoft.DotNet.ApiCompat.Task` package and `ValidateAssembliesTask` in strict mode.  We use explicit version of this package and the dependent `Microsoft.CodeAnalysis` package. VMR tests consume packages from online sources only. There is an associated maintenance cost

## Some notes on implementation

Original plan was to run an external process from the new test `CompareMsftToSbAPIs` to run `dotnet msbuild` of the `ValidateAssemblies.proj` but that failed in `source-build-sdk-diff-tests` pipeline - we always run these tests on the same host container, but try to use the rid-specific SDK - this doesn't work due to missing native dependencies.

As a result, an alternative option was needed. I added the new project as a `ProjectReference` of `Microsoft.DotNet.SourceBuild.Tests.csproj` and only if `MsftSdkTarballPath` is specified - which is the case in `sdk-diff` tests.


test/Microsoft.DotNet.SourceBuild.Tests/SdkContentTests.cs
```diff
@@ -225,4 +265,18 @@ private static string RemoveDiffMarkers(string source)
         Regex diffSegmentRegex = new("^@@ .* @@", RegexOptions.Multiline);
         return diffSegmentRegex.Replace(result, "@@ ------------ @@");
     }
+
+    private static void DeleteUnmatchedDlls(string sourceSdkDir, string targetSdkDir)
```
MichaelSimons: Where/how does this get used?

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ApiDiff.suppression
```diff
@@ -0,0 +1,304 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- https://learn.microsoft.com/dotnet/fundamentals/package-validation/diagnostic-ids -->
+<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
+    <DiagnosticId>CP0001</DiagnosticId>
```
MichaelSimons, line 5: Have you analyzed these diffs to know if they are expected or if they can/should be addressed?  If not I think that can be handled via a follow-up issue.

MichaelSimons, line 5: The end user UX of this diff files is fairly important IMO.  In other words, when there is a test failure, being able to interpret what is happening from the diff context alone is important.  With that in mind, the diagnostic Ids alone are not helpful.  Does APICompat have the ability to include the human readable description so that you don't have to lookup the id?

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ValidateAssemblies.proj
```diff
@@ -0,0 +1,67 @@
+<Project Sdk="Microsoft.Build.NoTargets">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
MichaelSimons: Can this use netcurrent?  If not a comment should be included.

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ValidateAssemblies.proj
```diff
@@ -0,0 +1,67 @@
...
+  </ItemGroup>
+
+  <UsingTask TaskName="ValidateAssembliesTask"
+             AssemblyFile="$(PkgMicrosoft_DotNet_ApiCompat_Task)/tools/net/Microsoft.DotNet.ApiCompat.Task.dll" />
+
+  <Target Name="RunApiCompat" DependsOnTargets="Restore" BeforeTargets="Build">
+    <MakeDir Directories="$(MsftPath)" Condition="!Exists('$(MsftPath)')" />
+    <MakeDir Directories="$(SbPath)" Condition="!Exists('$(SbPath)')" />
+
+    <Exec Command="tar -xzf $(MsftSdkTarballPath) -C $(MsftPath)" />
```
MichaelSimons, line 29: Should these be removed after the validation is complete?

MichaelSimons, line 29: In the spirit of fast tests, it looks as those the tests will not extract these tarballs multiple times (here and in CompareMsftToSbAssemblyVersions) which is unfortunate.

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ApiDiff.suppression
NikolaMilosavljevic: Will check on better messages.

NikolaMilosavljevic: There is an issue I created last week: https://github.com/dotnet/source-build/issues/5344

NikolaMilosavljevic: Some of these are expected, but we definitely need to investigate issues in few assemblies.

test/Microsoft.DotNet.SourceBuild.Tests/SdkContentTests.cs
NikolaMilosavljevic: Good catch - that's a remnant of the original implementation. Will remove it.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/2542/commits/94f4b11ac7525bcb5c5461c823b6f950c47c5acc

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ValidateAssemblies.proj
NikolaMilosavljevic: Yes - will fix.

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ValidateAssemblies.proj
NikolaMilosavljevic: Yeah, this was a leftover from the original test project - fixed with https://github.com/dotnet/dotnet/pull/2542/commits/082f8fe88aa710452d1758f9303060091280532c

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ValidateAssemblies.proj
NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/2542/commits/a54d2fab1eed01996a38281512a8a52f7c0eaf72

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ApiDiff.suppression
NikolaMilosavljevic: I'm not sure if it's possible to have more verbose suppression entries. @ViktorHofer is this not possible today? It would seem like a good feature to add to the task.

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ValidateAssemblies.proj
NikolaMilosavljevic: I'll think about a way to share the extracted layouts and will either follow up here or in a new PR. We'd definitely need to ensure that neither of the tests is modifying the extracted layouts.

test/Microsoft.DotNet.SourceBuild.Tests/assets/SdkContentTests/ApiDiff.suppression
MichaelSimons: Not a blocker but something that would improve the UX IMO.

NikolaMilosavljevic: I've updated https://github.com/dotnet/source-build/issues/5344 with some initial findings/groupings of various issues. More to follow in that issue.

NikolaMilosavljevic:
@MichaelSimons if you're OK with these changes, can you approve the PR?

MichaelSimons (approved):

PR #2585, opened 2025-09-24
[.NET10 RC2] Unset other DOTNET_ROOT env vars when launching apphosts

jjonescz:
Should fix problems when `DOTNET_ROOT_<arch>` or `DOTNET_ROOT(x86)` var is set from the outside and the apphost would use those instead of the `DOTNET_ROOT` var we set.

For example, in NuGet.Client CI: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1156878&view=logs&j=f105d2c8-32d0-5cff-b114-87cf330929dd&t=f92a2093-7359-5d67-e113-8e3692a1dec4&l=94441

Follow up on https://github.com/dotnet/roslyn/pull/80026.

### DIFF
--- a/src/roslyn/src/Compilers/Core/MSBuildTask/ManagedToolTask.cs
+++ b/src/roslyn/src/Compilers/Core/MSBuildTask/ManagedToolTask.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Resources;
 using System.Text;
 using Microsoft.Build.Framework;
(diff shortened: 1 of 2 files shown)


src/roslyn/src/Compilers/Core/MSBuildTask/ManagedToolTask.cs
```diff
@@ -239,10 +240,18 @@ internal static string GetBuildTaskDirectory()
...
             if (RuntimeHostInfo.GetToolDotNetRoot() is { } dotNetRoot)
             {
                 Log.LogMessage("Setting {0} to '{1}'", RuntimeHostInfo.DotNetRootEnvironmentName, dotNetRoot);
-                EnvironmentVariables = [.. EnvironmentVariables ?? [], $"{RuntimeHostInfo.DotNetRootEnvironmentName}={dotNetRoot}"];
+                EnvironmentVariables =
+                [
+                    .. EnvironmentVariables?.Where(static e => !e.StartsWith(RuntimeHostInfo.DotNetRootEnvironmentName, StringComparison.OrdinalIgnoreCase)) ?? [],
+                    .. Environment.GetEnvironmentVariables().Cast<System.Collections.DictionaryEntry>()
+                        .Where(e => ((string)e.Key).StartsWith(RuntimeHostInfo.DotNetRootEnvironmentName, StringComparison.OrdinalIgnoreCase))
+                        .Select(e => $"{e.Key}="),
```
JoeRobich, line 252: Could `DOTNET_ROOT` be in this list already? Is it ok to set both
```
DOTNET_ROOT=
DOTNET_ROOT={dotNetRoot}
```
Does last one win?

JoeRobich (approved):

src/roslyn/src/Compilers/Core/MSBuildTask/ManagedToolTask.cs
jjonescz: Yes, it could be in the list but the last one wins

PR #2586, opened 2025-09-24
Update Microsoft.NETCore.Native.Windows.targets

frg2089:
Fixed a bug where AOT compilation failed when the path contained spaces.


ViktorHofer:
@MichalStrehovsky PTAL

akoeplinger:
LGTM and thanks for the contribution.  Per the [contributing guidelines](https://github.com/dotnet/dotnet/blob/afabede14c25760c8f77a36e872dcfbd3daca5d2/README.md#contribution), this PR should be opened against the [runtime repo](https://github.com/dotnet/runtime)

> Contribution to the .NET product should currently be done mostly in the constituent repositories. The reasons for this are two-fold:
> We want to slowly ramp up direct VMR changes to avoid surprises.
> The individual repositories still have the best validation for most changes.

frg2089:
here dotnet/runtime#120368

PR #2615, opened 2025-09-25
Fix x64 runtime installer on x86/arm64

PranavSenthilnathan:
Fixes https://github.com/dotnet/runtime/issues/120108

When installing using an x64 installer on arm64, the PROGRAMFILES_DOTNET (used to place the host) is not correctly computed because WiX has not resolved `ProgramFiles6432Folder` yet. This change makes it use `ProgramFiles64Folder` instead which is valid because this code only applies to x64 installers on x86/arm64 machines.


NikolaMilosavljevic:
@mmitche

NikolaMilosavljevic (approved):

PR #2376, opened 2025-09-11
Add es-metadata.yml for each repo in the VMR.

mmitche:

PR #2657, opened 2025-09-29
[release/10.0.1xx] Add es-metadata.yml for each repo in the VMR

mmitche:
Backport of #2376

### DIFF
--- a/src/runtime/src/libraries/es-metatadata.yml
+++ b/src/runtime/src/libraries/es-metatadata.yml
@@ -5,4 +5,4 @@ accountableOwners:
 routing:
   defaultAreaPath:
     org: devdiv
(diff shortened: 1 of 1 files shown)


src/runtime/src/libraries/es-metatadata.yml
```diff
@@ -0,0 +1,8 @@
+schemaVersion: 0.0.1
+isProduction: true
+accountableOwners:
+  service: 7a9b52f6-7805-416c-9390-343168c0cdb3
+routing:
+  defaultAreaPath:
+    org: devdiv
+    path: "DevDiv\NET Libraries"
```
premun:
```suggestion
    path: DevDiv\NET Libraries
```

PR #2678, opened 2025-09-30
Fix ANCM path for WOW64 support

wtgodbe:
We always want to install to the 32-bit Program files directory for WOW64 support, rather than relying on Wix to determine the installer architecture.

Test build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2805462&view=results


src/aspnetcore/src/Installers/Windows/AspNetCoreModule-Setup/ANCMV2/aspnetcoremodulev2.wxs
```diff
@@ -260,7 +260,7 @@
                 </RegistryKey>
             </Component>
 
-            <StandardDirectory Id="ProgramFiles6432Folder">
+            <StandardDirectory Id="ProgramFilesFolder">
```
wtgodbe, line 263: This was the bug

wtgodbe: (I think)

PR #2680, opened 2025-09-30
Fix YAML syntax for file pattern matching

mmitche:


NikolaMilosavljevic (approved):

mmitche:
I'm not positive on this, but the current pattern is not working :) Looking at some other examples this is what I saw.

PR #2706, opened 2025-10-02
Temporarily disable hard linking on sdk installer layout creation

mmitche:
This hard linking is a problem because the original source for the hard link gets written to by the testing overlay update. Depending on timing, the bundled versions files may get updated, leading to incorrect implicit versions.


mmitche:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2806992&view=results

marcpopMSFT (approved):

PR #2720, opened 2025-10-03
Add tests for verification of expected set of Linux packages

NikolaMilosavljevic:
Fixes: https://github.com/dotnet/sdk/issues/51086

New tests will ensure that we are producing all the expected Linux packages. This will catch any issues early, way before the packages would need to be consumed, i.e. during product release.

### DIFF
--- a/test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
+++ b/test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
@@ -734,16 +734,9 @@ public partial class LinuxInstallerTests : IDisposable
         string extension = packageType == PackageType.Rpm ? "*.rpm" : "*.deb";
         string[] expectedPatterns = GetExpectedPackagePatterns(packageType);
 
-        // Find all packages of the specified type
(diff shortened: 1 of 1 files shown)


test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -712,4 +728,112 @@ private static List<string> ParseDebControlDependencies(string contents)
...
+            {
+                errorMessage.AppendLine("Unexpected packages:");
+                foreach (string pkg in extra)
+                {
+                    errorMessage.AppendLine($"  + {pkg}");
+                }
+            }
+
+            Assert.Fail(errorMessage.ToString());
+        }
```
mthalman: With `FluentAssertions`, this can be simplified to:
```suggestion
        normalizedActual.Should().BeEquivalentTo(expectedPatterns);
```

NikolaMilosavljevic: Thanks. Fixed with https://github.com/dotnet/dotnet/pull/2720/commits/8b2a067e165f500c2567c76c52b840c571f6d823

I've simplified this by using list comparison and custom logging to ensure we have a correct output with complete list of expected and actual packages.

PR #2750, opened 2025-10-06
Use non-stable version for SB SDK filename on publish

mthalman:
The source-built SDK that is published as an artifact of a VMR build uses the same filename that is produced by the output of the `sdk` repo. In the case of stable versioning, this can lead to collisions in the storage account because it doesn't have any other part of the path that produces uniqueness. This differs from the source-built artifacts tarball, for example, which always uses a non-stable version in its filename.

This is fixed by adding a symlink file which references the original SDK tarball. The symlink uses the non-stable version filename. The metadata constructed for publishing then references the symlink file so that the publishing manifest file is defined as we want. This allows the SDK tarball that is produced by the VMR built to still be named as it has always been but with the added benefit of always using a non-stable version filename when actually publishing that file.

Review the changes by viewing each commit. The first commit does a small refactor by moving logic source-build-specific logic from `Publishing.props` to `PublishSourceBuild.props`. The rest is actual changes to fix the issue.

### DIFF
--- a/eng/PublishSourceBuild.props
+++ b/eng/PublishSourceBuild.props
@@ -199,13 +199,15 @@
          This tarball should be versioned in the filename the same as the output of the sdk repo. -->
     <Copy SourceFiles="@(SdkRepoSdkTarball)"
           DestinationFolder="$(ArtifactsAssetsDir)"
-          SkipUnchangedFiles="true" />
-
+          SkipUnchangedFiles="true"
+          UseHardlinksIfPossible="true" />
+    
     <ItemGroup Condition="'$(IsPublishPass2)' != 'true'">
-      <SdkTarball Include="$(ArtifactsAssetsDir)/$(SourceBuiltArtifactsTarballName).*$(ArchiveExtension)">
+      <!-- Add SDK archive to assets root as source-only build partners expect the file to be there. -->
+      <SourceBuildArtifact Include="$(ArtifactsAssetsDir)/$(SdkFileNamePrefix)*$(ArchiveExtension)">
         <Kind>Blob</Kind>
         <IsShipping>true</IsShipping>
-      </SdkTarball>
+      </SourceBuildArtifact>
       <!-- Add the symbols archives to the publishing output. -->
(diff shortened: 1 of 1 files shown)


mmitche:
@mthalman Let me know when you're satisfied with this change, so I can validate it as part of the servicing exercise.

mthalman:
I'm still looking at it. Something's not quite right.

The file at `assets/Release` is named `dotnet-sdk-10.100-rtm-centos.10-x64.tar.gz` when it should be named `dotnet-sdk-10.0.100-ci-centos.10.x64.tar.gz`. I don't see this locally. Not sure if this something weird about using file linking and/or AzDO artifact publishing or not. Investigating.

mthalman:
This is so confusing. The `Copy` task says it's copying the SDK tarball to the assets location but when you look at the pipeline artifacts, it doesn't exists there:

```
Copy
    Assembly = /__w/1/s/.dotnet/sdk/10.0.100-rc.1.25451.107/Microsoft.Build.Tasks.Core.dll
    Parameters
        SourceFiles = /__w/1/s/artifacts/assets/Release/Sdk/10.0.100-rtm/dotnet-sdk-10.0.100-ci-centos.10-x64.tar.gz
        DestinationFolder = /__w/1/s/artifacts/assets/Release/
        SkipUnchangedFiles = True
    Copying file from "/__w/1/s/artifacts/assets/Release/Sdk/10.0.100-rtm/dotnet-sdk-10.0.100-ci-centos.10-x64.tar.gz" to "/__w/1/s/artifacts/assets/Release/dotnet-sdk-10.0.100-ci-centos.10-x64.tar.gz".
```

But that `assets/Release/dotnet-sdk-10.0.100-ci-centos.10-x64.tar.gz` file is not there in the artifacts:

image

I have no idea what's going on.

mmitche:
I think I see it. Your copy is from /s/artifacts to /s/artifacts. The artifact output is in the staging directory, /a/artifacts. Most of the output gets there via PushToBuildStorage.

mthalman:
Ok, this looks to be working now.

akoeplinger:
@mthalman was this ported to main?

PR #2722, opened 2025-10-03
Azure Linux DEPS packages do not have a newkey variant

NikolaMilosavljevic:
Follow up on https://github.com/dotnet/dotnet/pull/2720

Updates the test based on recent change in packages produced: https://github.com/dotnet/dotnet/pull/2694

PR #2773, opened 2025-10-07
[release/10.0.1xx] Add tests for verification of expected set of Linux packages

NikolaMilosavljevic:
Backport of https://github.com/dotnet/dotnet/pull/2720 and https://github.com/dotnet/dotnet/pull/2722

New tests will ensure that we are producing all the expected Linux packages. This will catch any issues early, way before the packages would need to be consumed, i.e. during product release.

The tests were enabled in main. This backport ensures that we have the same tests in 10.

## Notes

**Infra change**


NikolaMilosavljevic:
@jkoritzinsky @dotnet/product-construction - this is a simple backport. We'd like to have these tests in 10 to prevent potential regressions in the future.

PR #2788, opened 2025-10-08
Remove trailing dash on dotnet-sdk

mmitche:


eng/download-source-built-archive.sh
mmitche, line 95: The version delimeter is already appended, (- or .), so the sdk base file name shouldn't have the delimiter. However, I don't know all of the permutations that this is used in so PTAL.

mthalman (approved):

akoeplinger:
port to main?

mthalman:
It has been ported. I cherry-picked it into https://github.com/dotnet/dotnet/pull/2772.

PR #2794, opened 2025-10-08
Use backward-compatible GUID generation in Wix 5

NikolaMilosavljevic:
This includes the following changes:
- Update to `CreateWixBuildWixpack` to support additional build/compilation options
- Update to shared infra to pass `-bcgg` to wix build and wixpack task
- Update to `aspnetcore` and `sdk` infra to add `-bcgg` to Compiler options, and utilize `AdditionalOptions` in wixpack task to have the same option during wix repack
- wix 5 update to support `-bcgg` option

### DIFF
--- a/src/runtime/eng/Versions.props
+++ b/src/runtime/eng/Versions.props
@@ -168,11 +168,11 @@
     <!-- The package path for python in src/mono/mono.proj needs to be updated if this changes-->
     <EmsdkVersion>3.1.56</EmsdkVersion>
     <!-- WiX 5+ dependencies for MSI generation -->
-    <MicrosoftWixVersion>5.0.2-dotnet.2737382</MicrosoftWixVersion>
-    <MicrosoftWixToolsetUIWixextVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetUIWixextVersion>
-    <MicrosoftWixToolsetDependencyWixextVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetDependencyWixextVersion>
-    <MicrosoftWixToolsetUtilWixextVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetUtilWixextVersion>
-    <MicrosoftWixToolsetBalWixextVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetBalWixextVersion>
-    <MicrosoftWixToolsetHeatVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetHeatVersion>
+    <MicrosoftWixVersion>5.0.2-dotnet.2811440</MicrosoftWixVersion>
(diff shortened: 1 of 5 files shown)


src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/src/CreateWixBuildWixpack.cs
```diff
@@ -28,6 +28,8 @@ namespace Microsoft.DotNet.Build.Tasks.Installers
      */
     public class CreateWixBuildWixpack : Task
     {
```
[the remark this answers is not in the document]
NikolaMilosavljevic: Not necessary.

src/runtime/eng/Versions.props
```diff
@@ -168,11 +168,11 @@
...
-    <MicrosoftWixToolsetDependencyWixextVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetDependencyWixextVersion>
-    <MicrosoftWixToolsetUtilWixextVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetUtilWixextVersion>
-    <MicrosoftWixToolsetBalWixextVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetBalWixextVersion>
-    <MicrosoftWixToolsetHeatVersion>5.0.2-dotnet.2737382</MicrosoftWixToolsetHeatVersion>
+    <MicrosoftWixVersion>5.0.2-dotnet.2811440</MicrosoftWixVersion>
+    <MicrosoftWixToolsetUIWixextVersion>5.0.2-dotnet.2811440</MicrosoftWixToolsetUIWixextVersion>
+    <MicrosoftWixToolsetDependencyWixextVersion>5.0.2-dotnet.2811440</MicrosoftWixToolsetDependencyWixextVersion>
+    <MicrosoftWixToolsetUtilWixextVersion>5.0.2-dotnet.2811440</MicrosoftWixToolsetUtilWixextVersion>
+    <MicrosoftWixToolsetBalWixextVersion>5.0.2-dotnet.2811440</MicrosoftWixToolsetBalWixextVersion>
+    <MicrosoftWixToolsetHeatVersion>5.0.2-dotnet.2811440</MicrosoftWixToolsetHeatVersion>
```
joeloff, line 176: These should all be the same version eventually, ideally just use the version from Arcade.

PranavSenthilnathan: Are you suggesting avoiding the duplication by doing something like:
`<MicrosoftWixToolsetUIWixextVersion>$(MicrosoftWixVersion)</MicrosoftWixToolsetUIWixextVersion>`

Or is there a way to get the version from arcade itself instead of hardcoding it in this file?

NikolaMilosavljevic: @PranavSenthilnathan this should be fixed, in a separate PR.

PR #2806, opened 2025-10-09
Bootstrap repo fix for 2xx

mthalman:
VMR builds for `release/10.0.2xx` are failing for SB legs that are not based on CentOS with the following error:

`/repos/dotnet2/src/source-build-reference-packages/src/targetPacks/ILsrc/netstandard.library/2.0.3/NETStandard.Library.2.0.3.csproj error NU1603: NETStandard.Library.2.0.3 depends on runtime.ubuntu.24.04-x64.Microsoft.NETCore.ILAsm (>= 10.0.0-rc.1.25451.107) but runtime.ubuntu.24.04-x64.Microsoft.NETCore.ILAsm 10.0.0-rc.1.25451.107 was not found. runtime.ubuntu.24.04-x64.Microsoft.NETCore.ILAsm 10.0.0-rc.2.25479.113 was resolved instead. [/repos/dotnet2/.dotnet/sdk/10.0.100-rc.2.25479.113/NuGet.targets]`

This occurs because the `MicrosoftNETCoreILAsmPackageVersion` property is not being overridden. It's using the version from PSB instead of from the shared components (1xx branch).

This is due to the logic which collects the packages that are relevant for representation within the `PackageVersions.*.SharedComponents.props` file. That excludes _all_ packages when processing the `source-build-reference-packages`. But we do want to allow for packages where the SBRP repo itself has a dependency on.

This is fixed by explicitly allowing for SBRP's dependency packages to be included in this process of generating the SharedComponents.props file.

Another fix that was needed was to fix the pipeline logic to allow for artifact bootstrapping in 2xx builds. This was being skipped in the script.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,7 +1,7 @@
 <Project>
 
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" Condition="'$(SkipArcadeSdkImport)' != 'true'" />
-  <Import Project="$(RepositoryEngineeringDir)shared-source-only.targets" />
+  <Import Project="$(RepositoryEngineeringDir)shared-source-only.targets" Condition="'$(SkipSharedSourceOnlyImport)' != 'true'" />
 
   <UsingTask TaskName="Microsoft.DotNet.UnifiedBuild.Tasks.GetKnownArtifactsFromAssetManifests" AssemblyFile="$(MicrosoftDotNetUnifiedBuildTasksAssembly)" TaskFactory="TaskHostFactory" />
 
--- a/eng/clean-shared-components.proj
+++ b/eng/clean-shared-components.proj
@@ -3,6 +3,7 @@
 
   <PropertyGroup>
     <TargetFramework>$(NetCurrent)</TargetFramework>
+    <SkipSharedSourceOnlyImport>true</SkipSharedSourceOnlyImport>
   </PropertyGroup>
 
   <ItemGroup>


eng/shared-source-only.targets
```diff
@@ -4,9 +4,19 @@
 
   <Target Name="GetFilteredSharedComponentPackages"
           Outputs="@(SharedComponentFilteredPackages)">
+    
+    <ItemGroup>
```
MichaelSimons: Mirroring the version.details.xml list from SBRP here is not ideal IMO as it creates a maintenance issue.  Is there a way this can be read from SBRP?

repo-projects/Directory.Build.targets
```diff
@@ -353,7 +353,13 @@
     <MSBuild Projects="$(MSBuildProjectFullPath)"
              Targets="GetFilteredSharedComponentPackages"
              Properties="SharedComponentFilterMode=NonToolingOnly"
-             Condition="!$([System.String]::new(';$(BootstrapArcadeRepos);').Contains(';$(RepositoryName);'))">
+             Condition="$(RepositoryName) != 'source-build-reference-packages' and !$([System.String]::new(';$(BootstrapArcadeRepos);').Contains(';$(RepositoryName);'))">
```
MichaelSimons: Why isn't this also an issue for arcade, the other bootstrap repo?

eng/shared-source-only.targets
mthalman: Fixed in 764c207

repo-projects/Directory.Build.targets
mthalman: Fixed in 764c207

eng/shared-source-only.targets
```diff
@@ -1,12 +1,22 @@
...
+  <UsingTask TaskName="Microsoft.DotNet.UnifiedBuild.Tasks.GetDependencyNamesFromVersionDetails" AssemblyFile="$(MicrosoftDotNetUnifiedBuildTasksAssembly)" TaskFactory="TaskHostFactory" />
 
   <Target Name="GetFilteredSharedComponentPackages"
           Outputs="@(SharedComponentFilteredPackages)">
 
+    <PropertyGroup>
+      <_VersionDetailsXml Condition="'$(PackageVersionPropsFlowType)' == 'DependenciesOnly'">$(ProjectDirectory)eng/Version.Details.xml</_VersionDetailsXml>
+    </PropertyGroup>
+    <GetDependencyNamesFromVersionDetails VersionDetailsPath="$(_VersionDetailsXml)"
+                                          Condition="'$(SharedComponentFilterMode)' == 'BootstrapRepoDepsOnly'">
```
MichaelSimons: Can a shared property e defined to encapsulate this name given it is used in three places?

mthalman: Fixed with https://github.com/dotnet/dotnet/pull/2806/commits/182d8d0b13c11eebfdfd2aba1f298d94603767b7

PR #2889, opened 2025-10-14
Reset nuget.client to release/7.0.x

dkurepa:
FR request from @nkolev92


dkurepa:
oops wrong target branch, my bad

PR #2890, opened 2025-10-14
Reset nuget.client to release/7.0.x

dkurepa:
FR request from @nkolev92


dkurepa:
we don't actually need this PR, VMR is already on the correct branch because of the way nuget.client changed its config.
This is essentially a FF PR, so I'll just close it

nkolev92:
> we don't actually need this PR, VMR is already on the correct branch because of the way nuget.client changed its config.

Do I interpret this as, the branch is determined based on which branch the commit belongs to, assuming there's only 1 branch with that commit?

dkurepa:
> Do I interpret this as, the branch is determined based on which branch the commit belongs to, assuming there's only 1 branch with that commit?

The subscription forward flowing into VMRs `10.0.1xx` branch was set to flow builds from the `VS 18.0` channel.
since `nuget.clients` `release/7.0.x` is now publishing builds to `VS 18.0`, this means that we already have the correct branch.
This was probably done as soon as `dev` branched to `release/7.0.x` so there was no need for a reset, because the contents didn't diverge yet

PR #2893, opened 2025-10-14
Add native signing auth config from runtime repo

mthalman:
Signed builds in the 10.0.2xx branch fail for Windows legs with the following error in the `Setup WIF auth for ESRP client` step:

```
Get-Content : Cannot find path 'D:\a\_work\1\s\src\runtime\eng\native\signing\auth.json' because it does not exist.
```

Resolved by copying the auth.json file from the runtime repo into the VMR `eng` directory. Updated the pipeline to reference this (the new path is referenced by default so it's resolved by removing the overridden path).

### DIFF
--- a/eng/native/signing/auth.json
+++ b/eng/native/signing/auth.json
@@ -1,0 +1,20 @@
+{
+    "Version" : "1.0.0",
+    "AuthenticationType" : "AAD_MSI_WIF",
+    "TenantId" : "975f013f-7f24-47e8-a7d3-abc4752bf346",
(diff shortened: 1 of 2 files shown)


ViktorHofer:
Assuming we don't sign in runtime anymore, should we delete the file in runtime?

mthalman:
I don't know if it's still being used or not.

ViktorHofer:
@hoyosjs and/or @jkoritzinsky will know for sure

jkoritzinsky:
We can delete the files related to DAC signing from runtime infra and move them to the VMR infra. It's only being used in the VMR now (since runtime official build was disabled)

PR #2895, opened 2025-10-14
Clean existing .dotnet dir

mthalman:
Signed builds in the 10.0.2xx branch fail in the `SB_CentOSStream10_Offline_MsftSdk_x64` job with the following error in the Prep Source Build step:

```
mkdir: cannot create directory '/__w/1/s/.dotnet': File exists
```

This happens because the "Install .NET 8.0 SDK for MicroBuild Plugin" step has already installed an SDK at that location.

This is resolved by ensuring the `.dotnet` directory is removed. The SDK installed for MicroBuild is no longer needed at this point in the build.

### DIFF
--- a/eng/pipelines/templates/jobs/vmr-build.yml
+++ b/eng/pipelines/templates/jobs/vmr-build.yml
@@ -526,6 +526,7 @@ jobs:
 
           if [[ "$prepSdk" == "false" ]]; then
             customPrepArgs="${customPrepArgs} --no-sdk"
+            rm -rf $(sourcesPath)/.dotnet
             mkdir $(sourcesPath)/.dotnet
             previousSdkPath="$(sourcesPath)/prereqs/packages/archive/dotnet-sdk-*.tar.gz"
             eval tar -ozxf "$previousSdkPath" -C "$(sourcesPath)/.dotnet"


eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -526,6 +526,7 @@ jobs:
 
           if [[ "$prepSdk" == "false" ]]; then
             customPrepArgs="${customPrepArgs} --no-sdk"
+            rm -rf $(sourcesPath)/.dotnet
```
ellahathaway, line 529: This doesn't feel like the correct approach. Signing with MicroBuild requires .NET 8. In other words, it's not just the MicroBuild installation that requires .NET 8, it's the signing as well.

mthalman: I ran a signed build with these changes and it worked. It seems to work with a .NET 10 SDK.

ellahathaway: We've had issues in the past where the ESRP CLI expects .NET 8. Maybe this has changed recently or is only a requirement on mac? cc @mattnorflus @mmitche

mattnorflus: [We made this change to enable roll forward with esrpcli.](https://dev.azure.com/devdiv/Engineering/_git/Sign/commit/9603a199d019e1cf0b4379e30db49236d562d861?refName=refs/heads/main&path=/src/sign_v45/ESRPCliDll.cs) The tool is still natively .NET 8 so the official guidance is to use .NET 8, but as long as using roll-forward is an acceptable risk, we certainly have no objections to removing .NET 8 now and relying on .NET 10.

ellahathaway: Thanks for the info. In that case, I'm okay with this change.

FYI, this existing directory issue is also another example of something that will be fixed by https://github.com/dotnet/arcade/pull/16091.

MichaelSimons: Should a tracking issue be logged to remove this once dotnet/arcade#16091 flows in?

mthalman: Created issue to revert the change: https://github.com/dotnet/source-build/issues/5374

PR #2927, opened 2025-10-15
Merge from rc2 release tag

mmitche:


premun (commented):
@mmitche a lot of changes is file permission change and it seems to be going from 0755 to 0644. Which changed here: https://github.com/dotnet/dotnet/commit/3b9f4df53c51fc657ce9e536720df5add7eeb9ee

Maybe you some file copying/merging went wrong on Windows?

JanProvaznik (approved):

akoeplinger:
@premun 644 is the more correct one, those xml files shouldn't be marked executable, but I agree we shouldn't change it as part of this PR probably to keep the diff easier.

premun:
I agree that `+x` on XMLs also caught my eye but to me, the more correct one is what is in the original repo's 10.0 branch so that we don't get conflicts each time they change in the future.

akoeplinger:
I reverted the file permission changes here and we can fix the permissions in the source repo separately

akoeplinger (approved):

akoeplinger:
> we can fix the permissions in the source repo separately

I looked a bit and this seems to be some side effect of how the packages are extracted in source-build-reference-packages (possibly even in the nuget apis). I don't think I care that much to open that can of worms :)

akoeplinger:
> src/arcade/src/Microsoft.DotNet.SourceBuild/tasks/Microsoft.DotNet.SourceBuild.Tasks.csproj(0,0): error NU1903: (NETCORE_ENGINEERING_TELEMETRY=Restore) Package 'Microsoft.Build.Tasks.Core' 17.12.36 has a known high severity vulnerability, https://github.com/advisories/GHSA-w3q9-fxm7-j8fq

this just popped up, we need to fix it. not sure why it only shows on the iOSSimulator job though.

PR #2955, opened 2025-10-16
Update Ubuntu image used by Tag & Scan stage

NikolaMilosavljevic:
Removing the last use case of Ubuntu 20.04 image in VMR builds.


NikolaMilosavljevic:
Checks are hitting known issue with vulnerable package versions - being fixed with https://github.com/dotnet/dotnet/pull/2953

PR #2962, opened 2025-10-16
Generate source assets identical to GitHub assets

ellahathaway:
Fixes https://github.com/dotnet/source-build/issues/5375

Test run with the RC2 artifacts:
```
- sha256sum *tar.gz
38c6a77490c84810bc3045f4b96a7b89b65f79b291cddf8a88ccd4d5838257e9  dotnet-source-10.0.100-rc.2.25502.107.tar.gz (generated with CreateSourceArtifacts task)
38c6a77490c84810bc3045f4b96a7b89b65f79b291cddf8a88ccd4d5838257e9  v10.0.100-rc.2.25502.107.tar.gz (Generated by GitHub)
-  sha256sum *zip
c659acb345f5dc2ea951d682530d79c893d6f87233f05dd3d1b30284d39efa96  dotnet-source-10.0.100-rc.2.25502.107.zip (generated with CreateSourceArtifacts task)
c659acb345f5dc2ea951d682530d79c893d6f87233f05dd3d1b30284d39efa96  v10.0.100-rc.2.25502.107.zip (generated by GitHub)
```

### DIFF
--- a/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
+++ b/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
@@ -43,15 +43,7 @@ public class ProcessService(TaskLoggingHelper log, int timeout = 10 * 1000)
         {
             foreach (ProcessEnvironmentVariable env in environmentVariables)
             {
-                try
-                {
-                    processInfo.EnvironmentVariables.Add(env.Name, env.Value);
-                }
-                catch (ArgumentException)
-                {
-                    // If the environment variable already exists, update its value
-                    processInfo.EnvironmentVariables[env.Name] = env.Value;
-                }
+                processInfo.EnvironmentVariables[env.Name] = env.Value;
             }
         }
 


eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
```diff
@@ -44,6 +45,8 @@ public class CreateSourceArtifacts : BuildTask
     [Required]
     public required string OutputDirectory { get; init; }
 
+    private const string GitHubRepoName = "dotnet";
+    private const string GitHubTimezone = "America/Los_Angeles"; // GitHub uses this timezone for commit timestamps in zip metadata
```
ellahathaway, line 49: From what I can tell, this is the time zone GitHub uses when it generates the source assets. However, I cannot find concrete evidence to back this claim, which makes me a bit wary of setting the time zone as a constant.

The time zone is irrelevant for tars, but it is important for zips as it gets stored in the zip metadata. Without setting the time zone, the zips contain a DOS date/time diff of `2025 Oct 2 12:08:00` (GH) vs `2025 Oct 2 19:08:00` (CreateSourceArtifacts task).

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
```diff
@@ -43,7 +43,15 @@ public async Task<ProcessResult> RunProcessAsync(
         {
             foreach (ProcessEnvironmentVariable env in environmentVariables)
             {
-                processInfo.EnvironmentVariables.Add(env.Name, env.Value);
+                try
+                {
+                    processInfo.EnvironmentVariables.Add(env.Name, env.Value);
```
MichaelSimons: Can you use the indexer to handle both cases, add and update?  It would avoid exceptions all together.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
MichaelSimons: @mmitche - IIRC you looked into the GH logic in the past.  Thoughts?

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/Services/ProcessService.cs
ellahathaway: Addressed in https://github.com/dotnet/dotnet/pull/2962/commits/4aa15d526e7a89574b9ed3fb8cf437c004cf76f4

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
mmitche: This seems reasonable given the constraints. IIRC in the past they changed the gzip implementation and broke a bunch of stuff, so they will be careful to do this in the future. It's not ironclad, and if they did decide to change the generation method then all our signatures would get broken.

Because we're pre-generating the signatures, there is more risk because it also means that if our tools change (e.g. gzip implementation), then future artifacts could have broken signatures.

Is the pipeline is doing verification at release time? Ideally before the artifact is uploaded.

ellahathaway: Yes, the SB release pipeline does the signature verification at release time. We're transitioning away from using the SB release pipeline, so I will add verification to the MSFT RP on release day so we will at least know if the signatures don't match the public GH source for whatever reason.

PR #3004, opened 2025-10-20
[release/10.0.1xx] Add an SDK element to the root global.json.

mmitche:
Without this root element, CG's detection will fall back to trying to look at what SDKs are on the path. In certain contexts, SDKs that are certainly not going to get used are going to get reported, and lead to false positive SDK alerts in CG.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "10.0.100-rc.2.25502.107",
+    "version": "10.0.101",
     "paths": [
       ".dotnet",
       "$host$"


ericstj (approved):

global.json
```diff
@@ -1,4 +1,13 @@
 {
+  "sdk": {
+    "version": "10.0.100-rc.2.25502.107",
+    "rollForward": "latestFeature",
```
MichaelSimons: I encountered this [same issue in SBRP](https://github.com/dotnet/source-build-reference-packages/pull/1434) today.  9.0.305 was installed on the build agent where the global.json specifies 9.0.111.  With latestFeature it picks up 9.0.305 which is vulnerable.  In SBRP, latestPatch feels more appropriate.

mmitche:
Open question...should we just be using the sdk element?

ViktorHofer:
The error implies that the wrong SDK is used with this change in some cases. Not sure though.

akoeplinger:
yeah the error is 
> /__w/1/s/.dotnet/sdk/10.0.100-rc.2.25502.107/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(120,5): error NETSDK1083: The specified RuntimeIdentifier 'centos.10-x64' is not recognized. See https://aka.ms/netsdk1083 for more information. [/__w/1/s/artifacts/bin/Microsoft.DotNet.SourceBuild.Tests/Release/projects-202510230956048138/WebScenarioTests_WebApp_CSharp/WebScenarioTests_WebApp_CSharp.csproj]

which means it wasn't actually using the dotnet sdk from `/__w/1/s/artifacts/obj/extracted-dotnet-sdk`.

we probably need to put a special global.json there so it doesn't pick up the wrong one

Issue #3019, opened 2025-10-21
Microsoft.NET.Workload.Mono.ToolChain.Current.Manifest-10.0.100 is still built in runtime with 10.0.0 version


akoeplinger:
When we moved building the workload manifests from runtime to the sdk repo we kept the original one in the runtime repo in case we needed to revert. It was supposed to not be built but looks like that's still happening for some reason.

It means we now publish both a 10.0.0 and a 10.0.100 versioned package:

<img width="903" height="619" alt="Image" src="https://github.com/user-attachments/assets/4191cee5-2e5d-431f-b56d-1286343afb80" />

ViktorHofer:
Should this be closed @akoeplinger?

PR #3021, opened 2025-10-21
Don't publish Microsoft.NET.Workload.Mono.ToolChain.Current.Manifest from runtime build

akoeplinger:
This regressed with https://github.com/dotnet/runtime/commit/e5798b9f08ab6def1848d226512815357e9ccb80 which removed removing the manifest packages from publishing in the runtime build with a runtime version number.

These packages are unused (only the ones with sdk-style version numbers are used) but they are confusing and they also seem to mess with darc dependency tracking.

Fixes https://github.com/dotnet/dotnet/issues/3019

### DIFF
--- a/src/runtime/eng/Publishing.props
+++ b/src/runtime/eng/Publishing.props
@@ -5,6 +5,11 @@
     <AutoGenerateSymbolPackages>false</AutoGenerateSymbolPackages>
   </PropertyGroup>
 
+  <!-- remove the manifest packages, they are built in the SDK repo now -->
+  <ItemGroup>
+    <Artifact Remove="$(ArtifactsShippingPackagesDir)*.Manifest-*.nupkg" />
+  </ItemGroup>
+
   <!-- The PGO runtime should always have External visibility, even if someone changes the default artifact visibility -->
   <ItemGroup>
     <Artifact Update="$(ArtifactsNonShippingPackagesDir)dotnet-runtime-pgo-*" Visibility="External" />


mmitche:
@akoeplinger Any risk for 10.0 that means we should get this in today (or not?)

akoeplinger:
@mmitche in theory it has no risk as nothing should be using these packages. 

but that also means it's "just" about reducing human confusion so I'd be fine taking this in 10.0.1 too

src/runtime/eng/Publishing.props
```diff
@@ -5,6 +5,11 @@
     <AutoGenerateSymbolPackages>false</AutoGenerateSymbolPackages>
   </PropertyGroup>
 
+  <!-- remove the manifest packages, they are built in the SDK repo now -->
+  <ItemGroup>
+    <Artifact Remove="$(ArtifactsShippingPackagesDir)*.Manifest-*.nupkg" />
+  </ItemGroup>
```
ViktorHofer, line 11: For my own education, if those packages are now built in the sdk repo, why do they even exist in runtime?

akoeplinger: it needed some untangling in the runtime build that we didn't get to for 10.0

akoeplinger:
can you please approve so we can get this in for 10.0.1? thanks

Tell-mode, infrastructure change with no actual impact to the product, Servicing-approved.

PR #3026, opened 2025-10-21
[release/10.0.2xx] .NET Source-Build 10.0.100-rc.2.25502.107 October 2025 Updates

ellahathaway:
Manual port of https://github.com/dotnet/dotnet/pull/2907


ellahathaway:
@mthalman - I expected there to be a flow of these changes to 2xx. Was there supposed to be a flow? If yes, do you know why that flow didn't happen?

mthalman:
> @mthalman - I expected there to be a flow of these changes to 2xx. Was there supposed to be a flow? If yes, do you know why that flow didn't happen?

1xx doesn't automatically flow to 2xx in the VMR right now. See https://github.com/dotnet/arcade-services/issues/5378. But even if it did, it wouldn't handle these changes. These changes should be handled by the release pipeline, just as it is for 1xx. But that wouldn't happen yet since we haven't had a 2xx release yet.

mthalman (approved):

PR #3042, opened 2025-10-22
Update sdk diff and license scan tests to trigger on feature band branches

MichaelSimons:
Fixes https://github.com/dotnet/source-build/issues/5381

### DIFF
--- a/eng/pipelines/source-build-sdk-diff-tests.yml
+++ b/eng/pipelines/source-build-sdk-diff-tests.yml
@@ -2,8 +2,13 @@ schedules:
 - cron: "0 12 * * 1-5"
   displayName: Run on weekdays at 12pm UTC
   branches:
+    # The 2xx, 3xx, 4xx triggers exist to cover the feature band pre-release time periods.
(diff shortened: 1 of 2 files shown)


eng/pipelines/source-build-sdk-diff-tests.yml
```diff
@@ -2,8 +2,13 @@ schedules:
 - cron: "0 12 * * 1-5"
   displayName: Run on weekdays at 12pm UTC
   branches:
+    # The 2xx, 3xx, 4xx triggers exist to cover the feature band pre-release time periods.
+    # They can be removed once released as the pipeline triggers below will handle the branches after release.
```
ellahathaway, line 6: Can you please link a tracking issue for the removal of these triggers?

MichaelSimons: I see the removal of these being done on an as needed basis - if they aren't causing an issue, I don't see us removing them.  I wish there was a better solution here.

PR #3045, opened 2025-10-22
[infra] Backport of pipeline yml fix for sdk-diff-tests pipeline

NikolaMilosavljevic:
Backport of the infra change from https://github.com/dotnet/dotnet/pull/2829

This enables publishing of newly generated suppression file for Api-Diff tests.


MichaelSimons:
This should also get backported to 2xx.

PR #3052, opened 2025-10-22
Add entitlements to roslyn apphosts

jkoritzinsky:
Fixes https://github.com/dotnet/runtime/issues/120967

This has the existing entitlements that Roslyn has always used.

### DIFF
--- a/src/sdk/src/Layout/redist/targets/GenerateLayout.targets
+++ b/src/sdk/src/Layout/redist/targets/GenerateLayout.targets
@@ -67,7 +67,7 @@
                    EnableMacOSCodeSign="$(SharedFrameworkRid.StartsWith('osx'))" />
     
     <Exec Command="codesign --sign - --entitlements '$(MSBuildProjectDirectory)/roslyn-entitlements.plist' %(_RoslynAppHost.RootDir)%(_RoslynAppHost.Directory)%(_RoslynAppHost.Filename)$(ExeExtension)"
-          Condition="'$(SharedFrameworkRid.StartsWith('osx'))' == 'true'" />
+          Condition="$(SharedFrameworkRid.StartsWith('osx'))" />
   </Target>
 
   <Target Name="PublishNETAnalyzers">


jeffschwMSFT (approved):
approved

agocke (approved):

jkoritzinsky:
Let's wait until the mac legs run on the unified-build-full pipeline before merging, just to be sure things are working right.

mmitche:
I'm pretty sure this froze on both arm64 adn x64 before it even got past aspnetcore. No idea why. We aren't normally running the full build in the open so deserves investigation but I am going to merge for now.

src/sdk/src/Layout/redist/targets/GenerateLayout.targets
```diff
@@ -65,6 +65,9 @@
                    AppBinaryName="%(_RoslynAppHost.Filename)%(_RoslynAppHost.Extension)"
                    IntermediateAssembly="%(_RoslynAppHost.FullPath)"
                    EnableMacOSCodeSign="$(SharedFrameworkRid.StartsWith('osx'))" />
+    
+    <Exec Command="codesign --sign - --entitlements '$(MSBuildProjectDirectory)/roslyn-entitlements.plist' %(_RoslynAppHost.RootDir)%(_RoslynAppHost.Directory)%(_RoslynAppHost.Filename)$(ExeExtension)"
+          Condition="$(SharedFrameworkRid.StartsWith('osx'))" />
```
jjonescz, line 70: How do other apphosts (like `dotnet.exe`) get the entitlements? Shouldn't `CreateAppHost` task take care of that?

akoeplinger: we add them during the dotnet/runtime build: https://github.com/dotnet/runtime/blob/9bcdd873eb4b93934ae34328845599f63253d32a/src/native/corehost/dotnet/CMakeLists.txt#L28-L30

but I just noticed this is missing the `-f`/`--force` option and the official build fails due to that

akoeplinger: https://github.com/dotnet/dotnet/pull/3069

akoeplinger: > Shouldn't CreateAppHost task take care of that?

Ideally yes (https://github.com/dotnet/runtime/issues/113707), that's why it worked before, but we ran into issues:  https://github.com/dotnet/runtime/pull/119824

jjonescz: > we add them during the dotnet/runtime build:

That's for the `dotnet` apphost only, right? What about any other customer app which uses apphosts, wouldn't they need entitlements to work on macOS as well?

jjonescz: > Ideally yes ([dotnet/runtime#113707](https://github.com/dotnet/runtime/issues/113707)), that's why it worked before, but we ran into issues: [dotnet/runtime#119824](https://github.com/dotnet/runtime/pull/119824)

I see, thanks!

src/sdk/src/Layout/redist/roslyn-entitlements.plist
```diff
@@ -0,0 +1,16 @@
...
+    <key>com.apple.security.cs.allow-jit</key>
+      <true/>
+    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
+      <true/>
+    <key>com.apple.security.cs.disable-library-validation</key>
+      <true/>
+    <key>com.apple.security.cs.debugger</key>
+      <true/>
+    <key>com.apple.security.get-task-allow</key>
+      <true/>
```
akoeplinger, line 14: @agocke @jkoritzinsky re-reading https://github.com/dotnet/runtime/pull/119429#issuecomment-3308045968 I'm a bit confused now whether the swift issue was because of `com.apple.security.get-task-allow` and if we'd hit the same in roslyn given we set this?

agocke: This is what's on `dotnet`, so I think these should be fine:

```
[Dict]
	[Key] com.apple.security.cs.allow-dyld-environment-variables
	[Value]
		[Bool] true
	[Key] com.apple.security.cs.allow-jit
	[Value]
		[Bool] true
	[Key] com.apple.security.cs.debugger
	[Value]
		[Bool] true
	[Key] com.apple.security.cs.disable-library-validation
	[Value]
		[Bool] true
	[Key] com.apple.security.get-task-allow
	[Value]
		[Bool] true
```

jaredpar:
> This has the existing entitlements that Roslyn has always used.

We are entitled to these entitlements.

akoeplinger:
@jkoritzinsky can you please port this and https://github.com/dotnet/dotnet/pull/3069 to main? thanks

jkoritzinsky:
Once #2425 is ported to main, then we can port these.

Better yet, @jjonescz when you port #2425 to main, can you include this PR and #3069 in the port?

PR #3069, opened 2025-10-23
Add --force option to codesign flags for roslyn

akoeplinger:
The build currently fails with:
> /Users/runner/work/1/s/src/sdk/artifacts/bin/redist/Release/net10.0/Roslyn/bincore/csc: is already signed

Follow-up to https://github.com/dotnet/dotnet/pull/3052

This makes the codesign command now match what we do in dotnet/runtime: https://github.com/dotnet/runtime/blob/9bcdd873eb4b93934ae34328845599f63253d32a/eng/native/functions.cmake#L736


jjonescz (approved):

akoeplinger:
force-merging since this isn't tested in public and the official build is broken.

PR #3089, opened 2025-10-24
[release/10.0.2xx] Reset VMRs roslyn to main

dkurepa:
Current FFs are failing because the branches are detached. Resetting roslyn back to where it should be to unblock this


dkurepa:
I forgot this existed an opened up a new PR. Will close this one, since the new one has a fresher commit it's flowing https://github.com/dotnet/dotnet/pull/3162

PR #3118, opened 2025-10-27
Merge 10.0.1xx to 10.0.2xx

mthalman:
Merges in the missing changes from 10.0.1xx. This follows the process described in https://github.com/dotnet/arcade-services/issues/5378.

### DIFF
--- a/eng/pipelines/templates/variables/vmr-build.yml
+++ b/eng/pipelines/templates/variables/vmr-build.yml
@@ -64,7 +64,7 @@ variables:
     value: 'rtm'
 - ${{ else }}:
   - name: brandingType
-    value: 'rtm'
+    value: 'default'
 
 ## Do not set this when building source only to avoid including telemetry in the bootstrap artifacts
 ## handed off to distro partners.


ellahathaway (approved):

eng/pipelines/templates/variables/vmr-build.yml
```diff
@@ -64,7 +64,7 @@ variables:
     value: 'rtm'
 - ${{ else }}:
   - name: brandingType
-    value: 'default'
+    value: 'rtm'
```
ViktorHofer: Do we already want stable builds for release/10.0.2xx? cc @mmitche

mthalman: That doesn't set stable builds. It only defaults the branding value if `desiredFinalVersionKind` isn't set to one of the listed values. I specifically excluded the 1xx change to the pipeline file which sets it to stable branding.

mthalman: > It only defaults the branding value if desiredFinalVersionKind isn't set to one of the listed values.

Hmm... which does seem to be the case for the PR pipeline. So yeah, I'll have to revert this.

mthalman:
@premun - Do I want to use a merge commit here? Is that safe in the context of the VMR?

premun (approved):
@mthalman that's fine until it touches the `source-manifest.json`.
I think this can be merged using a merge commit.

PR #3134, opened 2025-10-28
[release/10.0.1xx] Rebranding for 10.0.1

mmitche:


mmitche:
@wtgodbe I think there are some things to do for Aspnetcore, correct?

wtgodbe:
>  I think there are some things to do for Aspnetcore, correct?

In the past we haven't been able to update the patch to `1` until RTM is actually released, since servicing builds want to have access to the new checked-in platformManifest/packageOverrides/baselines/etc. I can probably finagle it to get it working without that for now - how urgent is this?

ViktorHofer (approved):

mmitche:
> > I think there are some things to do for Aspnetcore, correct?
> 
> In the past we haven't been able to update the patch to `1` until RTM is actually released, since servicing builds want to have access to the new checked-in platformManifest/packageOverrides/baselines/etc. I can probably finagle it to get it working without that for now - how urgent is this?

I'd say I'd like to get this merged in the next day or two. We do have the platform manifest from the public builds...good enough?

wtgodbe:
Probably, I'll push a commit

src/sdk/eng/Versions.props
```diff
@@ -7,7 +7,7 @@
     <VersionMajor>10</VersionMajor>
     <VersionMinor>0</VersionMinor>
     <VersionSDKMinor>1</VersionSDKMinor>
-    <VersionFeature>00</VersionFeature>
+    <VersionFeature>1</VersionFeature>
```
JoeRobich:
```suggestion
    <VersionFeature>01</VersionFeature>
```

ViktorHofer:
@wtgodbe why is checking the shared framework files necessary in aspnetcore? We don't do this in runtime or windowsdesktop.

wtgodbe:
> @wtgodbe why is checking the shared framework files necessary in aspnetcore? We don't do this in runtime or windowsdesktop.

Sorry, was OOF last week - we rely on reference resolution to put transitive refs from runtime into the SharedFx, when they're not already in the runtime SharedFx. If there's ever a change in the surface area of the SharedFx/Targeting Pack, we want to catch it & make sure it was on purpose.

PR #3146, opened 2025-10-29
[source build] cloak the spectre-console/docs folder

jonathanpeppers:
Context: https://github.com/dotnet/source-build-reference-packages/pull/1447
Context: https://github.com/dotnet/source-build-reference-packages/pull/1287#issuecomment-3046537869

The docs folder contains binary files, that are not needed for building from source.

### DIFF
--- a/src/source-mappings.json
+++ b/src/source-mappings.json
@@ -89,7 +89,7 @@
             "name": "source-build-reference-packages",
             "defaultRemote": "https://github.com/dotnet/source-build-reference-packages",
             "exclude": [
-                // Contains undesired binaries
+                // Contains disallowed binaries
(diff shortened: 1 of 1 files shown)


src/source-mappings.json
```diff
@@ -87,7 +87,11 @@
         },
         {
             "name": "source-build-reference-packages",
-            "defaultRemote": "https://github.com/dotnet/source-build-reference-packages"
+            "defaultRemote": "https://github.com/dotnet/source-build-reference-packages",
+            "exclude": [
+                // Contains undesired binaries
```
MichaelSimons: I would word this stronger.  They are more than undesired, they are not allowed.

```suggestion
                // Contains disallowed binaries
```

MichaelSimons:
Merging ahead of validation as it won't actually validate the source-mappings.

PR #3181, opened 2025-10-31
Use --with-shared-components arg for source build

mthalman:
Fixes dotnet/source-build#5396

Updates the VMR pipeline to use the `--with-shared-components` arg when calling the `build.sh` script for source-only builds in the context of non-1xx configurations. This ensures that we're validating the same workflow that we expect distro maintainers to use when building the VMR.

This change uncovered some bugs in this build path
- Validation exists that will result in an error if you're source-building the 2xx branch without providing `--with-shared-components`. That was triggering in the prep script and when running tests, two scenarios when that arg isn't provided.
- Symbols from the shared components weren't correctly handled, resulting in the shared component symbols to be included in the artifacts layout which messed things up.


eng/init-source-only.proj
```diff
@@ -104,7 +108,7 @@
     <!-- Extract all symbols from the symbols tarball to the intermediate symbols root directory -->
     <Exec Command="tar -xzf $(BaseIntermediateOutputPath)/$(SourceBuiltSymbolsAllTarballName)-*$(ArchiveExtension)"
           WorkingDirectory="$(_SharedComponentsSymbolsDir)" />
-
+
```
MichaelSimons:
```suggestion

```

MichaelSimons (approved):
Good to see this flushed out some bugs.

PR #3202, opened 2025-11-03
Remove secure@microsoft.com from security reporting

sirredbeard:
The method of reporting suspected vulnerabilities via secure@microsoft.com has been deprecated, all vulnerabilities must now be reported via the MSRC portal.


akoeplinger:
See discussion in https://github.com/dotnet/runtime/pull/121303

akoeplinger (approved):

PR #3219, opened 2025-11-04
Support download of 1xx from internal storage

mthalman:
Fixes dotnet/source-build#5342

- Updates `download-source-built-archive.sh` to take the encoded storage key that can be used to access the internal blob storage location.
  - The internal URL will be used whenever the artifact version contains the `servicing` label.
- Updates the pipeline to pass in the storage key by calling the `enable-internal-runtimes.yml` template from Arcade which dynamically creates a SAS token and sets it to the `dotnetbuilds-internal-container-read-token-base64` variable.

### DIFF
--- a/eng/download-source-built-archive.sh
+++ b/eng/download-source-built-archive.sh
@@ -96,7 +96,8 @@ function DownloadArchive {
 
   archiveVersion="${versionDelimiter}${archiveVersion}${versionDelimiter}"
 
-  # Build array of base URLs to try in order
+  # Build array of base URL entries to try in order
+  # Each entry is in the format "url" or "url=storageKey"
   local baseUrls=()
   
   # Assets for the default RID are hosted in builds.dotnet.microsoft.com first
@@ -109,7 +110,7 @@ function DownloadArchive {
   
   # Try internal CI location if storage key is provided
   if [[ -n "$storageKey" ]]; then
-    baseUrls+=("https://ci.dot.net/internal/source-build")
+    baseUrls+=("https://ci.dot.net/internal/source-build=$storageKey")
   fi
 
   # Determine the archive filename based on property name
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -415,18 +415,31 @@ jobs:
         DownloadArchive "source-built SDK" "PrivateSourceBuiltSdkVersion" false "${{ parameters.artifactsRid }}" "$(sourcesPath)/prereqs/packages/archive/"
       displayName: Download Previously Source-Built SDK
 
+  - ${{ if eq(variables['System.TeamProject'], 'internal') }}:
```
mmitche: This should be unnecessary. The template doesn't do anything if the project is `public`

eng/download-source-built-archive.sh
```diff
@@ -89,16 +90,28 @@ function DownloadArchive {
...
 
   archiveVersion="${versionDelimiter}${archiveVersion}${versionDelimiter}"
 
   baseUrl="https://builds.dotnet.microsoft.com/dotnet/source-build"
+  local useStorageKey=false
   if [[ "$useCILocation" == true ]]; then
-    baseUrl="https://ci.dot.net/public/source-build"
+    # Determine CI location based on version suffix
+    if [[ $isServicingArchiveVersion == true ]]; then
+      baseUrl="https://ci.dot.net/internal/source-build"
```
mmitche: Is it correct to assume that the SB archive will always be on `internal`? Typically, what we've done is pass in alternate storage paths+creds to the download functionality. The download code then attempts to first find the asset on public storage, falling back to internal with auth if it gets a 404.

I prefer this approach in general because it has fewer issues at boundary times.

eng/download-source-built-archive.sh
```diff
@@ -89,16 +90,28 @@ function DownloadArchive {
     versionDelimiter="-"
   fi
 
+  local isServicingArchiveVersion=false
+  if [[ "$archiveVersion" == *"servicing"* ]]; then
+    isServicingArchiveVersion=true
```
mmitche: I wouldn't assume location based on servicing or not. An RC might be internal too, for instance. Instead, I'd use an approach where we look in a default public location first, then go to an alternate provided location after that.

eng/pipelines/templates/jobs/vmr-build.yml
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/3219/commits/53b12486a8d3a8b9dfa84eb3e4b06afe91834823

eng/download-source-built-archive.sh
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/3219/commits/b69b671c47a3ab1e67b4a9d62875872730195215

eng/download-source-built-archive.sh
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/3219/commits/b69b671c47a3ab1e67b4a9d62875872730195215

eng/download-source-built-archive.sh
```diff
@@ -89,39 +94,77 @@ function DownloadArchive {
...
+  if [[ $artifactsRid == $defaultArtifactsRid ]]; then
+    baseUrls+=("https://builds.dotnet.microsoft.com/dotnet/source-build")
+  fi
+  
+  # Always try public CI location
+  baseUrls+=("https://ci.dot.net/public/source-build")
+  
+  # Try internal CI location if storage key is provided
+  if [[ -n "$storageKey" ]]; then
+    baseUrls+=("https://ci.dot.net/internal/source-build")
```
mmitche: nit: You can avoid the string compare below to [ci.dot.net/internal](https://github.com/dotnet/dotnet/pull/3219/files#diff-45f7283695533c504fd2655f72aabfc6b924e17e4603028153d4aef5e5aa7f8aR139) by updating the baseUrls array to be a series of key and optional value pairs (e.g. "https://ci.dot.net/public/source-build" and "https://ci.dot.net/internal/source-build=$storageKey")

PR #3254, opened 2025-11-06
Use stable sdk verison in inner artifact directory

ellahathaway:
Fixes https://github.com/dotnet/release/issues/1709

The GitHub source artifacts associated with an SDK release use the stable SDK version in the inner directory of the source artifacts. In other words, the inner directory of the artifacts includes the SDK version tag without the `v` tag prefix. 

The source artifacts generated in our builds must have the same inner directory name as the GitHub source artifacts. Otherwise, the hashes of the artifacts do not match and the signatures are invalid.

### DIFF
--- a/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
+++ b/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
@@ -28,12 +28,12 @@ public class CreateSourceArtifacts : BuildTask
     public required string ArtifactName { get; init; }
 
     /// <summary>
-    /// Sdk version
+    /// The stable sdk version
     /// Should match the sdk tag associated with a release
     /// Used for artifact prefixing
     /// </summary>
     [Required]
-    public required string SdkVersion { get; init; }
+    public required string SdkStableVersion { get; init; }
 
     /// <summary>
     /// The build/non-stable sdk version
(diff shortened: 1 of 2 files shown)


eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
```diff
@@ -28,10 +28,19 @@ public class CreateSourceArtifacts : BuildTask
...
 
     /// <summary>
-    /// Version of the artifact to generate
+    /// Sdk version
+    /// Should match the sdk tag associated with a release
+    /// Used for artifact prefixing
     /// </summary>
     [Required]
-    public required string ArtifactVersion { get; init; }
+    public required string SdkVersion { get; init; }
```
mthalman: Consider being explicit about the version types and naming this `SdkStableVersion`.

mthalman (approved):

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifacts.cs
ellahathaway: Fixed in https://github.com/dotnet/dotnet/pull/3254/commits/40a83195649efa6c73696125528bdb19cbcb10db

ellahathaway:
I verified that these changes produce the same GA source artifacts as the GitHub generated source artifacts in [maestro-auth-test](https://github.com/maestro-auth-test/private-dotnet/releases/tag/v10.0.100).

mthalman (approved):

PR #3305, opened 2025-11-11
Reference 10.0 GA version for 1xx dependency

mthalman:


mthalman:
@mmitche - The tests are failing here because version 10.0.100-rtm.25523.111 of `Microsoft.NET.Runtime.WebAssembly.Sdk` doesn't exist in the public `dotnet10` feed. That represents the 10.0 GA release of that package. Are these not meant to be pushed to the public feed on release day?

mmitche:
Working on it.

mmitche:
@mthalman The webassembly SDK is on nuget.org and branded 10.0.0. So maybe something else is going on here?

mthalman:
Ok, so the logic can't assume that it will be the non-stable version, which is what `MicrosoftNETSdkVersion` represents. You can't really assume that if the version contains a label of `rtm` or `servicing` that it should be stable versioned either though, right? Or could you? How are those versioned during servicing for internal branches for example?

Here's the relevant logic: https://github.com/dotnet/dotnet/blob/63901a5c8a4aeae1c7f3f9514ed242421683599f/test/wasm-dependencies.proj#L2-L11

mmitche:
> Ok, so the logic can't assume that it will be the non-stable version, which is what `MicrosoftNETSdkVersion` represents. You can't really assume that if the version contains a label of `rtm` or `servicing` that it should be stable versioned either though, right? Or could you? How are those versioned during servicing for internal branches for example?
> 
> Here's the relevant logic:
> 
> https://github.com/dotnet/dotnet/blob/63901a5c8a4aeae1c7f3f9514ed242421683599f/test/wasm-dependencies.proj#L2-L11

I'd say it's generally not a good idea.  Why not just add a couple dependencies on 1xx runtime packages? Then there's no version manipulation.

mthalman:
> Why not just add a couple dependencies on 1xx runtime packages? Then there's no version manipulation.

I'm not sure what you mean by this.

eng/Version.Details.xml
```diff
@@ -14,9 +14,9 @@
       <Uri>https://github.com/dotnet/arcade-services</Uri>
       <Sha>4aa8e8c35aaf3cbdfc387d28af3366fcd9125bc8</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NET.Sdk" Version="10.0.100-rc.2.25502.107">
+    <Dependency Name="Microsoft.NET.Sdk" Version="10.0.100-rtm.25523.111">
       <Uri>https://github.com/dotnet/dotnet</Uri>
-      <Sha>b502b6eeec0db06720ead7fd9570befa39a6b2f7</Sha>
+      <Sha>b0f34d51fccc69fd334253924abd8d6853fad7aa</Sha>
     </Dependency>
```
mmitche, line 20:
```suggestion
    </Dependency>
    <Dependency Name="Microsoft.NETCore.App.Ref" Version="10.0.0">
      <Uri>https://github.com/dotnet/dotnet</Uri>
      <Sha>b0f34d51fccc69fd334253924abd8d6853fad7aa</Sha>
    </Dependency>
```

Example...add this to here (and to `Version.Details.props`). When you flow the 1xx build to 2xx, you'll get the runtime version information here.

mmitche: Then this line:

https://github.com/dotnet/dotnet/blob/63901a5c8a4aeae1c7f3f9514ed242421683599f/test/wasm-dependencies.proj#L5

Just becomes
`<_RuntimeVersion1xx>$(MicrosoftNETCoreAppRefVersion)</_RuntimeVersion1xx> `

PR #3335, opened 2025-11-13
Add sourcelink release/10.0.2xx to the VMR

ViktorHofer:
`darc vmr add-repo https://github.com/dotnet/sourcelink:3b5251dd0be65990d9d0572bfb710f51e6c21c3a`

https://github.com/dotnet/sourcelink/commit/3b5251dd0be65990d9d0572bfb710f51e6c21c3a

### DIFF
--- a/src/source-manifest.json
+++ b/src/source-manifest.json
@@ -67,10 +67,10 @@
       "commitSha": "3b5251dd0be65990d9d0572bfb710f51e6c21c3a"
     },
     {
-      "barId": 290633,
+      "barId": 290817,
       "path": "templating",
       "remoteUri": "https://github.com/dotnet/templating",
-      "commitSha": "0a93029ebf9ea28397b4291ad83ae7078134c556"
+      "commitSha": "ec5f0e074a6aeffe17f15a0ab49f3fd2122add48"
     },
     {
       "barId": 285868,
--- a/src/templating/NuGet.config
+++ b/src/templating/NuGet.config
@@ -4,7 +4,7 @@
     <clear />
     <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
     <!--  Begin: Package sources from dotnet-dotnet -->
-    <add key="darc-pub-dotnet-dotnet-e17b0d0" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-dotnet-e17b0d08/nuget/v3/index.json" />
+    <add key="darc-pub-dotnet-dotnet-f4701e0" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-dotnet-f4701e04/nuget/v3/index.json" />
     <!--  End: Package sources from dotnet-dotnet -->
     <!--  Begin: Package sources from dotnet-runtime -->
     <!--  End: Package sources from dotnet-runtime -->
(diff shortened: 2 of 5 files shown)


src/source-mappings.json
```diff
@@ -1,79 +1,65 @@
...
 {
-    "thirdPartyNoticesTemplatePath": "eng/THIRD-PARTY-NOTICES.template.txt",
-
-    // These defaults are added to all mappings unless `ignoreDefaults: true` is specified
-    // When no "include" filter is specified, "**/*" is used
-    // The default filters do not apply to submodules
-    // Only filters which start with submodule's path are applied when syncing submodules
-    "defaults": {
-        "defaultRef": "main"
+  "defaults": {
```
premun: @ViktorHofer I guess it manipulates the mappings and serializes them back into the JSON killing all the comments etc. Maybe we can keep the original file and add this manually?

ViktorHofer: Sure, it's annoying but not the end of the world.

premun: Or we can leave it without comments too, not a big deal

src/source-manifest.json
```diff
@@ -61,10 +61,16 @@
...
     },
     {
-      "barId": 279652,
+      "barId": null,
+      "path": "sourcelink",
+      "remoteUri": "https://github.com/dotnet/sourcelink",
+      "commitSha": "3b5251dd0be65990d9d0572bfb710f51e6c21c3a"
+    },
+    {
+      "barId": 290633,
```
premun:
```suggestion
      "barId": 279652,
```

src/source-manifest.json
```diff
@@ -61,10 +61,16 @@
...
       "path": "templating",
       "remoteUri": "https://github.com/dotnet/templating",
-      "commitSha": "fcc5f2933e4ad7fbfc85a9c4e78d9bfa3e967eb7"
+      "commitSha": "0a93029ebf9ea28397b4291ad83ae7078134c556"
```
premun:
```suggestion
      "commitSha": "fcc5f2933e4ad7fbfc85a9c4e78d9bfa3e967eb7"
```

ViktorHofer:
@premun I assume the THIRD-PARTY-NOTICES.TXT change is some form of bug of `dotnet vmr add-repo`. Should we file an issue for that?

premun:
@ViktorHofer it seems the file was not added in the 2xx branch when it was created. This tool re-generates it each time so that it collects TPNs from all repos and concatenates them. This is by design.

Or what is the problem you see there?

ViktorHofer:
The file contains content of all repos, even the ones removed in 2xx, like runtime: https://github.com/dotnet/dotnet/blob/2ce77758aac86fd0bb36215e2d58be35f904912a/THIRD-PARTY-NOTICES.txt#L3426

Is that intentional?

premun:
Ah, okay, that's a bug then. Will have a look - no idea where that content came from then.

I guess it's not a big deal and we can merge here?

ViktorHofer:
Maybe because the current one in release/10.0.2xx already has all that content and the command doesn't discard that?

premun:
@ViktorHofer I found out that the problem was that the TPN was already in the branch and contained all the 1xx repos. When I deleted it, it populated only the right repos.

premun:
@ViktorHofer I think we can merge now

ViktorHofer:
Perfect, thanks for the help.

Enabling auto-merge. I can't approve my own PR though :D

dkurepa:
I re enabled the backflow subscription

PR #3348, opened 2025-11-13
Set custom PSB dir for SB prep

mthalman:
Fixes the source build prep script to set the `CustomPreviouslySourceBuiltPackagesPath` MSBuild property to the directory passed via the `--with-packages` arg.

Fixes dotnet/source-build#5410


MichaelSimons:
Would fixing https://github.com/dotnet/source-build/issues/4427 have caught this regression?  I am wondering if we should emphasis fixing that?

mthalman:
> Would fixing [dotnet/source-build#4427](https://github.com/dotnet/source-build/issues/4427) have caught this regression? 

Yes

PR #3359, opened 2025-11-14
Improve log message

BenjaminBrienen-CanonPP:
Fixes the plural to be singular when it is 1 hour


akoeplinger:
Thank you, but changes to the individual repos in src/ should primarily be made there (see [README](https://github.com/dotnet/dotnet#where-to-make-changes)).

I'd suggest you open this PR in https://github.com/dotnet/templating instead.

PR #3403, opened 2025-11-18
[release/10.0.1xx] Share the Profile information for WindowsDesktop between runtime and targeting packs

ericstj:
Fixes https://github.com/dotnet/wpf/issues/11261, https://github.com/dotnet/windowsdesktop/issues/5405

## Customer Impact

- [x] Customer reported

Cannot use System.Drawing.Common from a WPF-only self-contained application.

## Regression

- [X] Yes - regressed in .NET 10.0 with package pruning.  Previous fixes addressed this for compilation, and runtime for shared framework apps, but not self contained.

## Testing
Manually applied fix to data files and tested.  Made fix to runtime pack and ensured it matched manually modified files.  [Diffed](https://gist.github.com/ericstj/f6d4574b5672d13260576b7665042214/revisions) the GA files with fix files to ensure nothing else changed.  Refactored how this data is defined to avoid this class of issue in the future.

## Risk
Low.  Diffs reduce risk so that we can say that this is only modifying what we want.  This will mean that self-contained WPF apps may see a size increase but they can mitigate that with linking / removing unused files.


ericstj:
Backport of https://github.com/dotnet/dotnet/pull/3402

ericstj:
Approved over mail.

PR #3424, opened 2025-11-18
Add MSBuild workflow for building projects

408apple:


ViktorHofer:
What's the reason for this change?

PR #3443, opened 2025-11-19
Reset to runtime release/10.0

dkurepa:
Essentially forward flows the runtime release/10.0 branch

Diff: https://github.com/dotnet/runtime/compare/765b715223fe612d0e982d9ff3d88ed88a059a65...138d5fe8f32e954a3a834be6360a834688130d4f


src/runtime/eng/Publishing.props
```diff
@@ -5,11 +5,6 @@
     <AutoGenerateSymbolPackages>false</AutoGenerateSymbolPackages>
   </PropertyGroup>
 
-  <!-- remove the manifest packages, they are built in the SDK repo now -->
-  <ItemGroup>
-    <Artifact Remove="$(ArtifactsShippingPackagesDir)*.Manifest-*.nupkg" />
-  </ItemGroup>
-
```
ViktorHofer, line 12: @akoeplinger we still need this block, right? Looks like this never backflowed.

dkurepa: The commit I based the reset on was too now, it should be the one right after the last backflow, my bad

PR #3451, opened 2025-11-19
[release/10.0.1xx] Bump patch version to 10.0.2

ViktorHofer:
Increment patch version for 11 repositories on release/10.0.1xx

Updated repositories:
- aspnetcore: 1 → 2
- command-line-api: 1 → 2
- efcore: 1 → 2
- emsdk: 1 → 2
- runtime: 1 → 2
- windowsdesktop: 1 → 2
- wpf: 1 → 2
- winforms: 1 → 2
- sdk: 1 → 2
- templating: 101 → 102
- fsharp: 101 → 102
- symreader: 1 → 2
- sourcelink: 1 → 102
- xdt: 1 → 2

Sets prerelease label to 'servicing' and iteration to empty string.

PR #3469, opened 2025-11-20
Clean-up Versions.props files

ViktorHofer:
These clean-ups are possible now that automation updates Version.Details.props instead and that we do centralized branding via the VMR


src/sourcelink/eng/Versions.props
```diff
@@ -1,19 +1,23 @@
...
-  <Import Project="Version.Details.props" Condition="Exists('Version.Details.props')" />
+
+  <Import Project="Version.Details.props" />
 
   <PropertyGroup>
-    <!-- This repo version -->
-    <VersionPrefix>11.0.0</VersionPrefix>
-    <PreReleaseVersionLabel>beta</PreReleaseVersionLabel>
-    <!-- Opt-in repo features -->
+    <VersionPrefix>11.0.100</VersionPrefix>
```
akoeplinger, line 6: I know you answered this is intentional in https://github.com/dotnet/dotnet/pull/3451#discussion_r2543555214 but I'd really appreciate doing this outside of a cleanup PR next time

PR #3508, opened 2025-11-22
Fixing wrong enum cast of (-1) on Linux, where enum was automatically handled as a uint8_t

bebenlebricolo:
While trying to build for my linux machine using Arch's PKGBUILD, I noticed that we had some C++ errors while building the coreclr project.

The issue specifically is an occurrence of a negative value (-1) being casted as an enum, which defaults to a uint8_t type in this case.

I haven't modified the value itself (-1) in the original version of this PR, but I believe the real value should be positive instead to keep the uint8_t nature of the enum.

## PKGBUILD script
The PKGBUILD version that showcased this issue is the version pointed by [https://gitlab.archlinux.org/bebenlebricolo/dotnet-core/-/commit/e981dc062a365961047ba03a091e470dbec221ae](https://gitlab.archlinux.org/bebenlebricolo/dotnet-core/-/commit/e981dc062a365961047ba03a091e470dbec221ae).

### DIFF
--- a/src/runtime/src/coreclr/debug/di/rspriv.h
+++ b/src/runtime/src/coreclr/debug/di/rspriv.h
@@ -6397,7 +6397,7 @@ private:
     // Lazily initialized.
     EXCEPTION_RECORD *  m_pExceptionRecord;
 
-    static const CorDebugUserState kInvalidUserState = CorDebugUserState(-1);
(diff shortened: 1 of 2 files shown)


bebenlebricolo:
Noticed it was mostly fixed in releases/10.0.1xx by changing the type CorDebugUserState by ... an int in the rspriv.h
So no need to proceed further

Issue #3345, opened 2025-11-13
Consider Default RTM Branding in Release Build Entry-Point


ellahathaway:
Related to https://github.com/dotnet/release/issues/1707

We should consider defaulting to RTM branding not just in the YML files (as done for 10.0 GA), but also in the build entry-point. This would ensure that the default experience when building from a release branch is an RTM build. However, RTM branding should not be the default for repo builds, to avoid unintentionally generating stable packages.

mmitche:
Agreed. This does mean that to switch back to non-stable branding, you'd end up passing `-branding default`. This is a little non-intuitive. So I think we'd do something like:

- Add a `branding prerelease` switch
- Make `branding default` == `branding rtm` for release/10.0.1xx

That leads to the following branding switches:
- **prerelease** - Force non-stable
- **rtm** - Stable RTM/Servicing builds
- **preview** - Stable preview builds
- **default** - Branch default.

I don't love the naming too much...`prerelease` and `preview` are too similar.

ViktorHofer:
OK looking at the [Arcade controls](https://github.com/dotnet/arcade/blob/e29f11a331730bd4f51024ba83fa067d2ab80ca9/Documentation/CorePackages/Versioning.md?plain=1#L223), there's no distinction between preview, rc and release in terms of the `DotNetFinalVersionKind` property. I don't think we need to control the label from the VMR centrally? We want the inner repos to control that.

That would give us only the following values:
- default (or unspecified will use VMR defaults for release branch builds or otherwise inner repo defaults)
- prerelease (non-stable, will produce assets suffixed with '.final')
- release (stable)

ViktorHofer:
Do we need the distinction between stable previews and release / servicing? I would have hoped that all three would use the same `DotNetFinalVersionKind` value: release. So that stable previews would have the same version syntax without a suffix, i.e. `11.0.0-preview1`. Why would we want the `.final` suffix?

If we could converge those we could have the following (I think it makes sense to have the `.final` suffix as an option when needed):
- prelease (Produces assets suffixed with '.final')
- release (Produces assets without a suffix)
- default (or unspecified uses VMR defaults for release branch builds or otherwise repo defaults)
- repodefault (Ignores VMR defaults, uses repo defaults)

I expect the branding switch in the YML or in the build entry point scripts to not be used much if the default reflects the correct value. I can imagine that we would use them for manual builds (i.e. for partner hand off) with a specific branding.

[...]

PR #3288, opened 2025-11-10
Add note about stable version strings

ellahathaway:
Related to https://github.com/dotnet/release/issues/1707

PR #3574, opened 2025-11-26
Allow default branding to be set by the VMR orchestrator

ViktorHofer:
Fixes https://github.com/dotnet/dotnet/issues/3345

### DIFF
--- a/eng/pipelines/official.yml
+++ b/eng/pipelines/official.yml
@@ -45,7 +45,7 @@ parameters:
   default: 'Default (uses VMR branding defaults for release branch builds or otherwise repo branding defaults)'
   values:
     - Default (uses VMR branding defaults for release branch builds or otherwise repo branding defaults)
-    - RepoDefault (uses the repo branding defaults)
+    - Repo default (uses the repo branding defaults)
     - Unstable (produces assets with the repo specified branding suffix)
     - Preview (produces assets with a '.final' branding suffix)
     - Release (produces assets without a branding suffix)
--- a/eng/pipelines/templates/variables/vmr-build.yml
+++ b/eng/pipelines/templates/variables/vmr-build.yml
@@ -53,7 +53,7 @@ variables:
   - name: ibcEnabled
     value: false
 
-- ${{ if eq(parameters.desiredFinalVersionKind, 'RepoDefault (uses the repo branding defaults)') }}:
+- ${{ if eq(parameters.desiredFinalVersionKind, 'Repo default (uses the repo branding defaults)') }}:
   - name: brandingType
     value: 'repodefault'
 - ${{ elseif eq(parameters.desiredFinalVersionKind, 'Unstable (produces assets with the repo specified branding suffix)') }}:


ViktorHofer:
@mmitche the PR is now ready. I also added a clear notice for distro maintainers that explicitly passed `-branding rtm` in for .NET 10 as they will need to react:

```
~/git/dotnet$ ./build.sh -branding rtm
Unsupported branding type 'rtm'.
The allowed values are repodefault, unstable, preview or release.

IMPORTANT: If you previously passed in the '-branding rtm' argument for .NET 10, remove it entirely. The default is now set correctly (VMR branding defaults for release branch builds or otherwise repo branding defaults).
```

ViktorHofer:
@ellahathaway regarding https://dev.azure.com/dnceng/internal/_workitems/edit/8757, distro maintainers don't need to pass that argument in anymore as the default is now correct. Someone would only want to pass this argument in explicitly when wanting to change the default (i.e. testing stable builds in the RC timeframe) which is quite uncommon. Do we want to send out another announcement or update the existing one? We should wait though until the backport PRs are merged into 10.0.1xx and 10.0.2xx.

mthalman:
> We should wait though until the backport PRs are merged into 10.0.1xx and 10.0.2xx.

I don't think we should backport this to 10.0.1xx as that would be a breaking change.

ViktorHofer:
`repodefault` is the default behavior, when not stabilizing which means the repository decides the `DotNetFinalVersionKind` value. This is mainly used in the main branch and during RC1 before we switch to internal and stabilized builds.

When stabilizing for previews, `RepoDotNetFinalVersionKind` will get set to `preview` and for GA and servicing, the value will be `release`. These are different options because a preview still has a branding suffix (.final) vs release which doesn't have one.

`unstable` will be used to force non-stable assets in release branches. https://github.com/dotnet/dotnet/issues/3345 has more details on that.

> I don't think we should backport this to 10.0.1xx as that would be a breaking change.

I really think we should backport to 1xx. The breaking change will be clearly visible (the build fails) and the error message indicates how to react. I want to include this as part of many UX build entry point changes that we did up to .NET 10 so that distro partners don't need to condition based on 1xx vs 2xx/main.  cc @tmds for input

mthalman:
cc @dotnet/distro-maintainers for visibility

MichaelSimons:
> > I don't think we should backport this to 10.0.1xx as that would be a breaking change.
> 
> I really think we should backport to 1xx. The breaking change will be clearly visible (the build fails) and the error message indicates how to react. I want to include this as part of many UX build entry point changes that we did up to .NET 10 so that distro partners don't need to condition based on 1xx vs 2xx/main. cc @tmds for input

@ViktorHofer - would it be feasible to support a migration?  What I mean by that is for one servicing release we support both the new and old options so that folks can migrate at their convenience. Then the next servicing release we make the actual breaking change and remove the old options?  I realize the transition period may be confusing with the option names but we could choose to support the old options as undocumented.

build.sh
```diff
@@ -15,7 +15,11 @@ usage()
   echo "  --rid, --target-rid <value>       Overrides the rid that is produced by the build. e.g. alpine.3.18-arm64, fedora.37-x64, freebsd.13-arm64, ubuntu.19.10-x64"
   echo "  --os, --target-os <value>         Target operating system: e.g. linux, osx, freebsd. Note: this is the base OS name, not the distro"
   echo "  --arch, --target-arch <value>     Target architecture: e.g. x64, x86, arm64, arm, riscv64"
-  echo "  --branding <preview|rtm|default>  Specify versioning for shipping packages/assets. 'preview' will produce assets suffixed with '.final', 'rtm' will not contain a pre-release suffix. Default or unspecified will use VMR repo defaults."
+  echo "  --branding                        Specify the branding suffix for shipping packages/assets. By default uses VMR branding defaults for release branch builds or otherwise repo branding defaults."
```
MichaelSimons: > By default uses VMR branding defaults for release branch builds or otherwise repo branding defaults

I feel like this deserves a more thorough explanation.  I don't see distro maintainers as being able to reason over what "VMR branding defaults" would mean for release branches.

ellahathaway:
> distro maintainers don't need to pass that argument in anymore as the default is now correct. Someone would only want to pass this argument in explicitly when wanting to change the default (i.e. testing stable builds in the RC timeframe) which is quite uncommon. Do we want to send out another announcement or update the existing one? We should wait though until the backport PRs are merged into 10.0.1xx and 10.0.2xx.

The announcement templates referenced in that work item served as a temporary workaround to address [dotnet/release#1707](https://github.com/dotnet/release/issues/1707) for the 10.0 GA release. As a result, the announcement templates were not permanently updated to include information about using the `--branding` argument. Since the long-term fix was implemented in [dotnet/dotnet#3288](https://github.com/dotnet/dotnet/pull/3288), we now need to update the documentation accordingly. Once that's done, everything should be in order.

build.sh
ViktorHofer: "VMR branding defaults for release branch builds" means whatever is set in eng/Versions.props in the VMR. I will try to clarify that in the description.

The most important part here is that the VMR now has the correct default and that distro partners don't need to pass a `--branding` switch in anymore, unless they want to override the default.

ViktorHofer:
> we now need to update the documentation accordingly. Once that's done, everything should be in order.

Ok. I was saying that with this change, distro partners don't need to pass `--branding` in anymore as the default is now correct. I think that means that we don't need to follow-up anywhere. This PR already updates the README.md

build.sh
ViktorHofer: I will not add the `rtm` value back for main or 10.0.2xx but only in 10.0.1xx. So not part of this PR.

omajid:
With my distro maintainer hat on, I never really understood the role of the branding argument. Requiring `--branding rtm` to build the right thing for 10.0 has been a bit surprising. I know some others were building wrong assets until they found about this argument and its purpose.

If this PR is switching to automatically building the right thing, then a big :+1: from me. I am even okay with the "build break" for 10.0 as this makes it easier to build the right thing correctly by default, and the error on using `--branding rtm` (the previous recommended value) seems clear and actionable.

ViktorHofer:
> If this PR is switching to automatically building the right thing, then a big 👍 from me

Yes, that's what this PR does 👍 While there's still the `--branding` switch, you would only want to pass that in for i.e. manual builds (locally or as a pipeline parameter) to use a different branding than what's chosen as the default for the given branch/SHA.

eng/pipelines/templates/variables/vmr-build.yml
```diff
@@ -53,18 +53,21 @@ variables:
   - name: ibcEnabled
     value: false
 
-- ${{ if eq(parameters.desiredFinalVersionKind, 'Default (repo specifies)') }}:
+- ${{ if eq(parameters.desiredFinalVersionKind, 'RepoDefault (uses the repo branding defaults)') }}:
```
mmitche:
```suggestion
- ${{ if eq(parameters.desiredFinalVersionKind, 'Repo default (uses the repo branding defaults)') }}:
```

eng/pipelines/official.yml
```diff
@@ -42,11 +42,13 @@ parameters:
...
   displayName: 'Final version kind'
   type: string
-  default: 'Default (repo specifies)'
+  default: 'Default (uses VMR branding defaults for release branch builds or otherwise repo branding defaults)'
   values:
-    - Default (repo specifies)
-    - Stable Preview
-    - Stable Final
+    - Default (uses VMR branding defaults for release branch builds or otherwise repo branding defaults)
+    - RepoDefault (uses the repo branding defaults)
```
mmitche:
```suggestion
    - Repo default (uses the repo branding defaults)
```

eng/Versions.props
```diff
@@ -1,12 +1,26 @@
...
+    <PreReleaseVersionIteration></PreReleaseVersionIteration>
+
+    <!-- 
+      Allowed values: 'repodefault', 'unstable', 'preview', 'release'
+      'repodefault' uses the repo branding defaults.
+      'unstable' produces assets with the repo specified branding suffix.
+      'preview' produces assets with a '.final' branding suffix.
+      'release' produces assets without a branding suffix.
+    -->
+    <RepoDotNetFinalVersionKind>repodefault</RepoDotNetFinalVersionKind>
```
mmitche, line 18: So this will get changed to "release" when this gets ported to 10.0?

ViktorHofer: Yes

PR #3578, opened 2025-11-26
Reset nuget.client 2xx

dkurepa:
The FF subscription changed to take builds from the VS
https://github.com/dotnet/arcade-services/issues/5564 18.3 channel, forcing us to reset from the latest build from it


aortiz-msft:
@dkurepa - I don't understand why this change is needed. Would you please elaborate on your PR description?

zivkan:
When 10.0.2xx eventually goes GA, it's going to ship NuGet 7.3, which we publish to darc's VS 18.3 channel. So, what the VMR currently has is old and needs to be updated (this PR).

What I'd like clarification on is the process on updating the VMR forward and back flows when we change darc channels. Now that VS and .NET have different shipping cadences, and NuGet has both VS and .NET components, we may be changing version numbers and therefore darc channels regularly.

PR #3591, opened 2025-11-27
Build VMR repo tools in PRs and fix issues

akoeplinger:
The ChangeValidation tool broke recently because DarcLib from arcade-services bumped Octokit and Newtonsoft.Json and they got downgraded via CPM in the VMR, but we never noticed since we only run the tool outside of darc dependency flow PRs.

Build the tools in the validation stage in PRs to prevent this and fix the version issues.


ViktorHofer (commented):
Any reason to not just always build them, when not building source only?

akoeplinger:
@ViktorHofer yeah I was wondering that too. the only reason I can think of is saving a bit of build time.

eng/tools/repo-tools.proj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.Build.NoTargets">
```
ViktorHofer:
```suggestion
<Project Sdk="Microsoft.Build.Traversal">
```

eng/tools/repo-tools.proj
ViktorHofer, line 1: I would make this project reference everything in this folder and rename it to tools.proj. I would then remove the few projects that are known to introduce prebuilts when building source-only.

eng/Versions.props
```diff
@@ -37,13 +37,14 @@
...
+    <MicrosoftExtensionsFileSystemGlobbingVersion>10.0.0</MicrosoftExtensionsFileSystemGlobbingVersion>
+    <MicrosoftExtensionsLoggingConsoleVersion>10.0.0</MicrosoftExtensionsLoggingConsoleVersion>
+    <MicrosoftExtensionsLoggingVersion>10.0.0</MicrosoftExtensionsLoggingVersion>
+    <MicrosoftExtensionsHostingVersion>10.0.0</MicrosoftExtensionsHostingVersion>
     <!-- sdk dependencies -->
     <MicrosoftDotNetApiCompatTaskVersion>10.0.100</MicrosoftDotNetApiCompatTaskVersion>
     <!-- external dependencies -->
-    <NewtonsoftJsonVersion>13.0.3</NewtonsoftJsonVersion>
-    <OctokitVersion>13.0.1</OctokitVersion>
+    <NewtonsoftJsonVersion>13.0.4</NewtonsoftJsonVersion>
```
ViktorHofer, line 61: I think Newtonsoft.Json is used in the tasks and updating the version to 13.0.4 without a re-bootstrap that includes the newer version from SBRP would introduce a prebuilt.

build.proj
```diff
@@ -18,6 +18,9 @@
                       Condition="'$(DotNetBuildSourceOnly)' == 'true'" />
     
     <!-- Pre-build: Init tools-->
+    <ProjectReference Include="$(RepositoryEngineeringDir)tools\repo-tools.proj"
+                      Condition="'$(BuildRepoTools)' == 'true'"
```
ViktorHofer: As discussed, I would remove the condition so that it always builds the tools.

```suggestion
    <ProjectReference Include="$(RepositoryEngineeringDir)tools\repo-tools.proj"
```

eng/tools/repo-tools.proj
```diff
@@ -0,0 +1,14 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+  </PropertyGroup>
```
ViktorHofer: Not needed when using the traversal SDK:

```suggestion
```

eng/Versions.props
akoeplinger: hmm interesting that it didn't fail in this PR?

ViktorHofer: Yeah, I'm surprised. Prebuilt detection runs in the source-build PR jobs. @dotnet/source-build any idea?

MichaelSimons: The [PSB package versions](https://github.com/dotnet/dotnet/blob/main/Directory.Build.props#L49) is overriding this.  We load the PSB so that we can use the latest versions.  In this case it causes a downgrade.

ViktorHofer: Interesting, I'm still learning how things flow with the SB infra :) NuGet treats downgrades in the dependency graph as errors. I assume the SB version flow infrastructure does not treat it as that because it just overrides that central property?

Do we see value in verifying that version downgrades don't happen - in this case by comparing the property version with what comes via PvP flow?

MichaelSimons: > I assume the SB version flow infrastructure does not treat it as that because it just overrides that central property?

That is correct.

> Do we see value in verifying that version downgrades don't happen - in this case by comparing the property version with what comes via PvP flow?

Yes - there would be value in that. https://github.com/dotnet/source-build/issues/5429

eng/tools/repo-tools.proj
```diff
@@ -0,0 +1,14 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="$(ToolsDir)BinaryToolKit\BinaryToolKit.csproj" />
+    <ProjectReference Include="$(ToolsDir)BuildComparer\BuildComparer.csproj" />
+    <ProjectReference Include="$(ToolsDir)ChangeValidation\ChangeValidation.csproj" />
+    <ProjectReference Include="$(ToolsDir)CreateBaselineUpdatePR\CreateBaselineUpdatePR.csproj" />
```
mthalman: Can we just use wildcards here to avoid maintenance?


```suggestion
    <ProjectReference Include="$(ToolsDir)**\*.csproj" />
```

ViktorHofer (approved):

PR #3616, opened 2025-12-01
Utilize CredScan suppressions file

NikolaMilosavljevic:
CredScan suppressions file was checked in, but wasn't referenced in YML. This should fix issues in `source-build-reference-packages`, `aspnetcore` and `msbuild`

@wtgodbe @rainersigwald

### DIFF
--- a/eng/pipelines/official.yml
+++ b/eng/pipelines/official.yml
@@ -106,6 +106,8 @@ extends:
         enabled: true
       tsa:
         enabled: true
+      credscan:
+        suppressionsFile: $(Build.SourcesDirectory)/.config/CredScanSuppressions.json
 
     containers:
       ${{ variables.almaLinuxContainerName }}:


wtgodbe (approved):

ViktorHofer (approved):

MichaelSimons (approved):

MichaelSimons:
The change validation failure is unrelated, merging.

akoeplinger:
credscan is not enabled by default anymore (it was replaced with 1ES Secret Scanning), before that credscan automatically respected the .config/CredScanSuppression.json so this should be a no-op now.

@NikolaMilosavljevic can you link the place where you saw errors?

NikolaMilosavljevic:
> credscan is not enabled by default anymore (it was replaced with 1ES Secret Scanning), before that credscan automatically respected the .config/CredScanSuppression.json so this should be a no-op now.
> 
> @NikolaMilosavljevic can you link the place where you saw errors?

This suppression method works in individual repos, i.e. `source-build-reference-packages`, but did not work in VMR due to the missing `credscan` parameter. At least that was the idea for this change.

This doesn't show in builds, only on s360 board.

NikolaMilosavljevic:
Hmm, `1ES Secret Scanning (SPMI)` tooling does not allow suppressions.

PR #3654, opened 2025-12-03
Fix installer tests URL for runtime-deps package

mthalman:
Package installer tests were broken for 10.0.2xx because they were attempting to download package URLs that don't exist (e.g. `https://ci.dot.net/public/Runtime/10.0.0-rtm.25523.111/dotnet-runtime-deps-10.0.0-rtm.25523.111-x64.deb`). This is because the filename should be using the stable version instead of the non-stable version but the directory name should continue to use the non-stable version.

This is resolved by making use of the `MicrosoftNETCoreAppRefVersion` property which will be the stable version in this case.

While working on this, I realized that the logic for computing the directory name is incorrect as well. It relies on the SDK version and then just converts that to a runtime version. But that's an incorrect assumption to make that the runtime version can be derived from the SDK version. It won't produce a correct runtime version in the scenario where we have an SDK-only release. So I've corrected this too by making use of the `MicrosoftNETCorePlatformsVersion` property which represents the non-stable runtime version.


ViktorHofer (commented):
Please make sure to get a review from @mmitche who recently (for 10.0 GA) changed which stable packages are used for version flow purposes.

PR #3656, opened 2025-12-03
Use Ubuntu container tag for package installer tests

mthalman:
The changes in https://github.com/dotnet/dotnet/pull/3634 cause an invalid tag to be referenced: `10.0-trixie-slim`. Debian images are [no longer provided](https://learn.microsoft.com/dotnet/core/compatibility/containers/10.0/default-images-use-ubuntu) for .NET 10 container images. Instead, the tag has been updated to use Ubuntu.


mmitche:
Resolves #3664

PR #3476, opened 2025-11-20
Fix -cross arg condition for "different OS" case

am11:
e.g. freebsd-x64 build failed with:

```
  /dotnet/src/runtime/src/coreclr/runtime.proj(114,5): error MSB3073: The command ""/dotnet/src/runtime/src/coreclr/build-runtime.sh" -x64 -release -portablebuild=false -keepnativesymbols -os freebsd -targetrid freebsd-x64 -cmakeargs "-DCLR_DOTNET_RID=linux-x64" -cmakeargs "-DCLR_DOTNET_HOST_PATH=/dotnet/.dotnet/dotnet" -cmakeargs "-DCDAC_BUILD_TOOL_BINARY_PATH=/dotnet/src/runtime/artifacts/bin/coreclr/freebsd.x64.Release/cdac-build-tool/cdac-build-tool.dll" -component runtime -cmakeargs "-DCLR_CMAKE_RUNTIME_CORECLR=1" -cmakeargs "-DCMAKE_PRODUCT_VERSION=10.0.0" -cmakeargs "-DCMAKE_CONTINUOUS_INTEGRATION_BUILD=false" -cmakeargs "-DCMAKE_NET_CORE_APP_CURRENT_VERSION=10.0" -cmakeargs "-DCMAKE_BUILD_LIBRARIES_CONFIGURATION=Release" -cmakeargs "-DCMAKE_BUILD_RUNTIME_CONFIGURATION=Release"" exited with code 1.
```

because there is no `-cross`.

PR #3700, opened 2025-12-05
Revert "Fix -cross arg condition for "different OS" case"

ViktorHofer:
Reverts dotnet/dotnet#3476

See https://github.com/dotnet/dotnet/pull/3476#issuecomment-3617107625


akoeplinger:
https://github.com/dotnet/dotnet/pull/3701 fixes the root cause instead.

Issue #3706, opened 2025-12-05
System.Drawing.Common 10.0.2 is targeted at a non-isolated feed "dotnet10-internal"


mmitche:
## Error

```
D:\a\_work\1\s\.packages\microsoft.dotnet.arcade.sdk\11.0.0-beta.25570.6\tools\SdkTasks\PublishArtifactsInManifest.proj(126,5): error : Package 'System.Drawing.Common' has stable version '10.0.2' but is targeted at a non-isolated feed 'https://pkgs.dev.azure.com/dnceng/internal/_packaging/dotnet10-internal/nuget/v3/index.json'
```

This likely means that the stability information for this package is incorrect.

## Links
- https://dev.azure.com/dnceng/internal/internal%20Team/_build/results?buildId=2854263&view=logs&j=ba23343f-f710-5af9-782d-5bd26b102304&t=74531eb2-9b39-5603-839e-94e3ba212b65&l=952

## Affected Releases

- 10.0.2

mmitche:
This is likely due to changes in the stable build handling.

mmitche:
This used to 'work' because we passed DotNetFinalVersionKind into the root of the repo. If in release mode, this would get interpreted by the publishing infra as being a stable build. Because the root repo uses V4 publishing, it would then annotate each repo's artifacts with CouldBeStable based on the shipping/non-shipping bit.

This meant we were annotating (incorrectly) a ton more assets with CouldBeStable), but now we miss System.Drawing.Common.

mmitche:
Completed

PR #3668, opened 2025-12-03
[release/10.0.1xx] Allow default branding to be set by the VMR orchestrator

ViktorHofer:
Manual backport of https://github.com/dotnet/dotnet/commit/f59b33bfb1617ae94b4c4f3fbb1e9733cde4b7e2

PR #3711, opened 2025-12-05
Fix stable publishing issue

mmitche:
When https://github.com/dotnet/dotnet/pull/3668 went it, it removed DotNetFinalVersionKind from being set at the root level of the VMR. This unintentionally was causing the VMR publishing process to mark all shipping assets as CouldBeStable=true when DotNetFinalVersionKind=release, rather than using the values that were coming from the repos. The fix here is to not set PublishingVersion at the repo level. This will cause VMR builds to use V4 publishing for the repo builds, which will set CouldBeStable appropriately. Only winforms and wpf were manually setting this.

### DIFF
--- a/src/winforms/eng/Publishing.props
+++ b/src/winforms/eng/Publishing.props
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
    <PropertyGroup>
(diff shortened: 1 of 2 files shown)


mmitche:
Resolves #3706

hoyosjs (approved):

Issue #3682, opened 2025-12-04
[10.0] Mac/Linux signing validation failure


mthalman:
```
/Users/runner/work/1/s/eng/signing-validation.proj(20,7): error : There are 48 unsigned files.
     Private.SourceBuilt.Artifacts.10.0.102-servicing.25603.112.centos.10-x64.tar.gz/Spectre.Console.0.52.0.nupkg/lib/net10.0/Spectre.Console.dll: Outcome="Unsigned" AuthentiCode="Timestamp: 12/04/25 6:33:00 (sha256RSA), AuthentiCode signed: True" StrongName="StrongName signed: False"
     dotnet-sourcelink.10.0.102.nupkg/tools/net8.0/any/shims/osx-x64/sourcelink: Outcome="Unsigned"
     dotnet-runtime-deps-10.0.2-sles.15-newkey-arm.rpm: Outcome="Unsigned" Error="gpg: Signature made Thu Dec  4 05:49:58 2025 UTC
  gpg:                using RSA key EE4D7792F748182B
  gpg: Can't check signature: No public key
  "
     dotnet-host-10.0.2-newkey-aarch64.rpm: Outcome="Unsigned" Error="gpg: Signature made Thu Dec  4 06:18:50 2025 UTC
  gpg:                using RSA key EE4D7792F748182B
  gpg: Can't check signature: No public key
  "
     dotnet-host-10.0.2-newkey-arm64.deb: Outcome="Unsigned" Error="gpg: Signature made Thu Dec  4 06:18:10 2025 UTC
  gpg:                using RSA key EE4D7792F748182B
  gpg: Can't check signature: No public key
  "
     dotnet-runtime-deps-10.0.2-opensuse.15-newkey-arm.rpm: Outcome="Unsigned" Error="gpg: Signature made Thu Dec  4 05:50:21 2025 UTC
  gpg:                using RSA key EE4D7792F748182B
  gpg: Can't check signature: No public key
  "
     dotnet-runtime-deps-10.0.2-newkey-arm64.deb: Outcome="Unsigned" Error="gpg: Signature made Thu Dec  4 06:18:15 2025 UTC
  gpg:                using RSA key EE4D7792F748182B
  gpg: Can't check signature: No public key
  "
     dotnet-runtime-deps-10.0.2-newkey-arm.deb: Outcome="Unsigned" Error="gpg: Signature made Thu Dec  4 05:50:16 2025 UTC
[...]

PR #3734, opened 2025-12-08
Add microsoft-2025 key for non-SHA1 repos

mmitche:


mmitche:
Should resolve https://github.com/dotnet/dotnet/issues/3682

ellahathaway:
Does this need to go into SignTool as well?

mmitche:
> Does this need to go into SignTool as well?

Yes thanks

PR #3824, opened 2025-12-11
Fix VMR source-only build to use --with-packages for prebuilts

csa7mdm:
Fixes https://github.com/dotnet/source-build/issues/4427

### Technical Analysis
The VMR validation pipeline (`vmr-build.yml`) was downloading prebuilt artifacts to the default `prereqs/packages/archive/` location, which allowed `build.sh` to pick them up implicitly. This left the explicit `--with-packages` argument path untested in CI.

### Changes
- Modified `eng/pipelines/templates/jobs/vmr-build.yml` to download prebuilts to a custom location (`prebuilt-packages/`).
- Added logic to pass `--with-packages` to `build.sh` when `reuseBuildArtifactsFrom` is set.
- This ensures the `CustomPreviouslySourceBuiltPackagesPath` logic in the build script is exercised and verified.

### Verification
- Validated locally that `build.sh` accepts the `--with-packages` argument.
- Verified the YAML logic constructs the command line correctly.

### DIFF
--- a/eng/pipelines/templates/jobs/vmr-build.yml
+++ b/eng/pipelines/templates/jobs/vmr-build.yml
@@ -632,10 +632,6 @@ jobs:
               exit 1
             fi
             
-            mkdir -p "$(previouslySourceBuiltSdkDir)"
-            echo "Searching for SDK tarball in $sdkSearchDir..."
-            find "$sdkSearchDir" -name "*.tar.gz" || true
-            
             previousSdkPath=$(find "$sdkSearchDir" -name "dotnet-sdk-*.tar.gz" | head -n 1)
             
             if [ -z "$previousSdkPath" ]; then
@@ -647,6 +643,7 @@ jobs:
             
             echo "Found SDK tarball: $previousSdkPath"
             echo "Extracting SDK..."
+            mkdir -p "$(previouslySourceBuiltSdkDir)"
             eval tar -xzf "$previousSdkPath" -C "$(previouslySourceBuiltSdkDir)"
             
             customPrepArgs="${customPrepArgs} --with-sdk $(previouslySourceBuiltSdkDir)"
@@ -662,9 +659,6 @@ jobs:
                exit 1
              fi
 
-             mkdir -p "$(previouslySourceBuiltArtifactsDir)"
-             
-             echo "Searching for Artifacts tarball in $(downloadedArtifactsDir)..."
              artifactsTarball=$(find "$(downloadedArtifactsDir)" -name "Private.SourceBuilt.Artifacts.*.tar.gz" | head -n 1)
              
              if [ -z "$artifactsTarball" ]; then
@@ -676,6 +670,7 @@ jobs:
              
              echo "Found Artifacts tarball: $artifactsTarball"
              echo "Extracting artifacts..."
+             mkdir -p "$(previouslySourceBuiltArtifactsDir)"
              tar -xzf "$artifactsTarball" -C "$(previouslySourceBuiltArtifactsDir)"
              
              customPrepArgs="${customPrepArgs} --with-packages $(previouslySourceBuiltArtifactsDir)"


eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -406,7 +406,7 @@ jobs:
             downloadFilePatterns: |
               assets/*/Private.SourceBuilt.Artifacts.*.tar.gz
               assets/*/dotnet-sdk-*.tar.gz
-            copyDestination: $(sourcesPath)/prereqs/packages/archive/
+            copyDestination: $(sourcesPath)/prebuilt-packages/
```
MichaelSimons: Changing this directory is going to have a ripple effect.  There are other steps that are retrieving the SDK tarball from this path that don't look to be updated.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -584,6 +584,11 @@ jobs:
               extraBuildProperties="$extraBuildProperties /p:DotNetSourceOnlyPostBuildSign=true"
             fi
 
+            if [[ '${{ length(parameters.reuseBuildArtifactsFrom) }}' -gt '0' ]]; then
+              mkdir -p "$(sourcesPath)/prebuilt-packages/"
+              customBuildArgs="$customBuildArgs --with-packages $(sourcesPath)/prebuilt-packages/"
```
MichaelSimons: IIRC, with-packages must point to a directory containing the extracted packages not the tarball.

eng/pipelines/templates/jobs/vmr-build.yml
MichaelSimons: I think it would be better to copy these artifacts outside of the sources directory all together e.g. Agent.TempDirectory.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -407,7 +407,7 @@ jobs:
               assets/*/Private.SourceBuilt.Artifacts.*.tar.gz
               assets/*/dotnet-sdk-*.tar.gz
             copyDestination: $(sourcesPath)/prereqs/packages/archive/
```
MichaelSimons: I think it would be better to avoid downloading these to the packages/archive directory all together because of the special logic that looks for archives here.  I would prefer to utilize the temp directory and modify the consuming locations to work with it.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -584,6 +584,13 @@ jobs:
               extraBuildProperties="$extraBuildProperties /p:DotNetSourceOnlyPostBuildSign=true"
             fi
 
+            if [[ '${{ length(parameters.reuseBuildArtifactsFrom) }}' -gt '0' ]]; then
+               # Extract the prebuilt packages to a temporary directory
+               mkdir -p "$(Agent.TempDirectory)/prebuilt-packages"
+               tar -xzf $(sourcesPath)/prereqs/packages/archive/Private.SourceBuilt.Artifacts.*.tar.gz -C "$(Agent.TempDirectory)/prebuilt-packages"
+               customBuildArgs="$customBuildArgs --with-packages $(Agent.TempDirectory)/prebuilt-packages/"
```
MichaelSimons: While these are a type of prebuilt-package, I think it would be better to call these previously-built-packages, psb, or previous-packages.  The system also has the ability to input prebuilt packages which are very different that previously built packages.

eng/pipelines/templates/jobs/vmr-build.yml
mthalman: We should avoid putting anything in this location to prevent any accidental usage of it by the build logic.

mthalman: Oh wait. We need the sdk to be there, but not the artifacts tarball.

mthalman: Actually, we should expand the scope of this PR a bit so that it includes both the artifacts tarball and the sdk tarball. That really gets to the intent of validating the workflow that would be used by distro maintainers. So rather than [extracting the sdk to the well-known `.dotnet` directory](https://github.com/dotnet/dotnet/blob/9add30f0238eabacba9a92acec51a8d714fd1881/eng/pipelines/templates/jobs/vmr-build.yml#L552C13-L553), we should follow the same pattern as your proposed changes and extract the sdk tarball to a temp directory and reference it with the `--with-sdk` option.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -406,8 +406,8 @@ jobs:
             downloadFilePatterns: |
               assets/*/Private.SourceBuilt.Artifacts.*.tar.gz
               assets/*/dotnet-sdk-*.tar.gz
-            copyDestination: $(sourcesPath)/prereqs/packages/archive/
-            flattenDirs: true
+            copyDestination: $(Agent.TempDirectory)/downloaded-artifacts/
+            flattenDirs: false
```
mthalman: What is the need for not flattening? That just requires wildcards for the directory names when needing to reference the files.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -547,13 +547,13 @@ jobs:
 
           if [[ "$prepSdk" == "false" ]]; then
             customPrepArgs="${customPrepArgs} --no-sdk"
-            rm -rf $(sourcesPath)/.dotnet
-            mkdir $(sourcesPath)/.dotnet
-            previousSdkPath="$(sourcesPath)/prereqs/packages/archive/dotnet-sdk-*.tar.gz"
-            eval tar -ozxf "$previousSdkPath" -C "$(sourcesPath)/.dotnet"
+            
+            mkdir -p "$(Agent.TempDirectory)/previously-source-built-sdk"
```
mthalman: Consider defining a variable for the directory path so it can be reused. Same applies to the artifacts usage.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -584,6 +584,13 @@ jobs:
               extraBuildProperties="$extraBuildProperties /p:DotNetSourceOnlyPostBuildSign=true"
             fi
 
+            if [[ '${{ length(parameters.reuseBuildArtifactsFrom) }}' -gt '0' ]]; then
+               # Extract the prebuilt packages to a temporary directory
+               mkdir -p "$(Agent.TempDirectory)/previously-source-built"
```
mthalman: For explicitness and consistency with sdk, consider naming this `previously-source-built-artifacts`.

mthalman (changes requested):

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -547,13 +554,22 @@ jobs:
 
           if [[ "$prepSdk" == "false" ]]; then
             customPrepArgs="${customPrepArgs} --no-sdk"
-            rm -rf $(sourcesPath)/.dotnet
-            mkdir $(sourcesPath)/.dotnet
-            previousSdkPath="$(sourcesPath)/prereqs/packages/archive/dotnet-sdk-*.tar.gz"
-            eval tar -ozxf "$previousSdkPath" -C "$(sourcesPath)/.dotnet"
-            eval rm -f "$previousSdkPath"
+
```
MichaelSimons, line 613: I think this logic to extract the sdk and package should be moved to the `prep the Build` step.  Both the `prep-source-build.sh` and `build.sh` should be passing in `--with-sdk` and `--with-packages` when the appropriate parameters are set.  If you look at the current prep logic, you see it is trying to use the downloaded sdk.

mthalman:
There's a script error: `/__w/_temp/8a1c070b-caf0-4c1a-9454-6aaa2bb147e2.sh: line 96: syntax error near unexpected token 'fi'`

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -584,6 +659,11 @@ jobs:
               extraBuildProperties="$extraBuildProperties /p:DotNetSourceOnlyPostBuildSign=true"
             fi
 
+            if [[ '${{ length(parameters.reuseBuildArtifactsFrom) }}' -gt '0' ]]; then
+               # Prebuilt packages are extracted in the Prep step
+               :
```
mthalman: I think this may be causing the script error?

mthalman:
@csa7mdm - I see activity from you of merging but not addressing the build issue. Are you planning on doing that?

csa7mdm:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/dotnet/dotnet/pull/3824#pullrequestreview-3588585512)

mthalman:
There is a failure in the prep script: `ERROR: --no-sdk requires --no-bootstrap or a pre-existing .dotnet SDK directory.  Exiting...`

mthalman:
@csa7mdm - Can you follow-up on the pipeline error that is showing up?

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -545,20 +572,89 @@ jobs:
...
           fi
 
+          
+          additionalBuildArgs=""
+          
+          echo "Debug Paths:"
+          echo "Pipeline Workspace: $(Pipeline.Workspace)"
+          echo "Build Sources Directory: $(Build.SourcesDirectory)"
+          echo "Agent Temp Directory: $(Agent.TempDirectory)"
+          echo "Downloaded Artifacts Directory: $(downloadedArtifactsDir)"
```
mthalman: I think these can be removed now.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -545,20 +572,89 @@ jobs:
...
           if [[ "$prepSdk" == "false" ]]; then
             customPrepArgs="${customPrepArgs} --no-sdk"
-            mkdir -p $(sourcesPath)/.dotnet
-            previousSdkPath="$(sourcesPath)/prereqs/packages/archive/dotnet-sdk-*.tar.gz"
-            eval tar -ozxf "$previousSdkPath" -C "$(sourcesPath)/.dotnet"
-            eval rm -f "$previousSdkPath"
+            
+            # Verify download directory exists
+            if [ ! -d "$(downloadedArtifactsDir)" ]; then
+              echo "Error: SDK extraction required but '$(downloadedArtifactsDir)' does not exist."
```
mthalman: Use the [Azure DevOps `error` logging syntax](https://learn.microsoft.com/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#example-log-an-error) instead. This applies to several other instances below too.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -583,6 +679,8 @@ jobs:
               extraBuildProperties="$extraBuildProperties /p:DotNetSourceOnlyPostBuildSign=true"
             fi
 
+
+
```
mthalman: delete

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -545,20 +572,89 @@ jobs:
...
+          if [[ -n "$additionalBuildArgs" ]]; then
+            echo "##vso[task.setvariable variable=additionalBuildArgs]$additionalBuildArgs"
           fi
 
           ./prep-source-build.sh $customPrepArgs
         displayName: Prep the Build
         workingDirectory: $(sourcesPath)
 
+
```
mthalman: delete

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -545,19 +572,95 @@ jobs:
...
+          
+          echo "Prep arguments: $customPrepArgs"
+          echo "Additional build arguments: $additionalBuildArgs"
+
+          if [[ -n "$additionalBuildArgs" ]]; then
+            echo "##vso[task.setvariable variable=additionalBuildArgs]$additionalBuildArgs"
           fi
 
           ./prep-source-build.sh $customPrepArgs
-        displayName: Prep the Build
```
mthalman, line 590: This should be added back. The build step just shows the name as a generic `CmdLine`.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -526,7 +553,9 @@ jobs:
           $(sourcesPath)/eng/common/native/install-dependencies.sh osx
         displayName: Install dependencies
     - ${{ if eq(parameters.buildSourceOnly, 'true') }}:
-      - script: |
+      - displayName: Prep the Build
+        workingDirectory: $(sourcesPath)
```
mthalman: This isn't valid syntax: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1269294&view=results. It needs to be:

```yaml
- script: |
    <script contents>
  displayName: Prep the Build
  workingDirectory: $(sourcesPath)
```

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -545,20 +572,95 @@ jobs:
...
+            if [ ! -d "$sdkSearchDir" ]; then
+              echo "##vso[task.logissue type=error]SDK extraction required but neither '$(downloadedArtifactsDir)' nor '$fallbackSdkDir' exists."
+              echo "Checking Pipeline Workspace content:"
+              ls -R "$(Pipeline.Workspace)" || true
+              exit 1
+            fi
+            
+            mkdir -p "$(previouslySourceBuiltSdkDir)"
+            echo "Searching for SDK tarball in $sdkSearchDir..."
+            find "$sdkSearchDir" -name "*.tar.gz" || true
```
MichaelSimons: Is this find necessary?  It feels redundant in a way with the ` ls -R "$sdkSearchDir" || true` cmd if the tarball is not found.  This pattern is also inconsistent with the logging for the artifactsTarball.

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -545,20 +572,95 @@ jobs:
...
+          if [[ '${{ length(parameters.reuseBuildArtifactsFrom) }}' -gt '0' ]]; then
+             # Verify download directory exists
+             if [ ! -d "$(downloadedArtifactsDir)" ]; then
+               echo "##vso[task.logissue type=error]Package extraction required but '$(downloadedArtifactsDir)' does not exist."
+               echo "Checking Pipeline Workspace content:"
+               ls -R "$(Pipeline.Workspace)" || true
+               exit 1
+             fi
+
+             mkdir -p "$(previouslySourceBuiltArtifactsDir)"
```
MichaelSimons: Consider moving just before it is used in by the tar cmd.  IMO it helps readability. Same comment applies to the sdk section.

```
             echo "Extracting artifacts..."
             mkdir -p "$(previouslySourceBuiltArtifactsDir)"
             tar -xzf "$artifactsTarball" -C "$(previouslySourceBuiltArtifactsDir)"
```

MichaelSimons (approved):

PR #3825, opened 2025-12-11
Test fix of Source code updates from dotnet/runtime

hoyosjs:
- **[runtime] Source update bce6119 → bf3f051 Diff: https://github.com/dotnet/runtime/compare/bce6119e41ecfbcf630c369836770669604c22c6..bf3f0518e024437ad128a4178a8a18239b70a9f6**
- **Update dependencies after merging main into darc-main-2e34fd53-2c9d-4f9f-859a-c0140a0ed320**
- **[runtime] Source update bf3f051 → 0fb17f7 Diff: https://github.com/dotnet/runtime/compare/bf3f0518e024437ad128a4178a8a18239b70a9f6..0fb17f7320fbaaa93e91c4c08b3a03175bb5ca15**
- **Update dependencies after merging main into darc-main-2e34fd53-2c9d-4f9f-859a-c0140a0ed320**
- **Don't use local packs for cdac unless it's a community supported configuration**
- **Test setting UseBootsrapLayout**

### DIFF
--- a/src/runtime/src/native/managed/Directory.Build.props
+++ b/src/runtime/src/native/managed/Directory.Build.props
@@ -2,7 +2,8 @@
   <Import Project="..\..\..\Directory.Build.props" />
   <Import Project=".\native-library.props" Condition="'$(IsSourceProject)' == 'true' and '$(Language)' == 'C#'" />
 
-  <PropertyGroup Condition="'$(UseBootstrapLayout)' == 'true' and ('$(TargetOS)' == 'freebsd' or '$(TargetArchitecture)' == 'loongarch64' or '$(TargetArchitecture)' == 'riscv64')">
+  <PropertyGroup Condition="'$(UseBootstrap)' == 'true'">
+    <UseBootstrapLayout>true</UseBootstrapLayout>
     <UseLocalTargetingRuntimePack>true</UseLocalTargetingRuntimePack>
   </PropertyGroup>
 </Project>


hoyosjs:
@steveisok This is a test of https://github.com/dotnet/dotnet/pull/3766 workarounds

src/runtime/src/native/managed/Directory.Build.props
```diff
@@ -1,4 +1,9 @@
 <Project>
-    <Import Project="..\..\..\Directory.Build.props" />
-    <Import Project=".\native-library.props" Condition="'$(IsSourceProject)' == 'true' and '$(Language)' == 'C#'" />
+  <Import Project="..\..\..\Directory.Build.props" />
+  <Import Project=".\native-library.props" Condition="'$(IsSourceProject)' == 'true' and '$(Language)' == 'C#'" />
+
+  <PropertyGroup Condition="'$(UseBootstrap)' == 'true'">
+    <UseBootstrapLayout>true</UseBootstrapLayout>
```
hoyosjs, line 6: @jkoritzinsky - this is a possible thing to make it use the bootstrap layout. This gets used in all the *_publish projects, but now this triggers toolAot.targets to try to import Microsoft.DotNet.ILCompiler.SingleEntry.targets from CoreCLRBuildIntegrationDir - `src\runtime\artifacts\bin\coreclr\windows.x64.Release\build`. That requires `nativeaot\BuildIntegration\BuildIntegration.proj` builds. I see it in the list - in the VMR prior to this, it seems we were using from package `ILCompilerTargetsPath = D:\a\_work\1\s\src\runtime\artifacts\.packages\microsoft.dotnet.ilcompiler\10.0.0\build\Microsoft.DotNet.ILCompiler.SingleEntry.targets`. On the one after this change we try to use a live build one - but the content hasn't been copied.  Is it a build ordering/dependency issue where we need to make sure the live copy is placed there?

jkoritzinsky: Maybe? If moving building `BuildIntegration.proj` earlier works, we can do that. Otherwise, ensuring that `BuildIntegration.proj` is built in the bootstrap and copied into the bootstrap layout would work as well.

PR #3857, opened 2025-12-12
Install .NET 8.0 when using darc

premun:
Copies of latest `eng/common` files from Arcade to allow us to unpin darc.


premun:
@ViktorHofer I wanted to temporarily override this to fix publishing. I would not care about the overwrites.

But we rolled back Maestro instead in the end.

premun:
Okay, we will have another go at unpinning so this needs to be in (before we rebootstrap which will bring in these files anyway).

PR #3895, opened 2025-12-15
Introduce more version properties

NikolaMilosavljevic:
This also fixes the hard-coded product version in Installer tests: https://github.com/dotnet/dotnet/blob/3e4f63732a2e61ad7ae1d567e820f70b2a55c2f1/test/Microsoft.DotNet.Installer.Tests/Microsoft.DotNet.Installer.Tests.csproj#L71-L75

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -6,8 +6,7 @@
     <VersionSDKMinor>1</VersionSDKMinor>
     <MajorVersion>11</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <MajorMinorVersion>$(MajorVersion).$(MinorVersion)</MajorMinorVersion>
-    <VersionPrefix>$(MajorMinorVersion).$(VersionSDKMinor)00</VersionPrefix>
+    <VersionPrefix>$(MajorVersion).$(MinorVersion).$(VersionSDKMinor)00</VersionPrefix>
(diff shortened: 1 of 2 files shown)


eng/Versions.props
```diff
@@ -4,7 +4,10 @@
 
   <PropertyGroup>
     <VersionSDKMinor>1</VersionSDKMinor>
-    <VersionPrefix>11.0.$(VersionSDKMinor)00</VersionPrefix>
+    <MajorVersion>11</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <MajorMinorVersion>$(MajorVersion).$(MinorVersion)</MajorMinorVersion>
```
ViktorHofer: I would prefer to avoid this MajorMinorVersion property and just concatenate in the places that need it, honestly.

NikolaMilosavljevic: Yeah, likely a rare use case for `MajorMinorVersion` property. @mmitche thoughts?

mmitche: There are a few use cases (e.g. aspnetcore repo installer branding uses something like this). But, if it's single use within the VMR then I'd say eliminate it.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/3895/commits/d908f16e87ef1e358ccfcd5656799b286078038a

PR #3897, opened 2025-12-15
Support maui/android/ios workload manifests for source build

mthalman:
Contributes to dotnet/source-build#3242

This supports these workload manifests by adding them as text-only packages to SBRP. Then it was just a matter of enabling them in a source-only configuration by removing the condition that excluded them.

While there is cost associated with maintaining these packages in SBRP, that is limited to just the preview part of the release cycle. After GA, the versions will be static.

The alternative to using SBRP is adding the maui, android, and macios repos to the VMR and building the workload manifest packages as part of the build. But this is particularly costly, particularly for the android and macios repos, as they don't follow standard arcade build processes. Instead they have custom logic in Makefiles which dynamically generate the content files that get included in the packages. This dynamic behavior has a number of dependency that are not source-build compatible and would require a fair amount of effort and coordination.

### DIFF
--- a/src/sdk/eng/Versions.props
+++ b/src/sdk/eng/Versions.props
@@ -150,6 +150,7 @@
     <!--
       Updates to any of these versions will require the updating of that package version in the
       source-build-reference-packages repo as a text-only package.
+      See https://github.com/dotnet/source-build-reference-packages?tab=readme-ov-file#text-only
     -->
     <MauiWorkloadManifestVersion>10.0.0</MauiWorkloadManifestVersion>
     <XamarinAndroidWorkloadManifestVersion>36.1.2</XamarinAndroidWorkloadManifestVersion>


mthalman:
```
[root@382e365ea858 extracted-dotnet-sdk]# ./dotnet workload search

Workload ID                 Description
---------------------------------------------------------------------------------------
android                     .NET SDK Workload for building Android applications.
android-net8                .NET SDK Workload for building .NET 8 Android applications.
macos                       .NET SDK Workload for building macOS applications.
maui-android                .NET MAUI SDK for Android
maui-tizen                  .NET MAUI SDK for Tizen
maui-windows                .NET MAUI SDK for Windows
wasi-experimental           .NET WASI experimental tooing for net10.0
wasi-experimental-net8      .NET WASI experimental
wasi-experimental-net9      .NET WASI experimental tooing for net9.0
wasm-experimental           .NET WebAssembly experimental tooling for net10.0
wasm-experimental-net7      .NET WebAssembly experimental tooling for net7.0
wasm-experimental-net8      .NET WebAssembly experimental tooling for net8.0
wasm-experimental-net9      .NET WebAssembly experimental tooling for net9.0
wasm-tools                  .NET WebAssembly build tools for net10.0
wasm-tools-net6             .NET WebAssembly build tools for net6.0
wasm-tools-net7             .NET WebAssembly build tools for net7.0
wasm-tools-net8             .NET WebAssembly build tools for net8.0
wasm-tools-net9             .NET WebAssembly build tools for net9.0
```

mthalman:
@jonathanpeppers - Since you've been updating the maui versions, these changes are relevant to you. The recommendation is to now apply these changes in `dotnet/dotnet` repo instead of `dotnet/sdk`. This is because corresponding changes are also going to be need to the `dotnet/source-build-reference-packages` repo. So you can do it all in one PR with the VMR.

Let me know if more clarity is needed in comments as far as what needs to be done.

src/source-build-reference-packages/src/textOnlyPackages/src/microsoft.net.sdk.android.manifest-10.0.100-preview.6/36.0.0-preview.6.169/data/WorkloadDependencies.json
```diff
@@ -0,0 +1,75 @@
+{
+  "microsoft.net.sdk.android": {
+    "workload": {
+      "alias": [
+        "android"
+      ],
+      "version": "36.0.0-preview.6.169"
```
jonathanpeppers: Can we regenerate these with the .NET 10 GA builds:
 https://github.com/dotnet/sdk/blob/2b7468dad75a2c8dc419cd30da4443ac6d699aa6/eng/Versions.props#L145-L153

I wouldn't want to commit the preview.6 ones here, even temporarily.

mthalman: I see. Those changes from preview.6 to GA haven't yet reached the VMR: https://github.com/dotnet/dotnet/blob/162eebc5c92f472c9255c3225b522b32485c2df2/src/sdk/eng/Versions.props#L148-L154. The updates will be coming in via https://github.com/dotnet/dotnet/pull/3754. I'll hold off until that PR gets merged. Then I'll update this PR to react to the updated versions.

PR #3960, opened 2025-12-18
Simplify BuildOS default value

am11:
One less hardcoded list; goodness.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,7 +6,7 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
-  <Import Project="$(MSBuildThisFileDirectory)eng/RuntimeIdentifier.props-$(NETCoreSdkPortableRuntimeIdentifier)" />
+  <Import Project="$(MSBuildThisFileDirectory)eng/RuntimeIdentifier.props" />
 
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" Condition="'$(SkipArcadeSdkImport)' != 'true'" />
 


eng/RuntimeIdentifier.props
```diff
@@ -1,15 +1,7 @@
...
-    <BuildOS>linux</BuildOS>
-    <BuildOS Condition="$(NETCoreSdkPortableRuntimeIdentifier.StartsWith('linux-musl'))">linux-musl</BuildOS>
-    <BuildOS Condition="$([MSBuild]::IsOSPlatform('OSX'))">osx</BuildOS>
-    <BuildOS Condition="$([MSBuild]::IsOSPlatform('FREEBSD'))">freebsd</BuildOS>
-    <BuildOS Condition="$([MSBuild]::IsOSPlatform('NETBSD'))">netbsd</BuildOS>
-    <BuildOS Condition="$([MSBuild]::IsOSPlatform('ILLUMOS'))">illumos</BuildOS>
-    <BuildOS Condition="$([MSBuild]::IsOSPlatform('SOLARIS'))">solaris</BuildOS>
-    <BuildOS Condition="$([MSBuild]::IsOSPlatform('HAIKU'))">haiku</BuildOS>
-    <BuildOS Condition="$([MSBuild]::IsOSPlatform('WINDOWS'))">windows</BuildOS>
+    <BuildOS Condition="'$(BuildOS)' == ''">$(BuildRid.Substring(0, $(BuildRid.LastIndexOf('-'))))</BuildOS>
```
ViktorHofer, line 7: Special case windows vs win. Rid is win but BuildOS should be windows.

ViktorHofer, line 7: You are reading from `BuildRid` before it's defined below.

am11: Ah, I was focused on the case where it's defined here: https://github.com/dotnet/dotnet/blob/d7e1f5b5b21c7f7e23d3686ab114be1879cb42d8/build.sh#L354

ViktorHofer: We support non script based build entry points as well, i.e. dotnet build on specific projects. We don't want to break that :)

build.sh
```diff
@@ -354,10 +354,16 @@ if [[ -n "$__DistroRid" ]]; then
   properties+=( "/p:BuildRid=$__DistroRid" )
 fi
 
+if "/usr/bin/ldd" --version 2>&1 | grep -q musl; then
+  properties+=( "/p:BuildOS=linux-musl-$arch" )
+else
+  properties+=( "/p:BuildOS=$os-$arch" )
+fi
+
```
ViktorHofer, line 362: Why is this now necessary and wasn't before? I would avoid this when possible. We should make sure that the calculation of Build* properties is consistent between the script and non-script based entrypoints.

am11: Yes, without it, build fails.

am11:
@ViktorHofer, could you please run the pipeline with alpine legs?

build.sh
ViktorHofer: Does this mean that the non-script based entry point has the wrong value?

ViktorHofer:
Happy to trigger them after we agreed on the solution as we are quite short on CI resources at the moment.

build.sh
am11: Yup, the BuildRid is the DistroRid of the build machine and there is no non-portable variant for build OS (we have PortableTargetOS but not PortableBuildOS). In main, we were checking each platform separately. My approach is: "since in the script context, we know what build OS non-portable name is, lets pass this on". For the Windows case, we can get it from the context since there is no non-portable RID there.

What is the use-case for non-script case? I thought the top level script is the only entrypoint on Unix. Otherwise, we will miss lots of context. 👀

Issue #4080, opened 2026-01-07
Build fails with MSB3883: Unexpected exception: Unknown error 65564


ericstj:
Attempting a source-build on centos stream 10 and got this:
```
/home/ericstj/src/dotnet/dotnet/.dotnet/sdk/11.0.100-dev/Roslyn/Microsoft.CSharp.Core.targets(89,5): error MSB3883: Unexpected exception:  [/home/ericstj/src/dotnet/dotnet/src/arcade/src/Microsoft.DotNet.Build.Tasks.Packaging/src/Microsoft.DotNet.Build.Tasks.Packaging.csproj]
/home/ericstj/src/dotnet/dotnet/.dotnet/sdk/11.0.100-dev/Roslyn/Microsoft.CSharp.Core.targets(89,5): Unknown error 65564 [/home/ericstj/src/dotnet/dotnet/src/arcade/src/Microsoft.DotNet.Build.Tasks.Packaging/src/Microsoft.DotNet.Build.Tasks.Packaging.csproj]
```

Not much else to go on.  Binlog is here: 
[Build.binlog.zip](https://github.com/user-attachments/files/24462200/Build.binlog.zip)

jkotas:
This is not very actionable without the exception details.

ericstj:
Yeah, seems the build should be set up better to capture crashes during execution.  We need a stack + dump.

@ViktorHofer  is looking into this one right now as it seems to be happening in stage2 build - meaning it's likely caused by the runtime flow that's being merged in https://github.com/dotnet/dotnet/pull/3873.  

Based on what we discover I think we should plan on follow up work that improves the diagnostics in our own builds -- since we'll be building the product using nightly builds it's more likely that the build will catch issues that unit tests did not.  Especially when folks are contributing code changes directly to VMR.

ViktorHofer:
Wait, I'm not looking into this one. I'm trying to get a crash dump for https://github.com/dotnet/dotnet/issues/4088 which still happens even after cherry-picking Jeremy's fix.

ericstj:
My hunch is that they are the same / similar root cause but I could be wrong.  When I filed this I hadn't realized it was only stage2 which was using unproven runtime.  Let's wait for the dust to settle around the runtime flow and #4088 and then decide what to do with this one.

ericstj:
I was able to get a callstack for this by adding logging to CSC:
```
[...]

PR #4102, opened 2026-01-08
Add logging and to tryConnectToServerAsync and don't rethrow

ericstj:


ericstj:
Not sure this is feasible - the other exception is occurring in TryLock so we'd need to swallow that one too.  It's possible -- just fallback to CLI tool like is done elsewhere, but that feels like punting.  We can still do that if all else fails.

https://github.com/dotnet/dotnet/issues/4080#issuecomment-3721892876

PR #4107, opened 2026-01-08
Cloak PFX files in external packages

NikolaMilosavljevic:
Resolves s360 violations in VMR for PFX files in `source-build-reference-packages` repo.


src/source-mappings.json
```diff
@@ -90,7 +90,8 @@
             "defaultRemote": "https://github.com/dotnet/source-build-reference-packages",
             "exclude": [
                 // Contains disallowed binaries
-                "src/source-build-reference-packages/src/externalPackages/spectre-console/docs/**/*"
+                "src/source-build-reference-packages/src/externalPackages/spectre-console/docs/**/*",
+                "src/source-build-reference-packages/src/externalPackages/**/*.pfx"
```
MichaelSimons, line 95: IIRC you have to manually remove these files, they won't automatically get removed.

mthalman, line 95: I'm hesitant to make this cloaked across the board for all external repos. If another external repo ends up including a pfx file, I'd rather we be aware of it and can take explicit action versus having it implicitly cloaked.

But something doesn't smell right with the reporting and suppression behavior. We have other cases of pfx files existing in the VMR: https://github.com/dotnet/dotnet/tree/main/src/aspnetcore/src/Shared/TestCertificates. Are these also being reported? They are similarly listed in the suppressions file: https://github.com/dotnet/dotnet/blob/1ecd1919b31db2c76d5f6685e4a5dfdfc0571c14/.config/CredScanSuppressions.json#L151

NikolaMilosavljevic: Yeah, there are 18 issues for aspnetcore PFX files. I am only fixing SBRP files as those aren't used in the build, so cloaking them is fine.

New scanning task does not allow for suppressions of any PFX files. Suppressions for previous CredScan tooling are not utilized.

I'll add a comment that no PFX files are allowed and keep the matching pattern.

NikolaMilosavljevic: Fixed with https://github.com/dotnet/dotnet/pull/4107/commits/08f40a0f165d21eee770c1e1ae70f3cff0b4196a

NikolaMilosavljevic: Added a comment with https://github.com/dotnet/dotnet/pull/4107/commits/a8ce8ad089c5a3c0d96af236149b0ad1ebc9d5cd

Issue #4017, opened 2025-12-29
Rebuild (--no-incremental) should clean the marker / semaphore files


ViktorHofer:
When doing a `dotnet build --no-incremental`, targets shouldn't get skipped in the VMR orchestration layer, i.e. the `RepoBuild` one.

We put those generated marker files on the FileWrites item, so I'm not entirely sure why this doesn't work today: https://github.com/dotnet/dotnet/blob/c93d94a7e854e38eb560698df4b1b3313e4efc13/repo-projects/Directory.Build.targets#L613

Maybe due to the use of the `Microsoft.Build.NoTargets` SDK?

ViktorHofer:
@ericstj shared this in https://github.com/dotnet/dotnet/pull/4130#discussion_r2683084371

> By moving this to AfterTargets="Build" I suspect that the use of FileWrites below will no longer impact the clean files. I think IncrementalClean target is what consumes that item and writes to CleanFile. IncrementalClean is a dependency of CoreBuild.

That makes total sense and we should reconsider the hook points when using the NoTargets SDK in our projects.

PR #4130, opened 2026-01-09
Update the stack to net11.0

ViktorHofer:
Follow-up on https://github.com/dotnet/dotnet/commit/4522825a839dd4bec07cdbbb63f5b54b7745cd3b

Needs the re-bootstrap changes when they are ready.

### DIFF
--- a/src/wpf/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Microsoft/Windows/Controls/Ribbon/Ribbon.cs
+++ b/src/wpf/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Microsoft/Windows/Controls/Ribbon/Ribbon.cs
@@ -901,14 +901,8 @@ namespace Microsoft.Windows.Controls.Ribbon
 
         internal void NotifyTabHeadersScrollOwnerChanged(ScrollViewer oldScrollViewer, ScrollViewer newScrollViewer)
         {
-            if (oldScrollViewer != null )
-            {
-                oldScrollViewer.ScrollChanged -= new ScrollChangedEventHandler(OnTabHeadersScrollChanged);
-            }
-            if (newScrollViewer != null)
-            {
-                newScrollViewer.ScrollChanged += new ScrollChangedEventHandler(OnTabHeadersScrollChanged);
-            }
+            oldScrollViewer?.ScrollChanged -= new ScrollChangedEventHandler(OnTabHeadersScrollChanged);
+            newScrollViewer?.ScrollChanged += new ScrollChangedEventHandler(OnTabHeadersScrollChanged);
         }
 
         private void OnTabHeadersScrollChanged(object d, ScrollChangedEventArgs e)
--- a/src/wpf/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Microsoft/Windows/Controls/Ribbon/RibbonApplicationMenu.cs
+++ b/src/wpf/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Microsoft/Windows/Controls/Ribbon/RibbonApplicationMenu.cs
@@ -291,10 +291,7 @@ namespace Microsoft.Windows.Controls.Ribbon
             base.OnApplyTemplate();
 
             // Cleanup the previous template elements
-            if (_popup != null)
-            {
-                _popup.Opened -= new EventHandler(this.OnPopupOpened);
-            }
+            _popup?.Opened -= new EventHandler(this.OnPopupOpened);
 
             _popup = this.GetTemplateChild(PopupTemplateName) as Popup;
             SubmenuPlaceholder = this.GetTemplateChild(SubmenuTemplateName) as FrameworkElement;
@@ -302,10 +299,7 @@ namespace Microsoft.Windows.Controls.Ribbon
             _footerPaneHost = this.GetTemplateChild(FooterPaneTemplateName) as ContentPresenter;
             _auxiliaryPaneHost = this.GetTemplateChild(AuxiliaryPaneTemplateName) as ContentPresenter;
 
-            if (_popup != null)
-            {
-                _popup.Opened += new EventHandler(this.OnPopupOpened);
-            }
+            _popup?.Opened += new EventHandler(this.OnPopupOpened);
         }
 
         /// <summary>
--- a/src/wpf/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Microsoft/Windows/Controls/Ribbon/RibbonComboBox.cs
+++ b/src/wpf/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Microsoft/Windows/Controls/Ribbon/RibbonComboBox.cs
@@ -521,11 +521,8 @@ namespace Microsoft.Windows.Controls.Ribbon
                 stringFormat = contentControl.ContentStringFormat;
             }
 
-            if (_clonedElement != null)
-            {
-                _clonedElement.LayoutUpdated -= CloneLayoutUpdated;
-                _clonedElement = null;
-            }
+            _clonedElement?.LayoutUpdated -= CloneLayoutUpdated;
+            _clonedElement = null;
 
             if (itemTemplate == null && itemTemplateSelector == null && stringFormat == null)
             {
--- a/src/wpf/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Microsoft/Windows/Controls/Ribbon/RibbonFilterMenuButton.cs
+++ b/src/wpf/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Microsoft/Windows/Controls/Ribbon/RibbonFilterMenuButton.cs
@@ -46,11 +46,7 @@ namespace Microsoft.Windows.Controls.Ribbon
             base.OnApplyTemplate();
 
             RibbonToggleButton filterToggleButton = this.Template.FindName(RibbonMenuButton.ToggleButtonTemplatePartName, this) as RibbonToggleButton;
-
-            if (filterToggleButton != null)
-            {
-                filterToggleButton.Loaded += new RoutedEventHandler(OnFilterToggleButtonLoaded);
-            }
+            filterToggleButton?.Loaded += new RoutedEventHandler(OnFilterToggleButtonLoaded);
         }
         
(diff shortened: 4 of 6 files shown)


src/efcore/src/EFCore.Tasks/EFCore.Tasks.csproj
AndriySvyryd:
```suggestion
  <ItemGroup Condition="'$(TargetFramework)' == '$(NetFrameworkToolCurrent)'">
```

AndriySvyryd:
```suggestion
  <ItemGroup Condition="'$(TargetFramework)' != '$(NetFrameworkToolCurrent)'">
```

src/efcore/test/Directory.Build.props
```diff
@@ -5,7 +5,7 @@
     <NoWarn>$(NoWarn);CA1707;1591;xUnit1000;xUnit1003;xUnit1004;xUnit1010;xUnit1013;xUnit1026;xUnit2013;xUnit1024</NoWarn>
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)..\rulesets\EFCore.test.ruleset</CodeAnalysisRuleSet>
     <LangVersion>preview</LangVersion>
-    <DefaultNetCoreTargetFramework>net10.0</DefaultNetCoreTargetFramework>
+    <DefaultNetCoreTargetFramework>net11.0</DefaultNetCoreTargetFramework>
```
AndriySvyryd: You should be able to just remove this now

src/winforms/src/Common/tests/TestUtilities/TestAccessors.cs
ViktorHofer, line 16: Need to follow-up with @CyrusNajmabadi and @jcouv who added this via https://github.com/dotnet/roslyn/pull/75581

src/winforms/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/SafeArrayScope.cs
```diff
@@ -45,7 +45,7 @@ public SafeArrayScope(SAFEARRAY* value)
         }
         else if (typeof(T) == typeof(int))
         {
-            if (value->VarType is not VARENUM.VT_I4 or VARENUM.VT_INT)
+            if (value->VarType is not VARENUM.VT_I4 and not VARENUM.VT_INT)
```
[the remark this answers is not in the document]
ViktorHofer: wrong

eng/extract-sdk-archive.proj
```diff
@@ -1,11 +1,11 @@
 <Project Sdk="Microsoft.Build.NoTargets">
 
   <PropertyGroup>
-    <TargetFramework>$(NetCurrent)</TargetFramework>
+    <TargetFramework>$(BundledNETCoreAppTargetFramework)</TargetFramework>
   </PropertyGroup>
 
   <Target Name="ExtractSdkArchive"
-          BeforeTargets="Build"
+          AfterTargets="Build"
```
ericstj, line 8: By moving this to `AfterTargets="Build"` I suspect that the use of `FileWrites` below will no longer impact the clean files.  I think `IncrementalClean` target is what consumes that item and writes to `CleanFile`.  `IncrementalClean` is a dependency of `CoreBuild`.

eng/finish-source-only.proj
```diff
@@ -21,7 +21,8 @@
 
   <UsingTask TaskName="Microsoft.DotNet.UnifiedBuild.Tasks.UsageReport.WritePackageUsageData" AssemblyFile="$(MicrosoftDotNetUnifiedBuildTasksAssembly)" TaskFactory="TaskHostFactory" />
   <Target Name="WritePrebuiltUsageData"
-          BeforeTargets="Build"
+          AfterTargets="Build"
```
ericstj, line 24: Same issue with `FileWrites` here and throughout this file.  I would expect we have an opinion about where `NoTargets` projects place the bulk of their `Build` work.  Maybe they hang it off `CoreBuild` or `CoreCompile`?

src/efcore/eng/helix.proj
```diff
@@ -73,7 +73,7 @@
   </ItemGroup>
 
   <PropertyGroup>
-    <XUnitPublishTargetFramework>net10.0</XUnitPublishTargetFramework>
+    <XUnitPublishTargetFramework>$(NetCurrent)</XUnitPublishTargetFramework>
```
ericstj, line 76: Is this redundant to what's defined in helix SDK by default? 
https://github.com/dotnet/arcade/blob/594f3bcba9361908a75c769c1d2d5668facc0dea/src/Microsoft.DotNet.Helix/Sdk/tools/xunit-runner/XUnitRunner.targets#L5-L6

src/runtime/src/libraries/Microsoft.NETCore.Platforms/src/Microsoft.NETCore.Platforms.csproj
```diff
@@ -42,15 +44,13 @@
     <Content Include="$(BaseOutputPath)PortableRuntimeIdentifierGraph.json" PackagePath="/" />
   </ItemGroup>
 
+  <!-- Reference these assemblies from the SDK directly. We do this to avoid bringing the package closure for assemblies we don't use here. -->
   <ItemGroup>
-    <!-- Manually reference these assemblies which are provided by MSBuild / .NET SDK -->
-    <PackageDownloadAndReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkVersion)" Folder="ref/net8.0" />
-    <PackageDownloadAndReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCoreVersion)" Folder="ref/net8.0" />
-    <PackageDownloadAndReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)" Folder="lib/netstandard2.0" />
+    <Reference Include="$(NetCoreSdkRoot)Microsoft.Build.Framework.dll" />
```
ericstj: You might want to set `Private="false"` on these loose binary references.

ericstj: Or maybe create an item like `SDKAssemblyReference` and then have an ItemDefinitionGroup which would apply Private as we did before.  I'm OK with us removing `PackageDownloadAndReference` and replacing with this, so long as we're certain it's always the "Correct" thing to do.  I can imagine it won't work for things that neeed to target NetFx.

ericstj (approved):
The overall changes LGTM.  Feedback is non-blocking and can be addressed here or in followup PR.

src/efcore/eng/helix.proj
ViktorHofer: Efcore could move faster than Helix. Back years before, that property in the Helix SDK didn't get updated early enough. Maybe that's why this got added. cc @AndriySvyryd

eng/extract-sdk-archive.proj
ViktorHofer: That's a great theory and would explain https://github.com/dotnet/dotnet/issues/4017

src/efcore/eng/helix.proj
AndriySvyryd: I think that was indeed the reason. If Helix is moving in lockstep with everything else now then we can remove this

src/winforms/src/System.Windows.Forms/System/Windows/Forms/WindowSubclassHandler.cs
```diff
@@ -97,7 +97,15 @@ public unsafe void Open()
         _originalWindowProc = (void*)PInvokeCore.SetWindowLong(
             _handle,
             WINDOW_LONG_PTR_INDEX.GWL_WNDPROC,
-            (nint)_windowProcDelegatePtr) ?? throw new Win32Exception();
+            (nint)_windowProcDelegatePtr);
+
+#pragma warning disable IDE0270 // Null check can be simplified
```
ViktorHofer, line 103: Need to create a roslyn issue. IDE0270 is suggested here but when applying, I get the following error:

> D:\a\_work\1\s\src\winforms\src\System.Windows.Forms\System\Windows\Forms\WindowSubclassHandler.cs(97,31): error CS0019: Operator '??' cannot be applied to operands of type 'void*' and '<throw expression>' [D:\a\_work\1\s\src\winforms\src\System.Windows.Forms\System.Windows.Forms.csproj]

src/winforms/src/Common/tests/TestUtilities/TestAccessors.cs
ViktorHofer: Filed https://github.com/dotnet/roslyn/issues/81986

src/efcore/src/EFCore.Tasks/Tasks/Internal/OperationTaskBase.cs
akoeplinger: I assume this needs to be updated as well? maybe the #if should be moved closer to this line?
ideally we'd be able to remove the version number and make it generic

src/wpf/eng/WpfArcadeSdk/tools/SdkReferences.targets
akoeplinger: nit: in comment cases like this I'd prefer if we just say `netX.0-windows` so we don't need to bump this for all eternity.

src/winforms/src/System.Windows.Forms/System/Windows/Forms/WindowSubclassHandler.cs
ViktorHofer: https://github.com/dotnet/roslyn/issues/81987

akoeplinger (approved):

eng/extract-sdk-archive.proj
ViktorHofer: Following up in another PR.

ViktorHofer:
The Ubuntu job passed before and is hit by flakiness in roslyn's build:

> /__w/1/s/.dotnet/sdk/11.0.100-alpha.1.26060.102/NuGet.Build.Tasks.Pack.targets(222,5): error : The process cannot access the file '/__w/1/s/src/roslyn/artifacts/packages/Release/Shipping/Microsoft.VisualStudio.Extensibility.Testing.SourceGenerator.5.4.0-ci.nupkg' because it is being used by another process. [/__w/1/s/src/roslyn/src/VisualStudio/IntegrationTest/Harness/SourceGenerator/Microsoft.VisualStudio.Extensibility.Testing.SourceGenerator.csproj]

The `iOSSimulator_Shortstack_arm64` job has a slow mac and will time out. Also passed previously.

PR #4131, opened 2026-01-09
[release/10.0.1xx] Support maui/android/ios workload manifests for source build

mthalman:
### Description

This adds supports for Maui workload manifests for Source Build. This is done by including the workload manifest packages in the source-build-reference-packages repo where they are built and packaged and made available to be referenced by the SDK.

Backport of #3897
Issue: https://github.com/dotnet/source-build/issues/3242

### Customer Impact

This allows users that have acquired a source built version of the .NET SDK (such as via Red Hat or Ubuntu) to have Maui-related workloads show up as available to install. Upon installation, the workloads are retrieved from NuGet which were built and published by Microsoft (i.e. not source built) but the workload _manifests_ which reveal these workloads to be available for install _are_ source built and included in the SDK.

### Regression?

Not a regression. This is new functionality.

### Risk

Low - This simply exposes the workloads to be available for install. The actual workloads are exactly the same as what already exist for non-source built SDK. Manually tested that installation works and you can create/build a new Maui project.


src/sdk/eng/Versions.props
```diff
@@ -142,6 +142,12 @@
...
   <PropertyGroup Label="Workload manifest package versions">
     <MauiFeatureBand>10.0.100</MauiFeatureBand>
+    
+    <!--
+      Updates to any of these versions will require the updating of that package version in the
+      source-build-reference-packages repo as a text-only package.
+      See https://github.com/dotnet/source-build-reference-packages?tab=readme-ov-file#text-only
+    -->
     <MauiWorkloadManifestVersion>10.0.0</MauiWorkloadManifestVersion>
     <XamarinAndroidWorkloadManifestVersion>36.1.2</XamarinAndroidWorkloadManifestVersion>
```
rmarinho, line 152: Can we change this name to just `AndroidWorkloadManifestVersion` ? drop the xamarin prefix

mthalman: That is outside the scope of these changes. These are already existing properties.

PR #4136, opened 2026-01-09
Comment out NuGet command addition in Parser.cs

baronfel:
Comment out the addition of NuGet commands to the root command to test if NuGet command initialization is impacting the output of `dotnet --info`


marcpopMSFT:
Real fix was https://github.com/dotnet/dotnet/pull/4139, closing

PR #4139, opened 2026-01-09
lazy initialize NuGet's instance of AnsiConsole

zivkan:
fixes: https://github.com/dotnet/sdk/issues/52278

Apparently using Spectre.Console's `AnsiConsole.Console` sends VT control codes on initialize, not first API usage, and this is messing up some of the dotnet CLI's C++ output, like `dotnet --info`

So, NuGet should not call `AnsiConsole.Console` until it actually needs it. Lazy<T> should do it.


aortiz-msft:
Could we please link this to an issue?

MiYanni:
> Could we please link this to an issue?

https://github.com/dotnet/sdk/issues/52278

zivkan:
I did a very basic manual test, and it looks to me like it fixed the issue.

PR #4169, opened 2026-01-13
Workload support

mthalman:
Port workload support from 10.0.2xx:
* https://github.com/dotnet/dotnet/pull/3705
* https://github.com/dotnet/dotnet/pull/4166
* https://github.com/dotnet/dotnet/pull/4210
* https://github.com/dotnet/dotnet/pull/4424

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -44,7 +44,7 @@
     <PrivateSourceBuiltSdkVersion>11.0.100-alpha.1.26064.118</PrivateSourceBuiltSdkVersion>
     <PrivateSourceBuiltArtifactsVersion>11.0.100-alpha.1.26064.118</PrivateSourceBuiltArtifactsVersion>
     <!-- workload manifest version for 1xx SDK feature band, used by 2xx+ branches to include 1xx workloads -->
-    <DotNet1xxWorkloadManifestVersion>$(MicrosoftNETWorkloadEmscriptenCurrentManifest110100Version)</DotNet1xxWorkloadManifestVersion>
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -43,6 +43,8 @@
     -->
     <PrivateSourceBuiltSdkVersion>11.0.100-alpha.1.26064.118</PrivateSourceBuiltSdkVersion>
     <PrivateSourceBuiltArtifactsVersion>11.0.100-alpha.1.26064.118</PrivateSourceBuiltArtifactsVersion>
+    <!-- workload manifest version for 1xx SDK feature band, used by 2xx+ branches to include 1xx workloads -->
+    <DotNet1xxWorkloadManifestVersion>$(MicrosoftNETWorkloadEmscriptenCurrentManifest100100Version)</DotNet1xxWorkloadManifestVersion>
```
akoeplinger: this will need to be updated to have the 110100 version for main. or maybe we can make this generic somehow?

repo-projects/sdk.proj
```diff
@@ -18,8 +18,10 @@
 
     <BuildArgs Condition="'$(PgoInstrument)' == 'true'">$(BuildArgs) /p:PgoInstrument=true</BuildArgs>
     <BuildArgs Condition="'$(EnablePoison)' == 'true' or '$(SkipUsingCrossgen)' == 'true'">$(BuildArgs) /p:SkipUsingCrossgen=true</BuildArgs>
-
+
```
akoeplinger: nit: adding whitespace

eng/Versions.props
mthalman: > or maybe we can make this generic somehow?

`DotNet1xxWorkloadManifestVersion` is intended to be the generic property. That's why it's here so that we can avoid referencing the version-specific one elsewhere. But it needs to reference the autogenerated property name which includes these dang version numbers in the package name. So I don't know of a better way to do this.

PR #4181, opened 2026-01-14
Add Re-Bootstrapping Doc to the VMR

ellahathaway:
Related to https://github.com/dotnet/release/issues/1564

Moves and updates the doc from https://github.com/dotnet/source-build/blob/main/Documentation/VMR-re-bootstrapping.md

### DIFF
--- a/docs/VMR-re-bootstrapping.md
+++ b/docs/VMR-re-bootstrapping.md
@@ -33,6 +33,11 @@ released.
 > Eligible builds must have published artifacts to a public channel.
 > For this reason, re-bootstrapping is only allowed with official builds
 > from the main branch and non-internal release branches.
+>
+> Additionally, re-bootstrapping should only be done with signed builds.
+> In exceptional cases (e.g., during early development), re-bootstrapping
+> with an unsigned build may be necessary, but this should be avoided
+> unless absolutely required.
 
 ### Automated
 


docs/VMR-re-bootstrapping.md
```diff
@@ -0,0 +1,68 @@
+# How to re-bootstrap the toolset used to build the VMR
+
+.NET utilizes itself to build. Therefore, in order to build .NET from source,
```
ViktorHofer: This applies to non-source-build as well. Maybe ".NET SDK"?

```suggestion
.NET utilizes itself to build. Therefore, in order to build the .NET SDK,
```

docs/VMR-re-bootstrapping.md
```diff
@@ -0,0 +1,68 @@
...
+of when re-bootstrapping is allowed, means that repos should, in general, wait
+to take a dependency on a new toolset feature until after that feature has been
+released.
+
+## Steps to re-bootstrap
+
+> [!IMPORTANT] 
+> Eligible builds must have published artifacts to a public channel.
+> For this reason, re-bootstrapping is only allowed with official builds
+> from the main branch and non-internal release branches.
```
ViktorHofer, line 35: Maybe mention that the build used for re-bootstrapping usually should be signed?

ellahathaway: Does it matter if we used a signed build or not? We previously cared because we wanted to avoid uploading unsigned artifacts to blob storage, but with UB we publish all official builds via darc regardless of signing status.

ViktorHofer: Yes, as using a non-signed SDK breaks the developer inner-loop, i.e. with vscode. We learned that last year when we re-bootstrapped to non-signed builds and got complaints from runtime devs.

ellahathaway: Gotcha. Is there ever a case when we would prefer an unsigned build? I ask because I would prefer to have it be a rule rather than a suggestion that re-bootstrap builds must be signed.

ViktorHofer: Yes, there are exceptional cases when re-bootstrapping with an unsigned build is necessary. But by default, it shouldn't be allowed. I really liked this UX in the re-bootstrap pipeline:

Image

docs/VMR-re-bootstrapping.md
```diff
@@ -0,0 +1,68 @@
...
+## When is it appropriate to re-bootstrap?
+
+As part of the release process, the toolset is updated (e.g. PRs are created via
+the release automation). Outside of a release, re-bootstrapping is only
+permitted during preview releases. It is not allowed during RC, GA, or servicing
+releases. The reason it is not allowed during non-preview releases is because of
+the negative impact it has on Linux distro maintainers who source build .NET. It
+is often a long and time-consuming process for them to re-bootstrap. It is
+likely to cause significant delays in the release/availability of .NET within
+the distros that are source built.
```
MichaelSimons: Maybe this can be emphasized more.

```suggestion
the distros that are source built which can "zero-day" customers when security fixes are involved.
```

docs/VMR-re-bootstrapping.md
ellahathaway: Addressed in [this commit](https://github.com/dotnet/dotnet/pull/4181/commits/e7bb4b19d3c8f57caa0d0a953ac717b39691e83c)

PR #4195, opened 2026-01-14
Fix missing metadata in GenerateBundledVersions.targets and a few additional fixes in templates, etc

ViktorHofer:
Diffed the produced BundledVersions.props file with the one from 10.0.102. Nothing else stood out.

> Remove specific RID entries from Net100MonoRuntimePackRids (linux-riscv64, linux-musl-riscv64, linux-loongarch64, linux-musl-loongarch64)

They were duplicated as they are already present in Net90MonoRuntimePackRids


ViktorHofer:
The iOSSimulator is a slow mac - known infra issue

PR #4204, opened 2026-01-14
Fix roslyn / razor area ownership

jaredpar:


jaredpar:
@premun FYI

premun:
Great, I will also add these tomorrow to have write access in the VMR

PR #4247, opened 2026-01-16
Remove NetToolCurrent and NetToolMinimum properties

ViktorHofer:
The SDK now has a property (BundledNETCoreAppTargetFramework) that provides the same functionality as NetToolCurrent but more accurately during retargeting. I expect that wpf, winforms, sourcelink, xdt, symreader, etc. (all non-tooling repos that aren't on the critical path during retargeting) to update to that property eventually.

Also remove NetToolMinimum in favor of NetMinimum.

There are two repositories that use this property that aren't part of the VMR which I will fix in the dependency flow PRs then.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj
@@ -64,8 +64,8 @@
 
     <ItemGroup>
       <!-- .pkgs and .app bundle tooling is only available on MacOS -->
-      <MacOSPkgCliItem Condition="$([MSBuild]::IsOSPlatform('OSX'))" Include="$(NuGetPackageRoot)microsoft.dotnet.macospkg.cli\$(MicrosoftDotNetMacOsPkgVersion)\tools\net1*\any\Microsoft.Dotnet.MacOsPkg.Cli.dll" />
-      <TarToolItem Include="$(NuGetPackageRoot)microsoft.dotnet.tar\$(MicrosoftDotNetTarVersion)\tools\net1*\any\Microsoft.Dotnet.Tar.dll" />
+      <MacOSPkgCliItem Condition="$([MSBuild]::IsOSPlatform('OSX'))" Include="$(NuGetPackageRoot)microsoft.dotnet.macospkg.cli\$(MicrosoftDotNetMacOsPkgVersion)\tools\net*\any\Microsoft.Dotnet.MacOsPkg.Cli.dll" />
+      <TarToolItem Include="$(NuGetPackageRoot)microsoft.dotnet.tar\$(MicrosoftDotNetTarVersion)\tools\net*\any\Microsoft.Dotnet.Tar.dll" />
     </ItemGroup>
 
     <Error Condition="'$(AllowEmptySignList)' != 'true' AND '@(ItemsToSign)' == ''" 


src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj
```diff
@@ -60,11 +60,14 @@
...
       <SNBinaryPath Condition="$([MSBuild]::IsOSPlatform('Windows'))">$(NuGetPackageRoot)sn\$(SNVersion)\sn.exe</SNBinaryPath>
-      
-      <!-- .pkgs and .app bundle tooling is only available on MacOS -->
-      <PkgToolPath Condition="$([MSBuild]::IsOSPlatform('OSX'))">$(NuGetPackageRoot)microsoft.dotnet.macospkg.cli\$(MicrosoftDotNetMacOsPkgVersion)\tools\$(NetToolCurrent)\any\Microsoft.Dotnet.MacOsPkg.Cli.dll</PkgToolPath>
     </PropertyGroup>
 
+    <ItemGroup>
+      <!-- .pkgs and .app bundle tooling is only available on MacOS -->
+      <MacOSPkgCliItem Condition="$([MSBuild]::IsOSPlatform('OSX'))" Include="$(NuGetPackageRoot)microsoft.dotnet.macospkg.cli\$(MicrosoftDotNetMacOsPkgVersion)\tools\net1*\any\Microsoft.Dotnet.MacOsPkg.Cli.dll" />
+      <TarToolItem Include="$(NuGetPackageRoot)microsoft.dotnet.tar\$(MicrosoftDotNetTarVersion)\tools\net1*\any\Microsoft.Dotnet.Tar.dll" />
```
akoeplinger: this is gonna bite us when net20.0 hits 😄 

I checked and both microsoft.dotnet.tar and microsoft.dotnet.macospkg.cli only pack the tool for netcore so we can just use `net*` here (fine in a follow-up PR so you don't need to restart CI)

src/arcade/tests/XHarness.Tests.Common.props
```diff
@@ -7,7 +7,7 @@
    -->
 
   <PropertyGroup>
-    <MicrosoftDotNetHelixSdkTasksAssembly Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)../artifacts/bin/Microsoft.DotNet.Helix.Sdk/$(Configuration)/$(NetToolCurrent)/publish/Microsoft.DotNet.Helix.Sdk.dll</MicrosoftDotNetHelixSdkTasksAssembly>
+    <MicrosoftDotNetHelixSdkTasksAssembly Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)../artifacts/bin/Microsoft.DotNet.Helix.Sdk/$(Configuration)/$(NetCurrent)/publish/Microsoft.DotNet.Helix.Sdk.dll</MicrosoftDotNetHelixSdkTasksAssembly>
```
akoeplinger: and here?

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj
ViktorHofer: Thanks. Updated. FYI, the netframework paths will be deleted with https://github.com/dotnet/arcade/pull/16413 eventually anyways.

PR #4251, opened 2026-01-17
[release/10.0.3xx] Initialize SDK band versions

mmitche:
Initialize SDK band versions on release/10.0.3xx.

Updated repositories:
- sdk: sdkminor 2 → 3, patch 00 → 00
- sourcelink: patch 200 → 300


ViktorHofer:
What about templating?

mmitche:
https://github.com/dotnet/dotnet/blob/release/10.0.3xx/src/templating/eng/Versions.props#L4

Already 10.0.300.

PR #4279, opened 2026-01-20
Use VS2026 to build on Windows + Source updates from dotnet/wpf

akoeplinger:


ViktorHofer:
wpf probably needs to get merged into this PR

akoeplinger:
hmm yeah but I'm a bit confused how aspnetcore built given we also set `<PlatformToolsetVersion>v143</PlatformToolsetVersion>` there

ViktorHofer:
@akoeplinger looks like Cpp props and targets already got improved to try the default version when the requested version isn't found:

image

image

akoeplinger:
yes but why does this work for aspnetcore but not wpf?

ViktorHofer:
Probably because wpf has a few hardcodes: https://github.com/dotnet/wpf/blob/6e82b1f838430a7bcb7648c3aaf82a6d2f626c2b/src/Microsoft.DotNet.Wpf/redist/VCRuntime/VCRuntime.vcxproj#L36-L37

akoeplinger:
I don't think so as that's just for redisting, but as long as it works with the new VS ¯\\_(ツ)_/¯

ViktorHofer:
The passing aspnetcore build:

image

The failing wpf build:

image

ViktorHofer:
Ok works for aspnetcore because they explicitly set `PlatformToolsetVersion`

ViktorHofer:
The previous run was already successful (the iOS job timed out). I don't know why AzDO requeued the build but that's unnecessary. Merging.

akoeplinger:
> Ok works for aspnetcore because they explicitly set PlatformToolsetVersion

Interesting. aspnetcore sets both PlatformToolset and PlatformToolsetVersion: https://github.com/dotnet/aspnetcore/blob/fda5da455a76554283e08c35365fcfa66978ff28/eng/targets/Cpp.Common.props#L23-L24

Issue #3421, opened 2025-11-18
Upgrade Windows pools to VS2026 and remove NETSDK1233 workaround


ellahathaway:
.NET 10 requires VS 2026. We were unable to upgrade from VS 2022 to VS 2026 in the .NET 10 GA rebootstrap PR due to the following error while using the VS 2026 images:
```
error MSB8052: MSVC Toolset Version '14.50.35503' is not compatible with 'v143' Platform Toolset. Please either change Platform Toolset to v145 or MSVC Toolset Version (VCToolsVersion property) to  the version with the format '14.3*.*'. To use default MSVC Toolset Version for a given Platform Toolset don't set VCToolsVersion property.
```

We should upgrade our windows images to use VS 2026 when @dotnet/dnceng says they're ready. With this upgrade, remove the NETSDK1233 error workaround (overwriting `_WarnWhenUsingNET10AndVSPriorTo18`) introduced in https://github.com/dotnet/dotnet/pull/3298/commits/eee59941d6e270f6215399c6833418a5d34e294c.

PR #4321, opened 2026-01-21
Remove vs2022 workaround

ViktorHofer:
Fixes https://github.com/dotnet/dotnet/issues/3415
Fixes https://github.com/dotnet/dotnet/issues/3421

### DIFF
--- a/eng/vs-workaround.targets
+++ b/eng/vs-workaround.targets
@@ -1,8 +1,0 @@
-<!-- Temporary workaround to silence that SDK warning. This file needs to be imported via the `AfterMicrosoftNETSdkTargets` property.
-     TODO: Remove this file when a fixed VS 2026 image is used: https://github.com/dotnet/dotnet/issues/3415. -->
-
-<Project>
-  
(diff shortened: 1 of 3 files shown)


akoeplinger:
should we keep the issues open until we've backported the VS2026 to the 10.0 branches?

ViktorHofer:
There's not really any value in backporting the workaround. It doesn't fix anything. We can but I don't feel the need for it in non-main.

akoeplinger:
@ViktorHofer I meant backporting _removing_ the workaround since afaik we do want to make the switch to VS2026 in 10.0 too right?

ViktorHofer:
The changes in this PR aren't necessary to switch to VS2026. Removing the workaround is nice to have but doesn't impact anything. It just disables a validation target in sdk that we don't care about for 10.0. If you want to backport it, go for it but I don't think it's necessary. Especially as it would need three PRs for 1xx, 2xx and 3xx.

akoeplinger:
ok thanks. I wasn't sure whether the workaround would have some unintended interactions with 2026

PR #4352, opened 2026-01-22
Add F# version overrides to 10.0.2xx

T-Gro:
dotnet/fsharp main now flows to:
main
10.0.2xx
10.0.3xx

The version numbering overrides will allow to customize package and product branding without raising git flow conflicts.

### DIFF
--- a/src/fsharp/eng/Versions.props
+++ b/src/fsharp/eng/Versions.props
@@ -26,7 +26,7 @@
 
     <!-- BEGIN - SPECIFIC OVERRIDES for 10.0.2xx. Those are overrides instead of updates to reduce risk of git conflicts in codeflow -->
     <FSharpPreReleaseIteration></FSharpPreReleaseIteration>
-    <PreReleaseVersionLabel>$(FSharpPreReleaseIteration)</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>servicing$(FSharpPreReleaseIteration)</PreReleaseVersionLabel>
     <FSMajorVersion>10</FSMajorVersion>
     <FSMinorVersion>0</FSMinorVersion>
     <FSBuildVersion>200</FSBuildVersion>


ViktorHofer:
> dotnet/fsharp main now flows to:
10.0.2xx

Don't you already have a vs18.3 / dev18.3 branch that flows into 10.0.2xx? 18.3 got snapped like two weeks ago and is already in QB mode.

T-Gro:
@ViktorHofer : With the evergreen model, we will only do dedicated branches for QB / Tactics just in time as needed, our QB cadency is far below 1/month (more like 1x or 2x a year).

(I hope this is not heavily deviating or process-breaking for some automatism)

abonie (approved):

src/fsharp/eng/Versions.props
```diff
@@ -22,6 +22,18 @@
     <FSBuildVersion>100</FSBuildVersion>
     <FSRevisionVersion>0</FSRevisionVersion>
     <!-- -->
+
+
+    <!-- BEGIN - SPECIFIC OVERRIDES for 10.0.2xx. Those are overrides instead of updates to reduce risk of git conflicts in codeflow -->
+    <FSharpPreReleaseIteration></FSharpPreReleaseIteration>
+    <PreReleaseVersionLabel>$(FSharpPreReleaseIteration)</PreReleaseVersionLabel>
```
T-Gro:
```suggestion
    <PreReleaseVersionLabel>servicing$(FSharpPreReleaseIteration)</PreReleaseVersionLabel>
```

src/fsharp/eng/Versions.props
```diff
@@ -22,6 +22,18 @@
     <FSBuildVersion>100</FSBuildVersion>
     <FSRevisionVersion>0</FSRevisionVersion>
     <!-- -->
+
+
+    <!-- BEGIN - SPECIFIC OVERRIDES for 10.0.2xx. Those are overrides instead of updates to reduce risk of git conflicts in codeflow -->
+    <FSharpPreReleaseIteration></FSharpPreReleaseIteration>
```
mmitche, line 28: This is probably going to cause havoc with the auto-rebranding tooling. I think I can work around it but not sure.

Also, since 10.0.2xx is supposed to go to RTM soon, isn't the right move to branch fsharp so that it has a specific 10.0.2xx corresponding branch? It shouldn't get 11.0 bits

T-Gro: Once dotnet/fsharp/main can no longer flow to dotnet/dotnet/10.0.2xx, this is no longer needed, I agree.

These overrides (there is a sibling branch for .3xx) are only for multiple source dotnet/dotnet branches for incoming contributions to dotnet/fsharp

PR #4381, opened 2026-01-23
Support internal feeds

mthalman:
The builds for 2xx+ require special consideration for the scenario where assets need to be pulled from internal locations.

- Flow package feed token to the build (specifically, the build of the `sdk` repo). This allows the [redist assets to be downloaded from the internal location](https://github.com/dotnet/dotnet/blob/1966e459956c1e4fb9a3714e3c5a7b5e42c15585/src/sdk/src/Layout/redist/targets/RestoreLayout.targets#L306-L311).
- Each product repo of the VMR may have a NuGet.config file targeting private feeds (which are disabled by default). These need to be enabled by running `eng/common/SetupNugetSources.ps1`. The pipeline is updated to enumerate all these repos and run that script.
- For source build tests, they need to download WASM dependencies which can exist in an internal source. The NuGet restore for this scenario is updated to use the same NuGet as the other source build tests (which is correctly configured to target internal locations).


eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -459,6 +465,27 @@ jobs:
...
+            $repoDir = $_.FullName
+            $setupScript = Join-Path $repoDir "eng/common/SetupNugetSources.ps1"
+            $nugetConfig = Join-Path $repoDir "NuGet.config"
+            
+            Write-Host "Setting up internal NuGet feeds for $($_.Name)"
+            & $setupScript $nugetConfig $token
+            Write-Host ""
+          }
+      env:
+        Token: $(dn-bot-dnceng-artifact-feeds-rw)
```
mmitche, line 487: You shouldn't need the token here. When running on dnceng/internal today we call NuGetAuthenticate and then the script to enable the feeds. We won't be able to use PATs forever so we should avoid building a bit of debt here.

mthalman: There are issues with that which I don't want to block this PR on. So I logged https://github.com/dotnet/dotnet/issues/4385

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -657,14 +687,6 @@ jobs:
       # Setup the NuGet sources used by the tests to use private feeds. This is necessary when testing internal-only product
       # builds where the packages are only available in the private feeds. This allows the tests to restore from those feeds.
```
akoeplinger, line 688: nit: now that the "Setup Private Feeds Credentials" doesn't happen here anymore this comment is a bit harder to understand. Maybe mention that the feed sources were configured earlier and that we just take it from the sdk nuget.config

PR #4389, opened 2026-01-23
Reset roslyn to 298541 for 10.0.3xx

mmitche:


333fred:
@mmitche how are we supposed to reason about this change? It's a giant blob of updates with no changelog.

PR #4414, opened 2026-01-26
Use AzureLinux3 image for the container host

akoeplinger:
AZL is the recommended image and makes our setup consistent between x64 and arm64.

Most builds run inside of containers anyway so this is a no-op.


akoeplinger:
> ##[error]Failed to request agent. Exception Image Azure-Linux-3-Amd64-Public doesn't exist in pool NetCore-Public-XL

This will get fixed with the next dnceng rollout, parking this PR for now.

PR #4420, opened 2026-01-26
Refactor logging and error handling in build targets

### LINKED ISSUE #705: Warnings from inner repo builds are never surfaced unless there's a failure
Due to the `MinimalConsoleLog` feature, warnings emitted by VMR inner repo builds never make it to the console and AzDO. That effectively makes them non-existent.

This affects repos that don't set TreatWarningsAsErrors=true / -warnaserror so warnings never get promoted to errors.

One options would be to globally pass /p:TreatWarningsAsErrors=true to all VMR inner repos (with a way to temporary exclude repos). That might solve the immediate need but some warnings are explicitly disabled from getting promoted to errors via `WarningsNotAsErrors` so we would still need a way to bubble those up.

ViktorHofer:
Fixes https://github.com/dotnet/dotnet/issues/705

Removes the minimal console log switch.
Use the `StandardOutputImportance` parameter to only log warnings and errors (with context).

### DIFF
--- a/repo-projects/Directory.Build.targets
+++ b/repo-projects/Directory.Build.targets
@@ -47,10 +47,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <!-- MinimalConsoleLogOutput determines if the repository build should be logged to a separate file or directly to the console.
-         Enable it by default as with prallel repository builds the log becomes non-readable. -->
     <RepoConsoleLogFile>$(ArtifactsLogRepoDir)$(RepositoryName).log</RepoConsoleLogFile>
-    <MinimalConsoleLogOutput Condition="'$(MinimalConsoleLogOutput)' == ''">$(BuildInParallel)</MinimalConsoleLogOutput>
   </PropertyGroup>
 
   <!-- Exclude shared repositories from the build, unless shared component building is explicitly enabled. -->
@@ -589,7 +586,7 @@
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Building $(RepositoryName)" />
     <Message Importance="High" Text="Running command:" />
     <Message Importance="High" Text="  $(BuildCommand)" />
-    <Message Importance="High" Text="  Log: $(RepoConsoleLogFile)" Condition="'$(MinimalConsoleLogOutput)' == 'true'" />
+    <Message Importance="High" Text="  Log: $(RepoConsoleLogFile)" />
     <Message Importance="High" Text="  With Environment Variables:"/>
     <Message Importance="High" Text="    %(EnvironmentVariables.Identity)" />
     <Message Importance="High" Text="    %(BuildEnvironmentVariable.Identity)" Condition="'@(BuildEnvironmentVariable)' != ''" />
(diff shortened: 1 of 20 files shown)


akoeplinger:
@ViktorHofer do I understand it correctly that this would mean we no longer log the full repo build log in case of a failure?

ViktorHofer:
Yes, only the relevant portion of the warning or failure (+ context) gets logged to the console. The full console log still gets written to the .log and the .binlog files. That essentially mirrors the behavior that you get in a dotnet build scenario.

repo-projects/Directory.Build.targets
```diff
@@ -589,7 +586,7 @@
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Building $(RepositoryName)" />
     <Message Importance="High" Text="Running command:" />
     <Message Importance="High" Text="  $(BuildCommand)" />
-    <Message Importance="High" Text="  Log: $(RepoConsoleLogFile)" Condition="'$(MinimalConsoleLogOutput)' == 'true'" />
+    <Message Importance="High" Text="  Log: $(RepoConsoleLogFile)" />
```
rainersigwald, line 589: I'd probably log the binlog path instead?

rainersigwald (approved):

akoeplinger:
@ViktorHofer then I'm not in favor of this change, the error message alone almost never tells you the full context and needing to download the build log all the time is cumbersome (Jared complained about this exact thing)

ViktorHofer:
This change makes it so that warnings and errors propagate to the summary view and the AzDO timeline. That's in response to what Jared asked - and to fix warnings not getting emitted. For anything more fidelity, you need to take a look at the log or the binlog.

ViktorHofer:
We can have both if we want that but the intent here was to log what's useful from an inner build so that it's easier to find the actual error (instead of scroll scroll scroll ... repeat 1000x)

mmitche:
Does the removal of the minimal console log switch mean that it will no longer be possible to print the running log? If so, then this won't work for some source-build partners. There are some that have system that watch console log output to attempt to determine whether a build is frozen or not. They were having issues getting a build to complete without the running log.

ViktorHofer:
Thanks, I forgot to make this configurable. The syntax is `/p:RepoStandardOutputImportance=high`. By default it's low which means warnings+errors only. https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tooltask.standardoutputimportance?view=msbuild-17-netcore

T-Gro (approved):

src/nuget-client/eng/dotnet-build/build.sh
```diff
@@ -119,7 +129,12 @@ dotnetArguments+=("/p:DotNetBuild=$product_build")
 dotnetArguments+=("/p:DotNetBuildSourceOnly=$source_build")
 dotnetArguments+=("/p:DotNetBuildFromVMR=$from_vmr")
 
-local bl=""
+warnaserror_switch=""
+if [[ $warn_as_error == true ]]; then
+  warnaserror_switch="/warnaserror"
```
zivkan: MSBuild `/warnaserror` has different semantics than the msbuild property `<TreatWarningsAsErrors>`, in particular it doesn't respect `<WarningsNotAsErrors>`, which we use to avoid NuGetAudit warnings breaking the build, while still getting the warnings. 

I removed this from NuGet a few months ago to unblock our CI after a patch tuesday that disrupted us: https://github.com/NuGet/NuGet.Client/pull/6867

ViktorHofer: Every "arcadified" repository uses this though: https://github.com/dotnet/arcade/blob/6dbcc0cd8057e7554a43ab29f2a252c670dc96a0/eng/common/tools.sh#L525

Why is nuget-client special? In particular when thinking about onboarding nuget-client to arcade.
I reverted my change meanwhile, thanks for letting me know.

I submitted https://github.com/NuGet/NuGet.Client/pull/7084 to make sure that we are better aligned on future `eng/dotnet-build/` changes in nuget-client going forward.

ViktorHofer: Btw I'm now on the MSBuild team and happy to look into this further. Which of the sub-issues are you referring to here in https://github.com/dotnet/msbuild/issues/10801?

repo-projects/Directory.Build.targets
ViktorHofer: We are just passing `-bl` in to the inner repo build so we don't know the exact name upfront. Some repos also produce multiple binlogs.

akoeplinger:
> That's in response to what Jared asked - and to fix warnings not getting emitted. For anything more fidelity, you need to take a look at the log or the binlog.

We don't have warnings anymore now that every repo uses warnAsError right? And Jared specifically mentioned that it takes a lot of clicks and is not obvious how to download the correct log artifact before we showed him the "scroll to the log output" approach.

I'm happy to give this a try and let me convince that this is better, but it'd be helpful if you could run a test build that introduces errors/warnings in places like C++ code, msbuild logic, custom bash/ps1 scripts like build-runtime.cmd etc so that we can take a look how these are surfaced

src/nuget-client/eng/dotnet-build/build.sh
nkolev92: > Why is nuget-client special? In particular when thinking about onboarding nuget-client to arcade.

Because we use audit sources, and we get vulnerability warnings.
https://github.com/dotnet/msbuild/issues/10801 is an issue for us because MSBuild is elevating a warning it has no business elevating.

ViktorHofer:
Closing. This got implemented via https://github.com/dotnet/dotnet/commit/0b5109c67228e2566ba35f47607106f97e62cc8a but in a largely different and non-breaking way.

PR #4424, opened 2026-01-26
Replace workload manifest dependency with SDK version parsing

mthalman:
This removes the dependency on `Microsoft.NET.Workload.Emscripten.Current.Manifest-10.0.100` which would have required maintenance over time due to the version being embedded in the package name. Instead the workload manifest version is derived from the 1xx SDK version.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -38,7 +38,7 @@
     <PrivateSourceBuiltSdkVersion>10.0.102-servicing.25612.105</PrivateSourceBuiltSdkVersion>
     <PrivateSourceBuiltArtifactsVersion>10.0.102-servicing.25612.105</PrivateSourceBuiltArtifactsVersion>
     <!-- workload manifest version for 1xx SDK feature band, used by 2xx+ branches to include 1xx workloads -->
-    <DotNet1xxWorkloadManifestVersion>$(MicrosoftNETWorkloadEmscriptenCurrentManifest100100Version)</DotNet1xxWorkloadManifestVersion>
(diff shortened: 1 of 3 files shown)


eng/Versions.props
```diff
@@ -38,7 +38,7 @@
     <PrivateSourceBuiltSdkVersion>10.0.102-servicing.25612.105</PrivateSourceBuiltSdkVersion>
     <PrivateSourceBuiltArtifactsVersion>10.0.102-servicing.25612.105</PrivateSourceBuiltArtifactsVersion>
     <!-- workload manifest version for 1xx SDK feature band, used by 2xx+ branches to include 1xx workloads -->
-    <DotNet1xxWorkloadManifestVersion>$(MicrosoftNETWorkloadEmscriptenCurrentManifest100100Version)</DotNet1xxWorkloadManifestVersion>
+    <DotNet1xxWorkloadManifestVersion>$([System.Text.RegularExpressions.Regex]::Match($(MicrosoftNETSdkVersion), '^(\d+\.\d+\.\d+)').Value)</DotNet1xxWorkloadManifestVersion>
```
mmitche, line 41: Will `MicrosoftNETSdkVersion` always be 1xx based? If 3xx requires a 2xx SDK to build, will that mean an update to this property?

mthalman: > Will `MicrosoftNETSdkVersion` always be 1xx based? If 3xx requires a 2xx SDK to build, will that mean an update to this property?

`MicrosoftNETSdkVersion` will always be 1xx based. It doesn't represent which SDK will be used to build. It represents the SDK build from where the shared components are retrieved.

mmitche: Perfect. Sounds good.

PR #4432, opened 2026-01-27
[main] Update branding to preview2

mmitche:
- aspnetcore: preview1 -> preview2
- command-line-api: preview1 -> preview2
- efcore: preview1 -> preview2
- emsdk: preview1 -> preview2
- fsharp: preview0 -> preview1
- runtime: preview1 -> preview2
- sdk: preview1 -> preview2
- templating: preview1 -> preview2
- windowsdesktop: preview1 -> preview2
- winforms: preview1 -> preview2
- wpf: preview1 -> preview2

### DIFF
--- a/src/fsharp/eng/Versions.props
+++ b/src/fsharp/eng/Versions.props
@@ -14,7 +14,7 @@
   <PropertyGroup>
     <!-- Don't use the built in support for pre-release iteration. The nuget repack task doesn't support
          the iteration format at the moment. https://github.com/dotnet/arcade/issues/15919 -->
-    <FSharpPreReleaseIteration>1</FSharpPreReleaseIteration>
+    <FSharpPreReleaseIteration>2</FSharpPreReleaseIteration>
     <PreReleaseVersionLabel>preview$(FSharpPreReleaseIteration)</PreReleaseVersionLabel>
     <!-- F# Version components -->
     <FSMajorVersion>11</FSMajorVersion>


src/fsharp/eng/Versions.props
```diff
@@ -14,7 +14,7 @@
   <PropertyGroup>
     <!-- Don't use the built in support for pre-release iteration. The nuget repack task doesn't support
          the iteration format at the moment. https://github.com/dotnet/arcade/issues/15919 -->
-    <FSharpPreReleaseIteration></FSharpPreReleaseIteration>
+    <FSharpPreReleaseIteration>1</FSharpPreReleaseIteration>
```
akoeplinger: is this correct given the comment above? @T-Gro

T-Gro: imo the comment is the reason why this is a custom prop and not plain PreReleaseVersionIteration

akoeplinger: ah ok I thought the "set-to-empty-value" might have special meaning

ViktorHofer:
What about sourcelink, symreader and xdt? They all now ship to nuget.org via the VMR.

mmitche:
> What about sourcelink, symreader and xdt? They all now ship to nuget.org via the VMR.

Missing from the net11 config. Fixing now.

src/fsharp/eng/Versions.props
mmitche: This should also be preview2.

PR #4479, opened 2026-01-28
Add BuildStep attribute to ProjectToBuild elements

ViktorHofer:
Manual backport of https://github.com/dotnet/dotnet/pull/4478

This makes sure that the RepoTasks get built first


akoeplinger (approved):

ViktorHofer:
This one can get force-merged as we don't validate the path in CI. I double checked that the msbuild syntax is correct.

Issue #4451, opened 2026-01-27
Create a CI job that builds composite repos with MSBuild's -mt mode


JanProvaznik:
Similar to https://github.com/dotnet/msbuild/pull/13124 we want an end-to-end correctness/reliability test that we can build dotnet in MSBuild's new -mt (MultiThreaded) mode which is actively developed across repositories (it requires migrations of MSBuild tasks). Currently we will need to exclude some projects from -mt build, as there are blockers tracked in MSBuild repo.

This will enable better validation of MSBuild task migration to the new mode.

Once the feature is stable the intent is to switch to using it by default for all repos, but right now that is not feasible.

1. identify which repos are not buildable
2. figure out how to inject -mt parameter for msbuild in repo builds
3. create the CI job with exclusions

baronfel:
@JanProvaznik as you're disabling specific repos on the VMR job, can you make sure that we have issues for the root cause for each disablement linked to this issue?

PR #4487, opened 2026-01-28
Add a pipeline for MSBuild -mt validation

JanProvaznik:
#4451

## What

Adds infrastructure for validating MSBuild's [multithreaded (-mt) mode](https://github.com/dotnet/msbuild/blob/main/documentation/specs/multithreading/multithreaded-msbuild.md) by building the VMR with it.

## Approach

Following [the same pattern as runtime](https://github.com/dotnet/runtime/blob/main/eng/pipelines/vmr-build-pr.yml), the pipeline lives in the **msbuild repo** (`src/msbuild/azure-pipelines/vmr-mt-validation.yml`) and extends the existing VMR build templates (`vmr-build.yml`).

The key design choice is that **-mt mode is only applied to stage 2** (which builds with the source-built MSBuild from stage 1). Stage 1 uses the LKG SDK whose MSBuild has known -mt bugs.

### VMR template changes
- **`vmr-build.yml`**: Add `extraPropertiesStage2` parameter (threaded to source-build-stages → source-build-and-validate)
- **`source-build-stages.yml`**: Thread `extraPropertiesStage2` to `source-build-and-validate.yml`
- **`source-build-and-validate.yml`**: Apply `extraPropertiesStage2` only to legs with `reuseBuildArtifactsFrom` (i.e., stage 2 legs)

### Exclusion logic
- **`repo-projects/Directory.Build.props`**: Add `MSBuildMTExcludedReposOverride` MSBuild property override (in addition to existing `MSBUILD_MT_EXCLUDED_REPOS` env var). Uses `%3B` encoding on the command line since MSBuild treats semicolons as property separators.
- Default exclusion list: `arcade;runtime;roslyn;aspnetcore;sdk;efcore;winforms;wpf;razor;source-build-reference-packages`

### MSBuild repo pipeline (new)
- **`src/msbuild/azure-pipelines/vmr-mt-validation.yml`** (~42 lines): Manually triggered pipeline with configurable `excludedRepos` parameter. Uses the same `vmr-build.yml@vmr` pattern as runtime.

### Removed
- **`eng/pipelines/msbuild-mt-2stage-validation.yml`**: Replaced by the simpler msbuild-repo pipeline above (was 227 lines of custom YAML duplicating VMR template logic).

## Pipeline parameters
- **`excludedRepos`**: Repos to exclude from -mt mode (`%3B`-separated). Configurable per-run to test MSBuild fixes against previously-broken repos.

## Usage
1. After merge, create pipeline in Azure DevOps pointing at `dotnet/msbuild` repo, file `azure-pipelines/vmr-mt-validation.yml`
2. Queue manually, optionally adjusting `excludedRepos` to test fixes
3. Expected volume: 0-5 runs per day during development

Once MSBuild team is confident in -mt correctness and performance, this mode would become the default for regular CI.

### DIFF
--- a/src/msbuild/azure-pipelines/vmr-sb-validation.yml
+++ b/src/msbuild/azure-pipelines/vmr-sb-validation.yml
@@ -1,7 +1,7 @@
-# MSBuild -mt (multithreaded) mode VMR validation pipeline.
+# MSBuild VMR source-build validation pipeline.
 #
-# Validates that MSBuild's -mt mode works correctly by building the VMR
-# with the current PR's changes and -mt enabled.
+# Builds the VMR in source-build mode to validate MSBuild changes.
+# Supports passing extra MSBuild properties to stage 2 builds (e.g., -mt mode).
 #
 # Trigger: Manual only (queue from Azure DevOps UI)
 # Uses the same pattern as runtime's vmr-build-pr.yml.
@@ -10,10 +10,10 @@ trigger: none
 pr: none
 
 parameters:
-- name: excludedRepos
-  displayName: 'Repos to exclude from -mt mode (use %3B as separator, decoded to ; by MSBuild)'
+- name: extraPropertiesStage2
+  displayName: 'Extra MSBuild properties for stage 2 (e.g., /p:DotNetBuildMT=true)'
   type: string
-  default: 'arcade%3Bruntime%3Broslyn%3Baspnetcore%3Bsdk%3Befcore%3Bwinforms%3Bwpf%3Brazor%3Bsource-build-reference-packages'
+  default: '/p:DotNetBuildMT=true /p:MSBuildMTExcludedReposOverride=arcade%3Bruntime%3Broslyn%3Baspnetcore%3Bsdk%3Befcore%3Bwinforms%3Bwpf%3Brazor%3Bsource-build-reference-packages'
 
 variables:
 - template: /eng/common/templates/variables/pool-providers.yml@self
@@ -37,6 +37,4 @@ stages:
     isBuiltFromVmr: false
     scope: lite
     verifications: [ "source-build-stage1", "source-build-stage2" ]
-    extraPropertiesStage2: >-
-      /p:DotNetBuildMT=true
-      /p:MSBuildMTExcludedReposOverride=${{ parameters.excludedRepos }}
+    extraPropertiesStage2: ${{ parameters.extraPropertiesStage2 }}


eng/common/tools.ps1
```diff
@@ -820,6 +820,11 @@ function MSBuild-Core() {
 
   $cmdArgs = "$($buildTool.Command) /m /nologo /clp:Summary /v:$verbosity /nr:$nodeReuse /p:ContinuousIntegrationBuild=$ci"
 
+  # Add -mt flag for MSBuild multithreaded mode if enabled via environment variable
+  if ($env:MSBUILD_MT_ENABLED -eq "1") {
+    $cmdArgs += ' -mt'
+  }
+
```
ViktorHofer: You will also need to do the same change in src/arcade/eng/common/tools.ps1 and sh. This file here is used by the bootstrapping repos (VMR orchestrator + SBRP). The one in Arcade is used by all other repos.

eng/common/tools.ps1
```diff
@@ -820,6 +820,11 @@ function MSBuild-Core() {
 
   $cmdArgs = "$($buildTool.Command) /m /nologo /clp:Summary /v:$verbosity /nr:$nodeReuse /p:ContinuousIntegrationBuild=$ci"
 
+  # Add -mt flag for MSBuild multithreaded mode if enabled via environment variable
+  if ($env:MSBUILD_MT_ENABLED -eq "1") {
+    $cmdArgs += ' -mt'
```
ViktorHofer: Should msbuild support turning -mt on via an env var in general?

eng/common/tools.ps1
JanProvaznik: hah there is an chicken and egg problem. released MSBuild can't build arcade or itself in -mt so we first need the second stage bootstrapped MSBuild from this branch before we can use MT

JanProvaznik: hm actually that's weird, the failing arcade build is with 18.4 msbuild but locally it works 💢

JanProvaznik: :( a real -mt bug https://github.com/dotnet/msbuild/issues/13140

JanProvaznik:
~~note: confirmed we don't want to merge this, this validation should run as part of msbuild ci/on demand, not here (but it's easier development)~~

ViktorHofer:
@JanProvaznik every repo builds with the LKG msbuild in stage1 builds. Only source-build's stage2 builds with the live msbuild. But if we get the msbuild fix merged into main, we can just easily do another re-bootstrap in the VMR.

> note: confirmed we don't want to merge this, this validation should run as part of msbuild ci/on demand, not here (but it's easier development)

Clarification: We want to merge this PR but have the job not run by default. Only via the VMR pipeline that we will add to msbuild and sdk.

ViktorHofer:
FWIW we usually don't put business logic (the filtering) in YML so that the local build behaves the same as the YML build. This is important as our source-build partners don't use our YML to build the VMR. Not important here as source-build partners likely don't want to build the VMR with the -mt mode but I could see myself wanting to build the VMR with it.

JanProvaznik:
@ViktorHofer is there a possible issue using the same config as the sourcebuild jobs? I know we want to avoid windows due to the pool being small, but I'm not sure if I'm overusing some other resources with sourcebuild config.

(Extracting to separate pipeline with separate trigger is pending once I get this working with the 2 stage build).

eng/pipelines/msbuild-mt-2stage-validation.yml
ViktorHofer, line 1: Are you sure all this is needed? We use this differently in repos like runtime: https://dev.azure.com/dnceng-public/public/_build?definitionId=331

https://github.com/dotnet/runtime/blob/main/eng/pipelines/vmr-build-pr.yml -> We extend the vmr yml there.

ViktorHofer:
@dotnet/source-build can you please review the YML changes around `extraPropertiesStage2`? Thanks

eng/pipelines/templates/stages/source-build-and-validate.yml
```diff
@@ -38,6 +38,11 @@ parameters:
   type: string
   default: ''
 
+# Extra MSBuild properties to pass only to source-build stage 2 legs (those that reuse stage 1 artifacts).
+- name: extraPropertiesStage2
```
ellahathaway, line 42: I think a boolean parameter would be more appropriate here, something like `enableMT`. This fits the established pattern used by other source-build options like `enablePoison`, `runOnline`, `withPreviousSDK`, etc.

JanProvaznik: I disagree,
we want 2 things from the pipeline:
- enable MT for stage 2,
- control which repos are excluded as a parameter

That would help us with answering for MSBuild/sdk correctness questions "did this change fix this exclusion when we have this opt in feature enabled"?

Being able to pass properties gives us the power to do that, whereas a single flag would not.
This is a powerful customization point in the pipeline template, but is that a problem?

Alternative design:
the pipeline lives in VMR instead of MSBuild and it'd sync the msbuild changes and the exclusion would be configurable in project files on VMR side because the pipeline would be launched on a writable custom VMR branch.

Let me know if I convinced you or you'd be more comfortable with alternative or neither.

ViktorHofer (approved):

src/msbuild/azure-pipelines/vmr-sb-validation.yml
```diff
@@ -0,0 +1,42 @@
...
+
+resources:
+  repositories:
+  - repository: vmr
+    type: github
+    name: dotnet/dotnet
+    endpoint: dotnet
+
+stages:
+- template: /eng/pipelines/templates/stages/vmr-build.yml@vmr
```
mthalman, line 35: Is this path correct? There will be two repos checked out (msbuild and dotnet). Doesn't it need to indicate which one?

mthalman (commented):
Is there a test run available that we can see how this pipeline works?

Is this pipeline intended to be temporary until `mt` validation is completed?

src/msbuild/azure-pipelines/vmr-sb-validation.yml
JanProvaznik: that's the `@vmr` at the end see: https://github.com/dotnet/runtime/blob/main/eng/pipelines/vmr-build-pr.yml

the pipeline hasn't run because it needs to be created from msbuild repo where it currently is not present (to be backflowed)... and I can't put it there because it relies on the changed VMR...

I was tinkering with the multistage config mechanism and running it as part of VMR CI, but that's way too much volume to merge. It seems that I accidentally forcepushed so I lost the history of how that went :(

Yes it'll be deleted in <a year. But if this proves to be a good approach I expect to want to do more of this kind of 2-stage build validation for WIP/opt-in features and PRs (for MSBuild & sdk at least)

ViktorHofer: Actually that's a good point. I would name this yml MT agnostic as we want to use it regardless of mt in the future.

Renaming a pipeline (and its YML) later has proven inconvenient.

mthalman: > the pipeline hasn't run because it needs to be created from msbuild repo where it currently is not present (to be backflowed)... and I can't put it there because it relies on the changed VMR...

You can push your changes into the Azure repo in a dev branch and create a new pipeline that references that branch and run from there to test.

JanProvaznik: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1298417&view=results

JanProvaznik: @mthalman

JanProvaznik: ah that review seems to be optional, we're confident so merging soon if everything passes and no more blocking comments :)

Issue #4182, opened 2026-01-14
[Operational Issue]: dotnet-unified-build fails on OSX in public


akoeplinger:
### Error

only happens on main (scheduled) builds presumably because we don't do dryrun signing on PR builds

```
    SignToolTask starting.
    DryRun: True
    Signing mode: Real
    MicroBuild signing logs will be in (Signing*.binlog): /Users/runner/work/1/s/src/runtime/artifacts/log/Release/
    MicroBuild signing configuration will be in (Round*.proj): /Users/runner/work/1/s/src/runtime/artifacts/tmp/Release/
##[error]artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(76,5): error : Unknown file format.
   at System.Reflection.PortableExecutable.PEHeaders..ctor(Stream peStream, Int32 size, Boolean isLoadedImage)
   at System.Reflection.PortableExecutable.PEReader.InitializePEHeaders()
   at Microsoft.DotNet.SignTool.VerifySignatures.IsSignedPE(String filePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/VerifySignatures.cs:line 192
   at Microsoft.DotNet.SignTool.Configuration.ExtractSignInfo(PathWithHash file, PathWithHash parentContainer, String collisionPriorityId, String wixContentFilePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 389
   at Microsoft.DotNet.SignTool.Configuration.TrackFile(PathWithHash file, PathWithHash parentContainer, String collisionPriorityId) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 217
   at Microsoft.DotNet.SignTool.Configuration.TryBuildZipData(FileSignInfo zipFileSignInfo, ZipData& zipData, String alternativeArchivePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 871
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(76,5): error : Unknown file format.
[...]

PR #4536, opened 2026-01-29
Remove R2R disable from LanguageServer.csproj

jkoritzinsky:
Testing out if we've since fixed the crossgen2 crash as part of other work.


akoeplinger:
The failure on OSX is https://github.com/dotnet/dotnet/issues/4182 and looks unrelated

PR #4561, opened 2026-01-30
[release/11.0.1xx-preview1] Set the branding argument directly from the brandingType

mmitche:
Using the variable in the pre-processor doesn't work.


ViktorHofer (commented):
Problem is that if we now set this unconditionally, it would always overwrite the Versions.props setting which would be wrong.

mmitche:
Hmm yeah. Alright well we'll need some more surgery then

mmitche:
Closing in favor of #4601

PR #4564, opened 2026-01-30
[release/10.0.2xx] Align VMR global branding with SDK style

mmitche:
This allows the branding tooling to easily handle branding in the root of the VMR.


jkoritzinsky (approved):

mmitche:
We need to pad the patch out. This separates this out into a different property for clarity.

PR #4634, opened 2026-02-03
Remove unused asset root url properties

mmitche:


akoeplinger:
Also here: https://github.com/dotnet/workload-versions/blob/5acda5b959c5b2283d22ea66e97526bb96f6ee5d/eng/Versions.props#L26-L31

PR #4657, opened 2026-02-04
Use N-1 SB version for source generators only

mthalman:
Fixes dotnet/source-build#5355

Removes the temporary workaround that changed Roslyn to always use N-1 artifacts in source-only builds. Replaces it with a more targeted solution that only uses N-1 artifacts for analyzers/source generators that get executed during the same build in which they were produced (in which case N-1 artifacts are the only option).


jkoritzinsky:
Cc @sbomer this might be a good path for the runtime work to go

sbomer:
I think this approach won't work for https://github.com/dotnet/runtime/issues/123559 because it looks like the "previous" versions are only available for DotNetBuildSourceOnly. We need to ensure a consistent shipping version for the official product build too.

mthalman:
@dotnet/source-build - Please review

PR #4689, opened 2026-02-05
Run validation script job on linux

akoeplinger:


adamzip:
I guess it's trying to build the .net sdk from the VMR itself and use it to run the changevalidation tool, but it's failing because it's missing a .net version needed to run the vmr build in the first place?

but if I'm not mistaken, this project should only need whatever .net version arcade-services needs (via the DarcLib dependency), so maybe it can be changed to just use .net 10 or something?

akoeplinger:
@adamzip it was just me messing up the arguments to InitializeDotnetCli when translating from powershell to bash, so it wasn't installing dotnet.

I switched to using dotnet.sh since that handles installing+running in one go, but it looks like for some reason it was missing execute permissions (but the failure in the Run Validation Script job shows that it works).

PR #4709, opened 2026-02-06
Toolset update: WiX v6

joeloff:
The PR includes updates to the WiX v6 toolset

* Suppressing HEAT5149 - Heat is being deprecated in v7, but the toolset actively surfacing warnings results in errors. 
* Authoring changes
    * `RelatedBundle@Id` is now `RelatedBundle@Code`

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix5/wix.targets
+++ b/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix5/wix.targets
@@ -298,7 +298,7 @@
       SuppressRootDirectory="true"
       DirectoryRefId="%(DirectoryToHarvest.DirectoryRef)"
       OutputFile="%(DirectoryToHarvest.WixSourceFile)"
-      AdditionalOptions="-sw5149" />
+      AdditionalOptions="$(AdditionalHeatArgs) -sw5149" />
 
     <!--
       Currently FileElementToStabilize assumes a single DirectoryToHarvest. If there were multiple,


src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix5/wix.targets
```diff
@@ -53,7 +53,7 @@
 
     <PropertyGroup>
       <WixExeDir>$(PkgMicrosoft_Wix)\tools\net6.0\any\</WixExeDir>
-      <WixExtensionsDir>wixext5</WixExtensionsDir>
```
wtgodbe, line 56: Need to change the directory name(s) to wix6?

src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix5/wix.targets
```diff
@@ -298,7 +298,7 @@
       SuppressRootDirectory="true"
       DirectoryRefId="%(DirectoryToHarvest.DirectoryRef)"
       OutputFile="%(DirectoryToHarvest.WixSourceFile)"
-      AdditionalOptions="$(AdditionalHeatArgs)" />
+      AdditionalOptions="-sw5149" />
```
NikolaMilosavljevic: Should we still support `$(AdditionalHeatArgs)` in addition to hardcoded `-sw5149`? I'm aware that it's currently unused.

joeloff: Yeah, I wanted to ask about that. I wanted to make sure we always set this. Perhaps we should do something like

```xml
<AdditionalHeatArgs Condition="!$(AdditionalHeatArgs.Contains('-sw5149'))">$(AdditionalHeatArgs) -sw5149</AdditionalHeatArgs>
<AdditionalHeatArgs>$(AdditionalHeatArgs.Trim())</AdditionalHeatArgs>
```
And we do it before executing the `HeatDirectory` task. That way, if it's already set, we don't set it again and if not, we always add it.

NikolaMilosavljevic: Yeah, something like that could work. Alternatively a simple append of `-sw5149` could work if wix doesn't care about duplicate arguments.

joeloff: @NikolaMilosavljevic I'm busy verifying the duplicate suppression to make sure it runs.

joeloff: okay, it's happy with multiple suppressions. Gonna take the copilot suggetion

PR #4720, opened 2026-02-07
Enable osx-x64 in PR validation

ViktorHofer:
Diagnosing "Run Tests" (scenario-tests) failures only in the official build later is really annoying. Would be good to guard full stack osx build in PR.


akoeplinger:
We build on x64 Macs in public so you can't run arm64 bits in scenario-tests there

akoeplinger:
I like it in theory but it will put even more stress on the hosted pool...

akoeplinger (approved):

ViktorHofer:
Agreed that this isn't helping with our limited capacity but we can always revert if necessary.

NikolaMilosavljevic (approved):

ViktorHofer:
```
    DryRun: True
    Signing mode: Real
    MicroBuild signing logs will be in (Signing*.binlog): /Users/runner/work/1/s/src/runtime/artifacts/log/Release/
    MicroBuild signing configuration will be in (Round*.proj): /Users/runner/work/1/s/src/runtime/artifacts/tmp/Release/
##[error]artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error : Unknown file format.
   at System.Reflection.PortableExecutable.PEHeaders..ctor(Stream peStream, Int32 size, Boolean isLoadedImage)
   at System.Reflection.PortableExecutable.PEReader.InitializePEHeaders()
   at Microsoft.DotNet.SignTool.VerifySignatures.IsSignedPE(String filePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/VerifySignatures.cs:line 192
   at Microsoft.DotNet.SignTool.Configuration.ExtractSignInfo(PathWithHash file, PathWithHash parentContainer, String collisionPriorityId, String wixContentFilePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 467
   at Microsoft.DotNet.SignTool.Configuration.TrackFile(PathWithHash file, PathWithHash parentContainer, String collisionPriorityId) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 217
   at Microsoft.DotNet.SignTool.Configuration.TryBuildZipData(FileSignInfo zipFileSignInfo, ZipData& zipData, String alternativeArchivePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 869
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error : Unknown file format.
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error :    at System.Reflection.PortableExecutable.PEHeaders..ctor(Stream peStream, Int32 size, Boolean isLoadedImage)
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error :    at System.Reflection.PortableExecutable.PEReader.InitializePEHeaders()
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error :    at Microsoft.DotNet.SignTool.VerifySignatures.IsSignedPE(String filePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/VerifySignatures.cs:line 192
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error :    at Microsoft.DotNet.SignTool.Configuration.ExtractSignInfo(PathWithHash file, PathWithHash parentContainer, String collisionPriorityId, String wixContentFilePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 467
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error :    at Microsoft.DotNet.SignTool.Configuration.TrackFile(PathWithHash file, PathWithHash parentContainer, String collisionPriorityId) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 217
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error :    at Microsoft.DotNet.SignTool.Configuration.TryBuildZipData(FileSignInfo zipFileSignInfo, ZipData& zipData, String alternativeArchivePath) in /_/src/arcade/src/Microsoft.DotNet.SignTool/src/Configuration.cs:line 869
##[error]artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error : Failed to build zip data for /Users/runner/work/1/s/src/runtime/artifacts/packages/Release/Shipping/dotnet-runtime-11.0.0-ci-osx-x64.tar.gz
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.Arcade.Sdk/tools/Sign.proj(79,5): error : Failed to build zip data for /Users/runner/work/1/s/src/runtime/artifacts/packages/Release/Shipping/dotnet-runtime-11.0.0-ci-osx-x64.tar.gz
    SignToolTask execution finished.
```

@mmitche these failures are interesting. Tells me:
- We probably are doing REAL signing even though dryrun=true is passed in
- `Unknown file format` for `dotnet-runtime-11.0.0-ci-osx-x64.tar.gz`

mmitche (approved):

ViktorHofer:
@ellahathaway @mmitche can you please help me diagnose if the added osx-x64 job tries to real sign here in public?

ellahathaway:
> @ellahathaway @mmitche can you please help me diagnose if the added osx-x64 job tries to real sign here in public?

The job is dry run signing, so it is not attempting to real sign.

ViktorHofer:
OK then that probably means that dry-run signing doesn't correctly handle symlinks, hardlinks or pkg files? cc @MichaelSimons

I see the same failures in the public scheduled (daily) build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1297350&view=logs&j=3ac9c5cd-f25e-53e7-1c9b-f058b1dbaa18&t=05376304-94d7-5e9d-7761-01060cc44912

ellahathaway:
> OK then that probably means that dry-run signing doesn't correctly handle symlinks, hardlinks or pkg files? cc @MichaelSimons
> 
> I see the same failures in the public scheduled (daily) build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1297350&view=logs&j=3ac9c5cd-f25e-53e7-1c9b-f058b1dbaa18&t=05376304-94d7-5e9d-7761-01060cc44912

Can we have the OSX job upload the runtime artifacts even when signing in that leg fails? I pulled the runtime artifacts from an official build that published the runtime artifacts, and they pass dry-run signing ([link to pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2906450&view=logs&j=595e44c8-424f-5de6-3262-bdb644f56258&t=3b82af4d-68d5-5e79-87f8-cb20f56fd4c7)). I’m wondering if the PR build is producing a different artifact structure/layout, and having the failing leg’s runtime artifacts would help confirm.

ViktorHofer:
That's a great idea but a bit hard to implement with our current infrastructure. We would need to make Publish.proj (Arcade) run even when Sign.proj fails.

EDIT: Let's see if that helps.

ellahathaway:
I was able to repro the error locally with the `dotnet-runtime-11.0.0-ci-osx-x64.tar.gz` artifact. When I run signing directly against the tarball, the signing step fails with `Unknown file format`. However, if I first extract the tarball to a directory and then run SignTool over the extracted folder contents, signing completes without hitting that error. This difference makes me think the failure is likely due to SignTool not correctly handling symlinks (or other non-regular entries) inside the tar. @MichaelSimons

MichaelSimons:
> I was able to repro the error locally with the `dotnet-runtime-11.0.0-ci-osx-x64.tar.gz` artifact. When I run signing directly against the tarball, the signing step fails with `Unknown file format`. However, if I first extract the tarball to a directory and then run SignTool over the extracted folder contents, signing completes without hitting that error. This difference makes me think the failure is likely due to SignTool not correctly handling symlinks (or other non-regular entries) inside the tar. @MichaelSimons

I discovered what I appears to be the same issue and fixed it in https://github.com/dotnet/arcade/pull/16536.

ViktorHofer:
Looks like dry run signing the .tar.gz file on osx is still broken. @ellahathaway / @MichaelSimons any idea what the next step should be here?

ellahathaway:
> Looks like dry run signing the .tar.gz file on osx is still broken. @ellahathaway / @MichaelSimons any idea what the next step should be here?

Is it possible that the artifact is structured incorrectly? We're not seeing this issue in official builds and Michael fixed the suspected SignTool issue.

ViktorHofer:
Do the default official builds dry run sign? I thought they would maybe test sign.

> Is it possible that the artifact is structured incorrectly? We're not seeing this issue in official builds and Michael fixed the suspected SignTool issue.

Hmm, not sure...

ellahathaway:
> Do the default official builds dry run sign? I thought they would maybe test sign.

I think default official builds on non-release branches don't sign at all, not even dry run signing. I would expect to see this error in any official build that real signs or dry-run signs though (test signing is not supported on non-windows)

mmitche:
> > Do the default official builds dry run sign? I thought they would maybe test sign.
> 
> I think default official builds on non-release branches don't sign at all, not even dry run signing. I would expect to see this error in any official build that real signs or dry-run signs though (test signing is not supported on non-windows)

We haven't seen this in the p2 build, so maybe something else is going on.

ViktorHofer:
This is only showing up in public... We could diff the .tar.gz produced here with one produced internally from the same SHA.

mmitche:
Looking a bit closer here:

```
Extracting file 'Microsoft.CSharp.dll' from '/Users/runner/work/1/s/src/runtime/artifacts/packages/Release/Shipping/dotnet-runtime-11.0.0-ci-osx-x64.tar.gz' to '/Users/runner/work/1/s/src/runtime/artifacts/tmp/Release/ContainerSigning/911/./shared/Microsoft.NETCore.App/11.0.0-ci/GNUSparseFile.0/Microsoft.CSharp.dll'. (Caching with perms: 644)
    Tracking file '/Users/runner/work/1/s/src/runtime/artifacts/tmp/Release/ContainerSigning/911/./shared/Microsoft.NETCore.App/11.0.0-ci/GNUSparseFile.0/Microsoft.CSharp.dll' isNested=True
```

```
1. GNU tar “sparse file” extension
In GNU tar format, sparse files (files with large zero-filled regions) are represented using special metadata records. Historically:

GNU tar stores sparse metadata using pseudo‑entries
These entries are often named internally like:
GNUSparseFile.<n>


They are not meant to appear as real directories or path segments when extracting

In a valid flow:

The sparse metadata is read
The real file is reconstructed
GNUSparseFile.* never becomes part of the output path
```

mmitche:
The "large sections of zero bytes" is pretty interesting there. Hypothesis would be that the `Microsoft.CSharp.dll` has some empty sections (maybe PE sig? Strong name sig?) that are empty in the dry run signed build.

Other possibility is that the macOS machines in internal CI vs. public have different tar implementations.

mmitche:
If this works, it's somewhat likely that there are bugs in .NET's tar implementation again.

src/arcade/src/Microsoft.DotNet.Build.Tasks.Archives/build/archives.targets
```diff
@@ -126,11 +126,11 @@
                   DestinationFile="$(_DestinationFileName)"
                   Condition="'$(_CurrentArchiveFormat)' == 'zip'"/>
     <!-- use parallel gzip implementation when available -->
-    <Exec Command="tar -cf - . | pigz &gt; &quot;$(_DestinationFileName)&quot;"
+    <Exec Command="tar --no-read-sparse -cf - . | pigz &gt; &quot;$(_DestinationFileName)&quot;"
```
akoeplinger, line 129: @mmitche the `tar` on my macOS 14 machine doesn't have the `--no-read-sparse` option, I guess it might also not be available in some Linux distros. But let's see if it indeed would fix the issue.

akoeplinger:
> Other possibility is that the macOS machines in internal CI vs. public have different tar implementations.

We're using macOS 15 both internally and externally, the only difference is public uses x64 while internal uses arm64 (not sure if there can be difference in the tar version)

ViktorHofer:
@mmitche your commits resolved the issue! Nice find. What are the next steps? Alex indicated that we can't merge this in given that not all tar installations have that command available.

akoeplinger:
I wonder if https://github.com/dotnet/dotnet/issues/5281 also plays into this.

mmitche:
We could use the workaround and scope to osx-only builds with some detected version, while we wait for resolution of  https://github.com/dotnet/runtime/issues/125281

mmitche:
> We could use the workaround and scope to osx-only builds with some detected version, while we wait for resolution of [dotnet/runtime#125281](https://github.com/dotnet/runtime/issues/125281)

ViktorHofer:
OK this is still an issue. The OSX legs are failing when triggering the osx job in public via azp run dotnet-unified-build-full. @mmitche do you remember what the cause was exactly? Would be good file a BCL issue in the Tar implementation if necessary.

ViktorHofer:
Closing. I'm not interested in adding the osx-x64 job to the PR validation when it takes over 4 hours. Until three days ago, OSX ran just fine publicly now in the scheduled public builds. So the previously discussed issue is already resolved.

PR #4723, opened 2026-02-07
[TestOnly] Test dotnet_ipc_created headers

mdh1418:
Validating whether the unified build will have the headers for this logic in .NET 10 per https://github.com/dotnet/runtime/pull/124058#issuecomment-3865051157


mdh1418:
I was able to validate against the 10.0.4 nuget packages for NativeAOT

PR #4740, opened 2026-02-09
Source tarball comparison validation

mthalman:
Fixes dotnet/source-build#5472

Adds validation which compares the contents of the VMR repo with the output of the `git archive` command to ensure they have the expected contents.

Refactors the test code to move the baseline/exclusion infrastructure from SourceBuilt.Tests to TestUtilities.

The new test is added to Microsoft.DotNet.Tests project which can then use the baseline/exclusion infrastructure from TestUtilities. This test gathers the list of files from the git repo using the `git ls-tree` command so there's no danger of including other non-repo files like build artifacts. This file list is compared to the list of files that get included in the output of the `git archive` command.


mthalman:
cc @omajid

test/Microsoft.DotNet.Tests/SourceTarballContentTests.cs
```diff
@@ -0,0 +1,120 @@
...
+        string actualContent = string.Join(Environment.NewLine, filteredMissing);
+        string? message = BaselineHelper.CompareBaselineContents(BaselineFileName, actualContent, OutputHelper, Config.LogsDirectory, BaselineSubDir);
+        Assert.True(message == null, message);
+    }
+
+    /// <summary>
+    /// Runs git archive and reads the tar stream directly from stdout to enumerate file entries
+    /// without writing the full archive to disk.
+    /// </summary>
+    private HashSet<string> GetGitArchiveFileList(string repoRoot)
```
ellahathaway: Could you switch this over to using the [`CreateSourceArtifact`](https://github.com/dotnet/dotnet/blob/main/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifact.cs) task instead?

Similarly, maybe this test could be run in-build on [the leg that produces the source artifact](https://github.com/dotnet/dotnet/blob/main/eng/pipelines/templates/stages/source-build-stages.yml#L206)? I realize this is more of a content validation than a typical build check, so wiring it in might be a bit involved—but it could end up being a more cohesive fit since it would validate the artifact at the point it’s created.

test/Microsoft.DotNet.Tests/SourceTarballContentTests.cs
```diff
@@ -0,0 +1,120 @@
...
+        StringBuilder stderr = new();
+        process.ErrorDataReceived += (_, e) => { if (e.Data is not null) stderr.AppendLine(e.Data); };
+        process.BeginErrorReadLine();
+
+        HashSet<string> files = TarHelper.GetEntryNames(process.StandardOutput.BaseStream)
+            .Where(name => !name.EndsWith('/'))
+            .ToHashSet();
+
+        process.WaitForExit(600000);
+
```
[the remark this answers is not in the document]
mthalman: no longer relevant

test/Microsoft.DotNet.Tests/SourceTarballContentTests.cs
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/4740/commits/fbc743363fb57f45d64dd5988b822a5ce01102f1

ellahathaway (approved):

PR #4758, opened 2026-02-10
[release/11.0.1xx-preview1] Source code updates from dotnet/fsharp

dkurepa:
Created using `vmr reset --build 301012 fsharp`


src/fsharp/eng/Versions.props
```diff
@@ -14,7 +14,7 @@
   <PropertyGroup>
     <!-- Don't use the built in support for pre-release iteration. The nuget repack task doesn't support
          the iteration format at the moment. https://github.com/dotnet/arcade/issues/15919 -->
-    <FSharpPreReleaseIteration>1</FSharpPreReleaseIteration>
+    <FSharpPreReleaseIteration>2</FSharpPreReleaseIteration>
```
dkurepa, line 17: this doesn't look correct. Maybe this subscription isn't configured correctly?

dkurepa: the same channel flowing to main is flowing here

dkurepa:
not needed as we already have a build for p1

PR #4768, opened 2026-02-10
Update Version_10_WithAliasSupport to preview.0.26110

nkolev92:
In theory this should replace https://github.com/dotnet/dotnet/pull/4555. 

Effectively, we are opting in the new .NET SDK into v4 lock files. 

I'm not really sure what this does to validation given that local build numbers are not always matching the real built ones. 

cc @MichaelSimons 

I want to use this PR to quickly test and address the problem in this repo before making the chnage NuGet side.


nkolev92:
This was just tests. CLosing

PR #4775, opened 2026-02-10
[release/10.0.3xx] Update to 10.0.103 released version

mmitche:


src/source-build-reference-packages/eng/Version.Details.props
mthalman, line 1: I don't think we want SBRP changes in this PR. There's only one 10.0 branch for SBRP: `release/10.0`. That should flow to all the feature bands for 10.0 in the VMR. Changing it here would just mean it would get stomped by that flow.

mmitche: Noted. Thanks.

mmitche: Does this also apply for flow from 1xx->2xx?

mmitche: e.g. https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet/pullrequest/57801?_a=files&path=/src/source-build-reference-packages/eng/Version.Details.props

mthalman: Yes, I guess it would apply to that too.

marcpopMSFT: want to just use my PR as it doesn't have SBRP or the nuget changes (which were only because the release wasn't done yet): https://github.com/dotnet/dotnet/pull/4829

PR #4790, opened 2026-02-11
Use net11.0 container images in vmr-build.yml

akoeplinger:
Looks like this got missed when doing the net11.0 bump


ViktorHofer:
For my own education, why is this needed? I assume those are images with pre-installed .NET 11? Do we have images without pre-installed .NET as well?

akoeplinger:
@ViktorHofer they don't have .net preinstalled but rather all the build dependencies and those can/do differ between branches

PR #4794, opened 2026-02-11
Remove NetSdkSupportingFeed from repo-projects

akoeplinger:
It's no longer needed.

### DIFF
--- a/test/Microsoft.DotNet.Tests/PackageSourceMappingsTests.cs
+++ b/test/Microsoft.DotNet.Tests/PackageSourceMappingsTests.cs
@@ -35,6 +35,7 @@ namespace Microsoft.DotNet.Tests
         [Fact]
         public void BuildWithMappingsNoLocalSources()
         {
+            string[] sources = [];
             RunTest("ub-mappings-nolocal.config", true, sources);
         }
 


repo-projects/Directory.Build.targets
```diff
@@ -281,7 +257,6 @@
       PrebuiltSourceName="$(PrebuiltNuGetSourceName)"
       SourceBuiltSourceNamePrefix="$(SourceBuiltSourceNamePrefix)"
       PreviousBuildPassSourceNamePrefix="$(PreviousBuildPassSourceNamePrefix)"
-      CustomSources="$(NetSdkSupportingFeedName)"
```
ViktorHofer, line 284: Looks like this one was a required parameter

akoeplinger: it wasn't actually, it just happened that NetSdkSupportingFeedName was the only feed in _BuildSources so SourceBuildSources parameter was empty

NikolaMilosavljevic:
Test assets should be updated as well, for example: https://github.com/dotnet/dotnet/blob/main/test/Microsoft.DotNet.Tests/assets/PackageSourceMappingsTests/original/ub-nomappings.config. There might be others.

I suggest running tests locally to update the `expected` content.

test/Microsoft.DotNet.Tests/PackageSourceMappingsTests.cs
```diff
@@ -36,8 +35,7 @@ public PackageSourceMappingsTests(ITestOutputHelper outputHelper)
         [Fact]
         public void BuildWithMappingsNoLocalSources()
         {
-            string[] sources = [NetSdkSupportingFeedName];
-            RunTest("ub-mappings-nolocal.config", true, sources, customSources: [NetSdkSupportingFeedName]);
+            RunTest("ub-mappings-nolocal.config", true, sources);
```
ViktorHofer, line 39: sources is now undefined here and in other places?

akoeplinger: meh, that's what you get when you fix things between meetings :D

ViktorHofer: Let Copilot fix it for you :)

ViktorHofer (approved):

PR #4820, opened 2026-02-12
[TEST] Add App Host Support for MSBuild

YuliiaKovalova:


src/msbuild/eng/Signing.props
```diff
@@ -12,6 +12,11 @@
 
     <FileSignInfo Include="RuntimeContracts.dll" CertificateName="3PartySHA2" />
 
+    <!-- MSBuild.exe is a native apphost that will be re-signed in the SDK layout.
+         Exclude from NuPkg signing to avoid AssemblyLoadsTracker cross-AppDomain
+         serialization issue in VS MSBuild (tracked separately). -->
+    <FileSignInfo Include="MSBuild.exe" CertificateName="None" />
```
ViktorHofer: Won't this cause the .NET Framework MSBuild.exe to also not get signed?

PR #4844, opened 2026-02-13
[release/10.0.3xx] Reset fsharp to 301592

dkurepa:
Fsharp changed which branch is flowing into the `release/10.0.3xx` branch, so we had to reset it in the VMR


premun:
If we decide this is what we want, we will need to fix the source-manifest because of the force-push I did for the target branch

PR #4847, opened 2026-02-13
[release/10.0.2xx] Reset Fsharp to 301592

dkurepa:


src/fsharp/eng/Versions.props
```diff
@@ -22,31 +22,23 @@
...
     <!-- -->
-
-
-    <!-- BEGIN - SPECIFIC OVERRIDES for 10.0.2xx. Those are overrides instead of updates to reduce risk of git conflicts in codeflow -->
-    <FSharpPreReleaseIteration></FSharpPreReleaseIteration>
-    <PreReleaseVersionLabel>servicing$(FSharpPreReleaseIteration)</PreReleaseVersionLabel>
-    <FSMajorVersion>10</FSMajorVersion>
-    <FSMinorVersion>0</FSMinorVersion>
-    <FSBuildVersion>200</FSBuildVersion>
-    <FSRevisionVersion>0</FSRevisionVersion>
```
T-Gro, line 33: @dkurepa , @premun : This is the place.

T-Gro:
@dkurepa , @premun :

Hi guys, is there a way we could be more in control if something like that happens again?
Unfortunately this "Reset" lead to losing versioning/branding information we had for 10.0.2xx and because of that, FSharp.Core out of that 10.0.2xx band got released to nuget.org as shipped (non preview) and branded as 11.0.xxx - because that is what we had in main at that time.


(cc @baronfel , this is the PR which @abonie  mentioned on Teams and it is why we had confused customers on reddit)

premun:
@T-Gro sorry about this. We will always wait for an approval of code owners next time

Issue #4895, opened 2026-02-16
vmr-sync clones the VMR SHA based on the repo's Version.Details.xml Sources SHA which results in torn state


ViktorHofer:
This build shows this in great detail: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1295880&view=results

The VMR gets synced via vmr-sync based on the repo's Version.Details.xml Sources SHA entry but the pipeline templates from the VMR get referenced from the default ref (SHA): https://github.com/dotnet/source-build-reference-packages/blob/cc90bb4ad53ff35dc109c2827428113e70a4681d/azure-pipelines/builds/vmr-build-pr.yml#L13-L18
-> no ref specified means default I guess.

This results in a torn state. The VMR and the pipelines must be from the same SHA. As a side-effect, if repositories have an older VMR backflow state (i.e. for msbuild which is intentional), the VMR cloned can be months old. Such a vmr validation pipeline wants to build with the live vmr usually though. That tells me that we could use the SHA from the resources.respositories section in the YML as the source-of-truth.

**Two intertangled action items here**:
1. Avoid the torn state by using the same SHA for both the resources.repositories section and the vmr-sync
2. Decide what should be the source-of-truth: the resources.repositories ref (defaults to main) or the repo's Version.Details Source tag SHA. I see reasons for either but whatever we do, we should think about repositories that have very old backflows, basically the MSBuild case from the link above.

cc @NikolaMilosavljevic @jkoritzinsky @JanProvaznik as this might impact your pipelines
cc @premun @mmitche

ViktorHofer:
The TL;DR why this is super bad:

When I make a change and pass in a new build argument from YML to the VMR's build entry point scripts, `i.e. build.sh --dummy` and add support for that in the build entry point scripts, all vmr validation pipelines would be broken.

premun:
We discussed this and since the YAML that will run the PR build comes from `main` but the code it will build is from whatever commit is in `V.D.xml`, this seems like it will break for non-main branches anyway.

So I am in favour of dropping this part:
https://github.com/dotnet/dotnet/blob/564380aa4a1fc71a14ca89f7be373342c4dd33eb/eng/common/templates/steps/vmr-sync.yml#L41-L60

And just using `main` and supporting `main` for these legs only.

We would not backport this to `release/10...` so if those branches still somehow build, we retain the most green setup.

ViktorHofer:
[...]

PR #4897, opened 2026-02-16
Remove vmr sha retrieval and checkout steps

ViktorHofer:
Removed scripts for obtaining vmr sha from Version.Details.xml for both Unix and Windows, and the checkout step for the VMR.

Use the commit from the ref specified in resources.repositories instead (defaults to refs/head/main).

One way to fix https://github.com/dotnet/dotnet/issues/4895
Test build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1296208&view=results

### DIFF
--- a/src/arcade/eng/common/vmr-sync.ps1
+++ b/src/arcade/eng/common/vmr-sync.ps1
@@ -103,8 +103,8 @@ Set-StrictMode -Version Latest
 Highlight 'Installing .NET, preparing the tooling..'
 . .\eng\common\tools.ps1
 $dotnetRoot = InitializeDotNetCli -install:$true
+$env:DOTNET_ROOT = $dotnetRoot
 $darc = Get-Darc
-$dotnet = "$dotnetRoot\dotnet.exe"
 
 Highlight "Starting the synchronization of VMR.."
 
@@ -152,7 +152,7 @@ else {
   & "$darc" $resetArgs
 
   if ($LASTEXITCODE -eq 0) {
-    Highlight "Successfully reset the VMR using `darc vmr reset`"
+    Highlight "Successfully reset the VMR using 'darc vmr reset'"
   }
   else {
     Fail "Synchronization of repo to VMR failed!"
--- a/src/arcade/eng/common/vmr-sync.sh
+++ b/src/arcade/eng/common/vmr-sync.sh
@@ -216,7 +216,7 @@ else
     --additional-remotes "$repo_name:$repo_root"
 
   if [[ $? == 0 ]]; then
-    highlight "Successfully reset the VMR using `darc vmr reset`"
+    highlight "Successfully reset the VMR using 'darc vmr reset'"
   else
     fail "Synchronization of repo to VMR failed!"
     fail "'$vmr_dir' is left in its last state (re-run of this script will reset it)."


ViktorHofer:
We would need to use `darc vmr reset` here which comes with its own kind of issues -> files in the VMR but not in msbuild due to recent changes, i.e.

ViktorHofer:
@premun suggested to update the vmr-sync scripts to do a `git reset --hard && darc vmr reset` in case `darc vmr forwardflow` fails.

src/arcade/eng/common/vmr-sync.ps1
```diff
@@ -109,6 +109,9 @@ $dotnet = "$dotnetRoot\dotnet.exe"
 Highlight "Starting the synchronization of VMR.."
 
 # Synchronize the VMR
+$repoRoot = Resolve-Path (Join-Path $PSScriptRoot '..\..') | Select-Object -ExpandProperty Path
+$repoName = Split-Path -Leaf $repoRoot
```
ViktorHofer: @premun the `darc vmr forwardflow` command didn't need the repo name and repo commit. It took the information from the repo in the cwd. Would it make sense to align these two commands?

If not, should repoName be passed in as a variable to these scripts instead?

src/arcade/eng/common/templates/vmr-build-pr.yml
```diff
@@ -34,6 +34,7 @@ resources:
     type: github
     name: dotnet/dotnet
     endpoint: dotnet
+    ref: refs/heads/main # Set to whatever vmr branch should be checked out
```
premun:
```suggestion
    ref: refs/heads/main # Set to whatever VMR branch the PR build should insert into
```

src/arcade/eng/common/vmr-sync.ps1
premun: Ah, we only really called `darc vmr reset` from within the VMR dir. But I guess your ask would make sense

src/arcade/eng/common/vmr-sync.ps1
```diff
@@ -130,9 +133,27 @@ if ($LASTEXITCODE -eq 0) {
...
-  Fail "'$vmrDir' is left in its last state (re-run of this script will reset it)."
-  Fail "Please inspect the logs which contain path to the failing patch file (use -debugOutput to get all the details)."
-  Fail "Once you make changes to the conflicting VMR patch, commit it locally and re-run this script."
-  exit 1
+  Highlight "Forwardflow failed, attempting recovery with 'darc vmr reset'.."
+  git -C $vmrDir reset --hard
+
+  $resetArgs = (
+    "vmr", "reset",
+    "${repoName}:HEAD",
```
premun, line 146: I think instead of `HEAD` we need to give `git rev-parse HEAD`

src/arcade/eng/common/vmr-sync.ps1
```diff
@@ -130,9 +133,27 @@ if ($LASTEXITCODE -eq 0) {
   Highlight "Synchronization succeeded"
 }
 else {
-  Fail "Synchronization of repo to VMR failed!"
-  Fail "'$vmrDir' is left in its last state (re-run of this script will reset it)."
-  Fail "Please inspect the logs which contain path to the failing patch file (use -debugOutput to get all the details)."
-  Fail "Once you make changes to the conflicting VMR patch, commit it locally and re-run this script."
-  exit 1
+  Highlight "Forwardflow failed, attempting recovery with 'darc vmr reset'.."
```
premun:
```suggestion
  Highlight "Failed to flow code into the local VMR. Falling back to resetting the VMR to match repo contents..."
```

src/arcade/eng/common/vmr-sync.ps1
```diff
@@ -130,9 +133,27 @@ if ($LASTEXITCODE -eq 0) {
...
+    "${repoName}:HEAD",
+    "--vmr", $vmrDir,
+    "--tmp", $tmpDir,
+    "--additional-remotes", "${repoName}:${repoRoot}"
+  )
+
+  & "$darc" $resetArgs
+
+  if ($LASTEXITCODE -eq 0) {
+    Highlight "Recovery with 'darc vmr reset' succeeded"
```
premun: If you don't know the command, this would have made it sound like it succeeded in a different way but instead it does something else.

```suggestion
    Highlight "Successfully reset the VMR using `darc vmr reset`"
```

src/arcade/eng/common/vmr-sync.sh
```diff
@@ -199,9 +202,22 @@ export DOTNET_ROOT="$dotnetDir"
 if [[ $? == 0 ]]; then
   highlight "Synchronization succeeded"
 else
-  fail "Synchronization of repo to VMR failed!"
-  fail "'$vmr_dir' is left in its last state (re-run of this script will reset it)."
-  fail "Please inspect the logs which contain path to the failing patch file (use --debug to get all the details)."
-  fail "Once you make changes to the conflicting VMR patch, commit it locally and re-run this script."
-  exit 1
+  highlight "Forwardflow failed, attempting recovery with 'darc vmr reset'.."
```
premun:
```suggestion
  highlight "Failed to flow code into the local VMR. Falling back to resetting the VMR to match repo contents..."
```

src/arcade/eng/common/vmr-sync.sh
```diff
@@ -199,9 +202,22 @@ export DOTNET_ROOT="$dotnetDir"
...
+  git -C "$vmr_dir" reset --hard
+
+  "$darc_tool" vmr reset \
+    "$repo_name:HEAD"                              \
+    --vmr "$vmr_dir"                               \
+    --tmp "$tmp_dir"                               \
+    --additional-remotes "$repo_name:$repo_root"
+
+  if [[ $? == 0 ]]; then
+    highlight "Recovery with 'darc vmr reset' succeeded"
```
premun:
```suggestion
    highlight "Successfully reset the VMR using `darc vmr reset`"
```

src/runtime/eng/pipelines/vmr-build-pr.yml
```diff
@@ -44,6 +44,7 @@ resources:
     type: github
     name: dotnet/dotnet
     endpoint: dotnet
+    ref: refs/heads/main # Set to whatever vmr branch should be checked out
```
premun:
```suggestion
    ref: refs/heads/main # Set to whatever VMR branch the PR build should insert into
```

src/source-build-reference-packages/azure-pipelines/builds/vmr-build-pr.yml
```diff
@@ -16,6 +16,7 @@ resources:
     type: github
     name: dotnet/dotnet
     endpoint: dotnet
+    ref: refs/heads/main # Set to whatever vmr branch should be checked out
```
premun:
```suggestion
    ref: refs/heads/main # Set to whatever VMR branch the PR build should insert into
```

src/arcade/eng/common/vmr-sync.ps1
ViktorHofer: Should I file a ticket in arcade-services? Are you ok with that current hacky way of finding the repo name and eventually remove that with better dard vmr reset support?

src/arcade/eng/common/vmr-sync.ps1
ViktorHofer: Hmm, HEAD worked locally for me

premun: Oh yeah, maybe `HEAD` might actually work because it will get resolved by `darc` just like it was a branch name. My bad

src/arcade/eng/common/vmr-sync.ps1
premun: Yeah, let's log an issue but instead of the current way (which will not work for some repos like nuget.client), could we maybe do what the removed code did?

```ps
[xml]$xml = Get-Content -Path $(Agent.BuildDirectory)/repo/eng/Version.Details.xml
$vmr_sha = $xml.SelectSingleNode("//Source").Sha
```

Instead of `Sha` we could grep `Mapping`.

premun (approved):

ViktorHofer:
Tested in https://dev.azure.com/dnceng-public/public/_build/results?buildId=1298417&view=logs&j=f8361ec4-8613-5f7f-2f10-d7dd7654e93f&t=6b2d8206-630e-54c9-322c-b59a8efc7b83. The sync now  works. This is ready to go.

src/arcade/eng/common/vmr-sync.ps1
ViktorHofer: Filed https://github.com/dotnet/arcade-services/issues/5973

PR #4928, opened 2026-02-18
Modify MSBuild debug output environment variable

ViktorHofer:
Updated MSBuild debug output path for better visibility.

### DIFF
--- a/repo-projects/Directory.Build.props
+++ b/repo-projects/Directory.Build.props
@@ -187,6 +187,9 @@
     <!-- Need to be passed in here so that repo builds don't restore into the orchestrator package cache (CI builds)
          or the user package cache (local dev builds). -->
     <EnvironmentVariables Include="NUGET_PACKAGES=$(RepoArtifactsPackageCache)" />
+
+    <!-- If MSBuild exits early, make sure debug output like 'MSBuild_*.failure.txt' ends up in a place we can see it. -->
+    <EnvironmentVariables Include="MSBUILDDEBUGPATH=$(ArtifactsLogRepoDir)" />
   </ItemGroup>
 
   <ItemGroup  Condition="'$(DotNetBuildSourceOnly)' == 'true'">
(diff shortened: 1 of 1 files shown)


repo-projects/Directory.Build.props
```diff
@@ -187,15 +187,16 @@
...
+
+    <!-- If MSBuild exits early, make sure debug output like 'MSBuild_*.failure.txt' ends up in a place we can see it. -->
+    <EnvironmentVariables Include="MSBUILDDEBUGPATH=$(ArtifactsLogRepoDir)" />
   </ItemGroup>
 
   <ItemGroup  Condition="'$(DotNetBuildSourceOnly)' == 'true'">
     <EnvironmentVariables Include="NuGetAudit=false" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(EnableExtraDebugging)' == 'true'">
```
ViktorHofer, line 199: @YuliiaKovalova please let me know if we should enable some of these by default as well. But think about perf in case some of those create extra work.

YuliiaKovalova (approved):

repo-projects/Directory.Build.props
YuliiaKovalova: it shouldn't cause any issues, since the logs reported only if crash happens.
The env var presence doesn't trigger any additional logging .

ViktorHofer: Sorry I should have been more explicit. I meant if any of those other env vars that are currently hidden under a feature flag should be on by default: https://github.com/dotnet/dotnet/blob/76e065cd2583952106b97ed29a13ed338df23470/repo-projects/Directory.Build.props#L200-L208

YuliiaKovalova: "MSBUILDDEBUGCOMM" can be useful, but it's expensive, so let's have MSBUILDDEBUGPATH only for now

PR #4932, opened 2026-02-18
Internal 2xx fixes

mthalman:
Port of the changes from https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet/pullrequest/58074

### Problem

Non-1xx feature band branches (2xx, 3xx, etc.) exclude runtime-dependent components from their builds. They only build SDK tooling and reuse the shared runtime from the 1xx band. Because the runtime isn't built, the bootstrap process (which downloads CentOS artifacts and bootstraps them to be portable) cannot work. Since the SDK being used is non-portable, it results in an attempt to get a non-portable CrossGen package which doesn't exist from the PSB artifacts (which are currently portable) -- see https://dev.azure.com/dnceng/internal/_build/results?buildId=2906286&view=logs&j=0c2a1884-6715-550d-a904-b6e3cc3a6e36&t=3cbdffe4-1ca5-5d9e-d978-72178c29357a. Instead, the pipeline needs to download non-portable, RID-specific previously source-built artifacts directly instead.

Additionally, when downloading archives from authenticated (internal) storage, curl -O derives the output filename from the full URL including the SAS token query string, resulting in a "File name too long" error (seen on AlmaLinux).

### Fix

1. Download RID-specific artifacts up front (vmr-build.yml): The "Download Source-Built Archives" step now downloads the non-portable, RID-specific Private.SourceBuilt.Artifacts archive into prereqs/packages/archive/. When prep-source-build.sh runs later, it finds the artifacts already present and skips both the centos download and the bootstrap. The prep step is also updated to pass --no-bootstrap --artifacts-rid explicitly.
1. Fix curl filename too long (download-source-built-archive.sh): DownloadWithRetries now accepts an optional output filename and uses curl -o <filename> instead of curl -O, preventing the SAS token query parameters from being included in the output filename.
1. Update leg names to reflect that they are using PSB SDK instead of Microsoft SDK in non-1xx builds.
1. Minor cleanup: Consolidated the SDK and shared-components download steps into a single script block with nested conditions, and fixed a YAML validation error where reuseBuildArtifactsFrom (an object/array parameter) was interpolated as a string — now uses length() instead.


NikolaMilosavljevic (approved):

eng/PublishSourceBuild.props
```diff
@@ -156,7 +156,7 @@
 
     <MSBuild Projects="$(MSBuildProjectFullPath)"
              Targets="Publish"
-             Condition="'$(DotNetSourceOnlyPublishArtifacts)' == 'true'"
+             Condition="'$(DotNetSourceOnlyPublishArtifacts)' == 'true' and '$(IsPublishPass2)' != 'true'"
```
ViktorHofer, line 159: I just noticed that this fix isn't present in main and forward ported it via https://github.com/dotnet/dotnet/pull/8276. I would double check that there isn't anything else from this PR missing.

PR #4945, opened 2026-02-18
BackgroundService: create task using TaskCreationOptions.LongRunning to prevent thread pool starvation

srnwmnn:
## Summary

This change updates how the background `ExecuteAsync` method is started within the `BackgroundService`.

Previously, `ExecuteAsync` was launched using `Task.Run`, which schedules work on the thread pool. This PR replaces that with `Task.Factory.StartNew` using `TaskCreationOptions.LongRunning` and `TaskScheduler.Default`.

## Rationale

- **Explicit long-running semantics**: `TaskCreationOptions.LongRunning` signals that the operation is expected to be long-lived, allowing the runtime to provision a dedicated thread instead of relying on the shared thread pool.
- **Improved thread pool behavior**: Avoids potential thread pool starvation for workloads where `ExecuteAsync` runs for the lifetime of the service.
- **Preserves existing behavior**: Cancellation semantics and lifecycle management remain unchanged; the host still observes and handles the `ExecuteAsync` task as before.

## Impact

- No functional changes to service startup or shutdown behavior.
- Improved alignment with best practices for long-running background operations.
- Reduced risk of thread pool contention under sustained load.


akoeplinger:
As per the contribution guidelines in https://github.com/dotnet/dotnet#contribution, please make this change in the dotnet/runtime repo directly. Thanks!

PR #4953, opened 2026-02-18
[release/10.0.2xx] Increase expiry of delegation sas token

mmitche:
The delegation SAS we generate has a 1 hour expiry right now by default. This has been the default forever in .NET builds. Typically we'd end up downloading the runtime or sdk or whatever within a couple minutes of build startup. However, VMR builds build SDK late, and therefore we tend to attempt download of the runtime bits after the hour expiry at times. Up the expiry based on this.


akoeplinger:
port to main?

PR #4957, opened 2026-02-18
Expose HTTP.sys kernel response buffering control for HttpListener on Windows

karimsalem1:
This PR adds an opt-in `AppContext` switch to enable kernel response buffering in HttpListener (Windows implementation only). When enabled, the flag `HTTP_SEND_RESPONSE_FLAG_BUFFER_DATA` is set on all calls to `HttpSendHttpResponse` and `HttpSendResponseEntityBody` (consistent with [Win32 docs](https://learn.microsoft.com/en-us/windows/win32/api/http/nf-http-httpsendresponseentitybody#:~:text=HTTP_SEND_RESPONSE_FLAG_BUFFER_DATA)). This mirrors the behavior for setting the `HttpSysOptions.EnableKernelResponseBuffering` flag in [ASP.NET HTTP.sys code](https://github.com/dotnet/aspnetcore/blob/2a88012113497bac5056548f16d810738b069198/src/Servers/HttpSys/src/RequestProcessing/ResponseBody.cs#L499).

Motivation in [issue #123425](https://github.com/dotnet/runtime/issues/123425).

Sample usage:
```csharp
AppContext.SetSwitch("System.Net.HttpListener.EnableKernelResponseBuffering", true);

using var listener = new HttpListener();
listener.Prefixes.Add("http://localhost:8080/");
listener.Start();

Console.WriteLine("Listening...");
while (true)
{
    var context = await listener.GetContextAsync().ConfigureAwait(false);
    _ = Task.Run(() => HandleRequestAsync(context));
}
```

Notes:
- No public API changes
- Default behavior remains unchanged
- Since API is internal, tests use reflection to validate flag behavior.
- Created a dedicated test class for Windows HttpListener and HttpResponseStream. Moved existing Windows-only tests into those classes for consistency.
- Added dedicated Windows-only test classes for `HttpListener` and `HttpResponseStream`, and consolidated existing Windows-only tests there for consistency.

**TODO**: add manual validation repro showing perf improvements using this feature

Thanks @ManickaP for all the help!

### DIFF
--- a/src/runtime/src/libraries/System.Net.HttpListener/tests/HttpListenerTests.Windows.cs
+++ b/src/runtime/src/libraries/System.Net.HttpListener/tests/HttpListenerTests.Windows.cs
@@ -12,7 +12,7 @@ namespace System.Net.Tests
     public class HttpListenerWindowsTests
     {
         [Fact]
-        public void EnableKernelResponeBuffering_DefaultIsDisabled()
+        public void EnableKernelResponseBuffering_DefaultIsDisabled()
         {
             using (var listener = new HttpListener())
             {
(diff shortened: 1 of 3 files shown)


ManickaP:
This should be made against dotnet/runtime, not dotnet/dotnet.

akoeplinger:
please recreate the PR in dotnet/runtime as mentioned in the contribution docs: https://github.com/dotnet/dotnet#contribution

PR #4937, opened 2026-02-18
Sign Source Asset for Source-Build Release

ellahathaway:
Related to https://github.com/dotnet/release/issues/1416

This PR enables the signing and signature verification of the source-built source-asset

[Dry-run signing run](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1298031&view=logs&j=514c02f0-737b-56ba-c340-bc19bd9edf9b)

PR #5043, opened 2026-02-23
[release/10.0.1xx] Sign Source Asset for Source-Build Release

ellahathaway:
Backport of https://github.com/dotnet/dotnet/pull/4937 to release/10.0.1xx

### DIFF
--- a/eng/PublishSourceBuild.props
+++ b/eng/PublishSourceBuild.props
@@ -4,7 +4,7 @@
   <!-- Add targets here that run logic that depends on the merged asset manifest and published packages. -->
   <UsingTask TaskName="Microsoft.DotNet.SourceBuild.Tasks.LeakDetection.CheckForPoison" AssemblyFile="$(MicrosoftDotNetSourceBuildTasksLeakDetectionAssembly)" TaskFactory="TaskHostFactory" Condition="'$(EnablePoison)' == 'true'" />
   <Target Name="ReportPoisonUsage"
-          AfterTargets="DiscoverArtifacts"
+          AfterTargets="Publish"
           DependsOnTargets="CreatePrivateSourceBuiltArtifactsArchive"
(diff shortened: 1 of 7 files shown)


ellahathaway:
This can be merged after the rebootstrap PR for the upcoming .NET 10 release has been merged. Adding the `No-MERGE` label until then.

ellahathaway:
@dotnet/source-build - can I please get a review on this? Thanks in advance.

eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/CreateSourceArtifact.cs
```diff
@@ -52,10 +52,17 @@ public class CreateSourceArtifact : BuildTask
     [Required]
     public required string OutputDirectory { get; init; }
 
+    /// <summary>
+    /// Create an empty detached signature file alongside the source artifact
+    /// Ensures the signature is published via darc when signing occurs post-build
+    /// </summary>
+    public bool CreateEmptyDetachedSignature { get; init; } = false;
```
mmitche, line 59: Curious as to why we don't get the artifact published if the empty detached sig doesn't exist pre-signing. When you say "post-build" signing, do you mean signing in the release pipeline?

ellahathaway: The source-build signing runs in a separate step after the build step. It is configured this way because signing occurs in an offline leg.

Because the vertical manifest is generated during the build step before the post-build signing step runs, the .sig file wouldn’t exist yet and therefore wouldn’t be picked up/published. Creating an empty detached signature placeholder up front ensures the .sig path is present in the vertical manifest, so when the signing step replaces it with the real signature, the artifact is still included and gets published.

PR #5079, opened 2026-02-25
Fix Debian md5sums file

GreenBeard:
They are supposed to have two (not one) spaces separating the md5 hash and the path as per https://manpages.debian.org/testing/dpkg-dev/deb-md5sums.5.en.html . It seems that upper case hashes are supposed to be allowed, but both `debsums` and `dpkg --verify` seem to complain in practice about the upper case hash so I have also changed it to be lower case.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/src/CreateMD5SumsFile.cs
+++ b/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/src/CreateMD5SumsFile.cs
@@ -43,7 +43,7 @@ namespace Microsoft.DotNet.Build.Tasks.Installers
 #if NET
                 writer.Write($"{Convert.ToHexStringLower(hash)}  {relativePath}\n");
 #else
-                writer.Write($"{BitConverter.ToString(hash).Replace("-", "")} {relativePath}\n");
+                writer.Write($"{BitConverter.ToString(hash).Replace("-", "").ToLower()}  {relativePath}\n");
 #endif
             }
 


src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/src/CreateMD5SumsFile.cs
```diff
@@ -41,7 +41,7 @@ public override bool Execute()
                 string relativePath = file.ItemSpec.Substring(RootDirectory.Length).TrimStart(Path.DirectorySeparatorChar).Replace('\\', '/');
                 // Always use Linux line-endings
 #if NET
-                writer.Write($"{Convert.ToHexString(hash)} {relativePath}\n");
+                writer.Write($"{Convert.ToHexStringLower(hash)}  {relativePath}\n");
 #else
                 writer.Write($"{BitConverter.ToString(hash).Replace("-", "")} {relativePath}\n");
```
akoeplinger: let's fix the netframework version too:

```suggestion
                writer.Write($"{BitConverter.ToString(hash).Replace("-", "").ToLower()}  {relativePath}\n");
```

GreenBeard: Yeah, sorry I'm dumb and was tired when I fixed this. I don't know how I missed something so obvious.

akoeplinger:
@mthalman should we backport this to 10.0?

src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/src/CreateMD5SumsFile.cs
akoeplinger: no worries, thanks for finding this clear bug :)

PR #5094, opened 2026-02-25
[release/10.0.1xx] Update ValueTuple and SqlClient to latest

ericstj:
Update to package versions with the following fixes:
- **SqlClient** -- updated package description https://github.com/dotnet/maintenance-packages/pull/228
- **ValueTuple** -- fixes for consumption on net471+ where the framework provides the assembly https://github.com/dotnet/maintenance-packages/issues/243

This change is already made in main - https://github.com/dotnet/dotnet/pull/4671

This is a backport + added one additional update in windowsdesktop.

## Customer Impact

- [x] Customer reported
- [ ] Found internally

When updating to the latest ValueTuple package, users may face an issue due to a missing reference, a warning-as-error from RAR about a version conflict, or unwanted bindingRedirect.  Fix this by including a partial framework list which tells MSBuild that the latest version of ValueTuple is part of .NETFramework (it is, and is handled by assembly unification).

## Regression

- [x] Yes
- [ ] No

Initially regressed when we moved ValueTuple from CoreFx to maintenance-packages.  When doing so we represented it as inbox in the latest versions of .NETFramework so that folks would no longer need to ship the assembly/package.  This broke things because the version in the .NET Framework targeting packs (reference assemblies) was lower than the one already in the nuget package.  Many customers will perceive this regression when getting those new packages transitively when updating to other 10.0 packages.

## Testing

Customers validated the fix, validated against all repros provided, adhoc repros.  Tested through ingestion in VMR + runtime.  Package already shipped, this is updating our dependencies to it.

## Risk

Low.  We pinned assembly versions when making this package change to avoid new bindingRedirects / mismatches between host/libraries.

### DIFF
--- a/src/roslyn/eng/Versions.props
+++ b/src/roslyn/eng/Versions.props
@@ -29,7 +29,7 @@
     <SystemNumericsVectorsVersion>4.6.1</SystemNumericsVectorsVersion>
     <SystemRuntimeCompilerServicesUnsafeVersion>6.1.2</SystemRuntimeCompilerServicesUnsafeVersion>
     <SystemThreadingTasksExtensionsVersion>4.6.3</SystemThreadingTasksExtensionsVersion>
-    <SystemValueTupleVersion>4.6.2</SystemValueTupleVersion>
+    <SystemValueTupleVersion>4.6.1</SystemValueTupleVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' != 'true'">
     <MicrosoftIORedistVersion>6.1.0</MicrosoftIORedistVersion>


ericstj:
I'm double checking with VS, MSBuild, and Roslyn that they'll be ok with this.

ericstj:
VS had challenges uptaking the fix since they already had a hand-authored bindingRedirect that downgraded ValueTuple to 4.0.0.0.  They can now remove that, but I don't want codeflow of roslyn to force them to do it.  https://github.com/dotnet/maintenance-packages/issues/258

@jjonescz undid this in main with https://github.com/dotnet/roslyn/pull/82639.

src/roslyn/eng/Versions.props
jozkee, line 41: Does this need to be bumped too? Asking because I saw dotnet/roslyn#82639 modified both `DotNetBuildSourceOnly` branches.

ericstj: Non-sourcebuild group does not need to be bumped here since in 10.0 they were still on the older (4.8 vs 4.9) version.  I only updated the packages which were on the versions we serviced.  Updating this one would imply a newer version which could involve more changes including more visual studio bindingRedirects.

You see both updated in main because they were on the newer versions for non-sourcebuild in 11.0.

PR #5173, opened 2026-03-02
Update source-mappings.json to exclude antlr4 non-C# code

jkoritzinsky:
Added exclusions for various antlr4 runtime files to limit compliance work.

Opening before we merge in https://github.com/dotnet/source-build-reference-packages/pull/1585

### DIFF
--- a/src/source-mappings.json
+++ b/src/source-mappings.json
@@ -95,7 +95,7 @@
             "include": [
                 "**/*",
                 "src/externalPackages/src/antlr4/runtime/CSharp/**"
-            ]
+            ],
             "exclude": [
                 // Contains disallowed binaries
                 // 1ES Secret Scanning does not allow any PFX files


src/source-mappings.json
```diff
@@ -96,7 +96,27 @@
...
+                "src/externalPackages/src/antlr4/scripts/**",
+                "src/externalPackages/src/antlr4/tool-testsuite/**",
+                "src/externalPackages/src/antlr4/tool/**"
+                "src/externalPackages/src/antlr4/runtime/Cpp/**",
+                "src/externalPackages/src/antlr4/runtime/Dart/**",
+                "src/externalPackages/src/antlr4/runtime/Go/**",
+                "src/externalPackages/src/antlr4/runtime/Java/**",
+                "src/externalPackages/src/antlr4/runtime/JavaScript/**",
+                "src/externalPackages/src/antlr4/runtime/Python3/**",
+                "src/externalPackages/src/antlr4/runtime/Swift/**"
```
jkoritzinsky: From what I can tell of the code that reads this file, we can't use more specific includes to re-include files that were previously excluded. As a result, we need to do a pretty wide set of excludes to only pull in the C# logic.

akoeplinger: Copilot told me this about the implementation:
```
  This works out nicely with the include filter in source-mappings.json. Here's why:

  The key behavior

  The GetSubmoduleFilters logic (line 631-638) only passes through include/exclude patterns
  that literally start with the submodule's path.
  The glob **/* does NOT start with e.g. ext/my-sub/, so it's filtered out.
  Then at line 130-136, if the submodule ends up with an empty include list, it
  defaults to **/* (include everything).

  Solution

  Use both a catch-all for the parent repo and a path-prefixed filter for the submodule:

   {
     "name": "my-repo",
     "defaultRemote": "https://github.com/org/my-repo",
     "defaultRef": "main",
     "include": [
       "**/*",
       "path/to/submodule/desired-directory/**/*"
     ]
   }
```

jkoritzinsky: I'll try that

jkoritzinsky: @akoeplinger updated to use includes.

PR #5174, opened 2026-03-02
Simplify RID graph: keep floating and last versioned RIDs

richlander:
Softer approach to RID graph simplification. Instead of removing all versioned and distro-specific RIDs, this keeps:

- All portable/base RIDs (any, unix, linux, win, etc.)
- All floating (unversioned) OS RIDs (alpine, ubuntu, osx, etc.)
- The last versioned RID for each OS family and its arch variants

This reduces runtime.json from ~4600 lines (798 RIDs) to ~1500 lines (266 RIDs) while preserving a migration path for packages that reference the most recent versioned RIDs.

Upstream PR: https://github.com/dotnet/runtime/pull/125076

### DIFF
--- a/src/runtime/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json
+++ b/src/runtime/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json
@@ -47,615 +47,51 @@
         "linux-musl-x86"
       ]
     },
-    "alpine.3.10": {
-      "#import": [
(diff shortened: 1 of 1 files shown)


richlander:
@omajid @tmds @ericstj @matteo-prosperi @dviererbe @elinor-fung

tmds:
Let's use https://github.com/dotnet/runtime/pull/125076 for feedback.

PR #5187, opened 2026-03-03
Reference Microsoft.NETCore.App in repos which are downstream of runtime

ericstj:
Retry https://github.com/dotnet/dotnet/pull/4801

Address https://github.com/dotnet/source-build/issues/5502#issuecomment-3970401495

Testing internally with https://dev.azure.com/dnceng/internal/_build/results?buildId=2916776&view=results

Will iterate on internal build pipeline to ensure this doesn't break official build.

### DIFF
--- a/repo-projects/Directory.Build.targets
+++ b/repo-projects/Directory.Build.targets
@@ -531,7 +531,7 @@
   </Target>
 
   <Target Name="SetBuildProperties"
-          DependsOnTargets="PrepareRepositoryReferences;SetRepoOriginalSourceRevisionId">
+          DependsOnTargets="PrepareRepositoryReferences;SetRepoOriginalSourceRevisionId;GetProducedPackagesFromTransitiveReferences">
     <!-- Make everything that depends on live runtime repo uses the live targeting pack to avoid a mismatch between the API/version packages expect to be in framework and what the actual targeting pack provides. -->
     <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '@(TransitiveRepositoryReference->AnyHaveMetadataValue('Identity', 'runtime'))' == 'true'">
       <!-- Use the targeting pack version as a shortcut for other pack versions since they all version consistently -->


ViktorHofer:
I think this will break the existing repo source build jobs. The tooling repos target net10.0 in repo source build as they use Arcade 10 which sets NetCurrent to net10.0.

Using live packs is a product build concept only. Also, these repos don't consume build dependencies from the VMR so the added Version.Details entries will never get updated.

That leads me to think that this is truely a VMR only concept and should be handled there.

ericstj:
> Testing internally with https://dev.azure.com/dnceng/internal/_build/results?buildId=2916776&view=results

It looks like this completed successfully, modulo some build flakiness which looks a bit like out-of-disk space on a single job.

> The tooling repos target net10.0 in repo source build as they use Arcade 10 which sets NetCurrent to net10.0.
> That leads me to think that this is truely a VMR only concept and should be handled there.

I see - in that case I would remove the D.B.props changes but keep the Version changes, then uncomment the global properties in the VMR project?  Tooling repos can always decide to add these properties to their own D.B.props if they see a backflow change that requires it.

ViktorHofer:
Yes that sounds like the right path forward.

I would also revert the Version.Details changes as they will be stale and don't match versions from other runtime dependencies in Version.Details.

ericstj:
> I would also revert the Version.Details changes as they will be stale and don't match versions from other runtime dependencies in Version.Details.

~~Doesn't VMR source build *require* these entries in order to flow the package properties (thus all those comments about being required for source-build).  My expectation was we add them once then they get fixed up after first flow.~~

I see what you mean - this is using _DependencyProducedPackage which is all the package outputs, not just those that happen to be subscribed.

ericstj:
All right using that approach.  Test build is https://dev.azure.com/dnceng/internal/_build/results?buildId=2917634&view=results.

Note to approvers -- ~~do not merge until the linked build is also green.~~ build is green ✅

ViktorHofer:
> I see what you mean - this is using _DependencyProducedPackage which is all the package outputs, not just those that happen to be subscribed.

Similarly to P2Ps it holds the outputs from direct and transitive repository references.

repo-projects/Directory.Build.targets
```diff
@@ -532,15 +532,14 @@
 
   <Target Name="SetBuildProperties"
           DependsOnTargets="PrepareRepositoryReferences;SetRepoOriginalSourceRevisionId">
-    <!-- Make everything that depends on live runtime target current pack versions to avoid assembly version mismatches with the LKG SDK that uses 10.0.0.0 versions when targeting net11.0.
-         Experimental, therefore commented out. -->
+    <!-- Make everything that depends on live runtime target current pack versions to avoid assembly version mismatches with the LKG SDK that uses 10.0.0.0 versions when targeting net11.0. -->
```
ViktorHofer: I would update this comment as 10.0.0.0 assembly versions aren't in play here anymore.

repo-projects/Directory.Build.targets
```diff
@@ -532,15 +532,14 @@
...
     <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '@(TransitiveRepositoryReference->AnyHaveMetadataValue('Identity', 'runtime'))' == 'true'">
       <RuntimePackVersion>@(_DependencyProducedPackage->WithMetadataValue('Identity', 'Microsoft.NETCore.App.Ref')->Metadata('Version'))</RuntimePackVersion>
```
ViktorHofer, line 538: Consider adding a comment explaining that the ref pack version is used as an approximation for the other pack versions.

ericstj:
Looks like our internal build passed, so this should avoid the previous regression here.  Thanks for your help @ViktorHofer -- I updated the comments.

ericstj:
I'm now skeptical if this is working.  I am hitting a similar failure as before in roslyn in source build -- https://dev.azure.com/dnceng-public/public/_build/results?buildId=1319208&view=logs&j=f8361ec4-8613-5f7f-2f10-d7dd7654e93f&t=f0818e4f-8d19-558f-794d-f0f07fd49e52
Seems to me like it's using new packages without new targeting pack.

Binlog shows values not set --
```
property:KnownFrameworkReferenceMicrosoftNETCoreAppCurrentTargetingPackVersion= --property:KnownFrameworkReferenceMicrosoftNETCoreAppCurrentDefaultRuntimeFrameworkVersion= --property:KnownFrameworkReferenceMicrosoftNETCoreAppCurrentLatestRuntimeFrameworkVersion= --property:KnownRuntimePackMicrosoftNETCoreAppCurrentMonoLatestRuntimeFrameworkVersion= --property:KnownRuntimePackMicrosoftNETCoreAppCurrentNativeAOTLatestRuntimeFrameworkVersion= 
```

repo-projects/Directory.Build.targets
```diff
@@ -532,15 +532,15 @@
 
   <Target Name="SetBuildProperties"
           DependsOnTargets="PrepareRepositoryReferences;SetRepoOriginalSourceRevisionId">
```
ViktorHofer:
```suggestion
          DependsOnTargets="PrepareRepositoryReferences;SetRepoOriginalSourceRevisionId;GetProducedPackagesFromTransitiveReferences">
```

ViktorHofer: I forgot to bring that change over back then.

ericstj:
Ok, need a new official build test for this.  https://dev.azure.com/dnceng/internal/_build/results?buildId=2918850&view=results

I'd also like to see https://github.com/dotnet/dotnet/pull/5206 make it past the roslyn failures around using the old targeting pack.

NO MERGE until those happen.

repo-projects/Directory.Build.targets
ericstj: Thank you @ViktorHofer

ericstj:
Official build failing with:
```
[xUnit.net 00:02:18.25]     Microsoft.DotNet.SourceBuild.Tests.SourcelinkTests.VerifySourcelinks [FAIL]
[xUnit.net 00:02:18.25]       Assert.True() Failure
[xUnit.net 00:02:18.25]       Expected: True
[xUnit.net 00:02:18.25]       Actual:   False
[xUnit.net 00:02:18.25]       Stack Trace:
[xUnit.net 00:02:18.25]         /_/test/Microsoft.DotNet.SourceBuild.Tests/SourcelinkTests.cs(68,0): at Microsoft.DotNet.SourceBuild.Tests.SourcelinkTests.VerifySourcelinks()
[xUnit.net 00:02:18.25]            at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
[xUnit.net 00:02:18.25]            at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
[xUnit.net 00:02:18.25]       Output:
[xUnit.net 00:02:18.25]         Executing: tar xzf /__w/1/s/artifacts/assets/Release/dotnet-symbols-all-11.0.100-preview.3.26154.118-centos.10-x64.tar.gz -C /__w/1/s/artifacts/bin/Microsoft.DotNet.SourceBuild.Tests/Release/SourcelinkTests/symbols
[xUnit.net 00:02:18.25]         Sourcelink verification failed for the following files:
[xUnit.net 00:02:18.25]         /__w/1/s/artifacts/bin/Microsoft.DotNet.SourceBuild.Tests/Release/SourcelinkTests/symbols/runtime/obj/System.Composition.Convention/Release/net11.0/System.Composition.Convention.pdb
```

Given the isolated and disconnected failure, I think it might be flaky.  Rerunning.  

https://github.com/dotnet/dotnet/pull/5206 made it through phase1 so I think this change is good from a completeness perspective.

ericstj:
> Given the isolated and disconnected failure, I think it might be flaky. Rerunning.

Indeed it was.  Passed on rerun.  Will file an issue.

PR #5210, opened 2026-03-03
Add SuppressValidation option for CreateLightCommandPackageDrop

joeloff:
Conditionally add `-sval` option for `Light` if original command included it.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix/wix.targets
+++ b/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix/wix.targets
@@ -354,9 +354,6 @@
       <_wixArgs>$(_wixArgs) @(DirectoryToHarvest -> '"%(WixObjFile)"', ' ')</_wixArgs>
       <_wixArgs>$(_wixArgs) $(AdditionalLightArgs)</_wixArgs>
       <_lightCommand>light.exe $(_wixArgs)</_lightCommand>
-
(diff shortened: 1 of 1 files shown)


src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix/wix.targets
```diff
@@ -354,6 +354,9 @@
       <_wixArgs>$(_wixArgs) @(DirectoryToHarvest -> '"%(WixObjFile)"', ' ')</_wixArgs>
       <_wixArgs>$(_wixArgs) $(AdditionalLightArgs)</_wixArgs>
       <_lightCommand>light.exe $(_wixArgs)</_lightCommand>
+
+      <_suppressValidation>false</_suppressValidation>
+      <_suppressValidation Condition="'$(_wixArgs.Contains(-sval))'">true</_suppressValidation>
```
mmitche: This looks wrong...does the `-sval` need to be quoted here?

mmitche: Is there any reason not to always just suppress validation?

joeloff: 1. I tested this in msbuild and it worked fine.
2. We could just always suppress it in the wixpack. We still need to include it for the initial linking.

mmitche (approved):

PR #5226, opened 2026-03-04
[Validation] Disable full signing by default on non-Windows platforms

aw0lid:
This change moves the global default for FullAssemblySigningSupported to false on non-Windows platforms within the Arcade SDK tools.

this is a validation PR in the VMR to ensure that disabling full signing (to avoid RSA+SHA-1 issues on modern Linux) does not impact official builds or other repository workflows.

Ref: dotnet/runtime#123010
Ref: dotnet/arcade#16566

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/StrongName.props
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/StrongName.props
@@ -21,6 +21,6 @@
   -->
   
   <PropertyGroup>
-    <FullAssemblySigningSupported Condition="'$(FullAssemblySigningSupported)' == '' and ('$(OS)' != 'Windows_NT' and '$(OfficialBuilder)' != 'Microsoft'))">false</FullAssemblySigningSupported>
+    <FullAssemblySigningSupported Condition="'$(FullAssemblySigningSupported)' == '' and '$(OS)' != 'Windows_NT' and '$(OfficialBuilder)' != 'Microsoft'">false</FullAssemblySigningSupported>
   </PropertyGroup>
 </Project>


src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/StrongName.props
```diff
@@ -22,6 +22,6 @@
   -->
   
   <PropertyGroup>
-    <FullAssemblySigningSupported Condition="'$(FullAssemblySigningSupported)' == '' and '$(DotNetBuildSourceOnly)' == 'true'">false</FullAssemblySigningSupported>
+    <FullAssemblySigningSupported Condition="'$(FullAssemblySigningSupported)' == '' and ('$(DotNetBuildSourceOnly)' == 'true' or '$(OS)' != 'Windows_NT')">false</FullAssemblySigningSupported>
```
mmitche: I think this will need to be accompanied by a change in some MS build pipeline yaml, since we do support this on non-Windows platforms (at least for now).

akoeplinger: > since we do support this on non-Windows platforms

what do you have in mind?

mmitche: Pass `FullAssemblySigningSupported=true` in our yaml. I think otherwise we will miss signing with the checked in full keys on non-Windows platforms.

mmitche: Alternatively, we could add an `OfficialBuilder == Microsoft` check to this.

aw0lid: I'm curious, why is it still necessary to support full signing on non-Windows platforms for official builds? Given that modern Linux distributions have deprecated RSA+SHA-1, what is the technical reason for keeping it enabled by default outside of Windows?

I was under the impression that for official Linux distro builds, full signing was already being disabled manually as part of the source-build process.

mmitche: Some packages today have full signing enabled using open keys: https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.Arcade.Sdk/tools/snk/

I can't remember the exact specifics of why and what scenarios this is used in. My guess would be packages that contain netframework binaries? Anyone remember? /cc @ericstj @dotnet/non-microsoft 

Anyways, if we want to eliminate this we'd need to be sure we didn't break some scenarios.

aw0lid: That's a fair point. I was hoping the VMR vertical builds and validation pipelines would catch any breakages in those scenarios.

If the CI passes, does that give us enough confidence, or are there specific legacy scenarios that these pipelines might miss? I believe we should address the root cause here; if full signing is only needed for a few specific packages, it might be better to enable it explicitly in those projects instead of keeping a global default that affects everyone on non-Windows platforms

aw0lid: > that contain netframework binaries

I think since .NET Framework is Windows-specific, and my change only affects non-Windows platforms, wouldn't those legacy scenarios be safe anyway? Any package that specifically requires full signing for .NET Framework compatibility would likely be built or validated on Windows, where the default remains true.
Am I missing a blind spot here?

ericstj: We do full signing because anything that uses the open key has a private key published for use in the open.  When we went OSS any new assemblies we created would opt for the open key rather than the old closed keys which have a secret private key.  A fully-signed assembly is better because you can use it in all scenarios on .NETFramework (like install to the GAC, partial trust, etc) and that was our goal for OSS -- people could do as much as possible themselves with our OSS projects as we can.

akoeplinger: >  Any package that specifically requires full signing for .NET Framework compatibility would likely be built or validated on Windows, where the default remains true.
Am I missing a blind spot here?

One case that comes to my mind is e.g. building the MSBuild tasks for Browser/wasm. We build that vertical on Linux but the produced task (which multi-targets netcore and netframework) needs to be able to be loaded into .NET Framework MSBuild on Windows.

aw0lid:
Hi all,

I noticed the Linux VMR build is failing in `efcore` with the following error:

```text
FSC(0,0): error FS2014: A problem occurred writing the binary '...Microsoft.EntityFrameworkCore.FSharp.FunctionalTests.dll': A call to StrongNameSignatureSize failed (Invalid Public Key blob)
```

This is the exact issue addressed in my F# compiler fix ([dotnet/fsharp#19242](https://github.com/dotnet/fsharp/pull/19242)). I believe the build is using a prebuilt F# compiler rather than building it from source, as the logs show:

`/.../build.sh ... /p:UsingToolMicrosoftNetCompilers=false`

This suggests it's using an outdated bootstrap compiler that doesn't have the fix yet.
Could someone confirm or clarify this?

**Is there a way to force the VMR to build and use the source-built F# compiler for this validation? Or should we wait for the VMR's bootstrap SDK to be updated with the fix?**

akoeplinger:
>  I believe the build is using a prebuilt F# compiler rather than building it from source, as the logs show /.../build.sh ... /p:UsingToolMicrosoftNetCompilers=false

That's correct, but it's unrelated to UsingToolMicrosoftNetCompilers (that property is specific to Roslyn-based compilers which the fsharp one isn't).

Most repos generally use the compilers which come with the bootstrap sdk, i.e. the one defined in the global.json in the VMR root. One exception to that is the `SB_CentOSStream10_Offline_CurrentSourceBuiltSdk_x64` job which uses the SDK built in a previous job to build itself i.e. it will have the latest changes from main (that job doesn't build repo tests though).

I think the easiest option is to wait until we do another rebootstrap to bump to a version of the sdk that includes the fsharp change.

aw0lid:
> I think the easiest option is to wait until we do another rebootstrap to bump to a version of the sdk that includes the fsharp change.

Got it. I'll keep this PR open until that happens, so we can verify the fix with the updated F# compiler. 
Thanks!

akoeplinger:
rebootstrapping happened and this is green now :)

aw0lid:
Great! Is the OfficialBuild property sufficient to cover all internal scenarios and avoid any regressions? I want to be 100% sure before moving this to https://github.com/dotnet/arcade
/cc @jkotas

akoeplinger:
yes that should be enough. we can also just merge it here, moving to arcade doesn't gain us much since nothing tests it there

aw0lid:
That sounds safer. Let's merge the change here first for further peace of mind. Once we verify that everything is solid in the next SDK build, we can then consider moving forward with the Arcade repo For compatibility and to avoid any conflict

akoeplinger:
the changes in src/arcade will get synced to the arcade repo as part of the normal codeflow.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/StrongName.props
```diff
@@ -18,10 +18,11 @@
...
                                       throwing an exception. For more details and an example, see
                                       https://github.com/dotnet/runtime/issues/65874. We default to full assembly
-                                      signing not being supported in source-only modes.
+                                      signing not being supported in source-only modes and on non-Windows platforms
+                                      when the OfficialBuilder is not Microsoft.
   -->
   
   <PropertyGroup>
-    <FullAssemblySigningSupported Condition="'$(FullAssemblySigningSupported)' == '' and '$(DotNetBuildSourceOnly)' == 'true'">false</FullAssemblySigningSupported>
+    <FullAssemblySigningSupported Condition="'$(FullAssemblySigningSupported)' == '' and ('$(DotNetBuildSourceOnly)' == 'true' or ('$(OS)' != 'Windows_NT' and '$(OfficialBuilder)' != 'Microsoft'))">false</FullAssemblySigningSupported>
```
jkotas: Would it make sense to delete `DotNetBuildSourceOnly` part of the condition to simplify this?

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/StrongName.props
```diff
@@ -18,10 +18,11 @@
                                       functional/available, and trying to use full signing results in the runtime
                                       throwing an exception. For more details and an example, see
                                       https://github.com/dotnet/runtime/issues/65874. We default to full assembly
-                                      signing not being supported in source-only modes.
+                                      signing not being supported in source-only modes and on non-Windows platforms
+                                      when the OfficialBuilder is not Microsoft.
```
jkotas: >. We default to full assembly
>                                      signing not being supported in source-only modes.
>                                      signing not being supported in source-only modes and on non-Windows platforms
>                                     when the OfficialBuilder is not Microsoft.

This part of the comment is not useful. It would be better to deleted it instead.

Comments that just rephrase what the next line of code does in plain English are not useful.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/StrongName.props
aw0lid: Good catch!

jkotas:
@aw0lid Thank you!

PR #5252, opened 2026-03-05
[main] Unblock msbuild FF - reset msbuild to latest 18.6 build

adamzip:
This PR was created using `darc vmr reset --build 304540 msbuild`


rainersigwald:
@adamzip do we need this? I think we're flowing msbuild, e.g. dotnet/dotnet#5711

PR #5253, opened 2026-03-05
Update AGENTS.md with pull request guideline

JanProvaznik:
Added section on working with pull requests to AGENTS.md.

In some cases reactions to changes in other constituent repos are required in codeflow PRs. Some people use Copilot CLI to do those changes, but Copilot CLI loves forcepushing, so let's nudge it away from that as it may break Maestro.

https://github.com/dotnet/dotnet/pull/5204#issuecomment-4000579452


JanProvaznik:
@ViktorHofer do you have an opinion if this should be in agents.md or in a skill?

PR #5275, opened 2026-03-06
[main] SBRP - improve file cloaking rules

adamzip:
https://github.com/issues/assigned?issue=dotnet%7Carcade-services%7C6053

Remove a pathspec from the inclusion rules for SBRP, because:
1. The pathspec seems to be redundant, as we already have `**/*` in the inlcusion
2. The paths no longer exist on VMR main, and are making the codeflow fail on `git rm` that is executed during opposite-direction flow


dkurepa:
I think this was mistakenly added as an inclusion rule in https://github.com/dotnet/dotnet/pull/5173 and we want to have it as an excluded, can you confirm @akoeplinger ?

adamzip:
Moved antlr4 to the exclusion rules as per @dkurepa 's suggestion

akoeplinger:
no this was intentionally an inclusion rule. we want it for https://github.com/dotnet/source-build-reference-packages/pull/1585 so that only the C# sources are mirrored into the VMR

dkurepa:
Ah ok, in that case, the sbrp FF will be broken until a build with the https://github.com/dotnet/source-build-reference-packages/pull/1585 flows

adamzip:
> Ah ok, in that case, the sbrp FF will be broken until a build with the [dotnet/source-build-reference-packages#1585](https://github.com/dotnet/source-build-reference-packages/pull/1585) flows

I think that any build will fail with this inclusion rule because it will attempt the inclusion on the VMR before the patch is applied, and it will fail with the same exception: 
`pathspec ':(glob,attr:!vmr-ignore)src/externalPackages/src/antlr4/runtime/CSharp/**' did not match any files`




> no this was intentionally an inclusion rule. we want it for [dotnet/source-build-reference-packages#1585](https://github.com/dotnet/source-build-reference-packages/pull/1585) so that only the C# sources are mirrored into the VMR

I don't understand exactly what the desired configuration is, but this doesn't seem to be it. As stated in this PR's description, it seems that adding the antlr4 filespec as an inclusion rule is redundant anyway because there's also inclusion on `**/*`, which will already include everything.

adamzip:
And btw the reason why we want to fix it is because the codeflow fails whenever one of the inclusion rules doesn't match any existing files on the VMR

akoeplinger:
>  it seems that adding the antlr4 filespec as an inclusion rule is redundant anyway because there's also inclusion on **/*, which will already include everything.

See the discussion at https://github.com/dotnet/dotnet/pull/5173#discussion_r2874789778, let me know if Copilot lied about the algorithm :D

akoeplinger:
@adamzip to be clear, we _only_ want the `src/externalPackages/src/antlr4/runtime/CSharp` directory, nothing else from `src/externalPackages/src/antlr4`

akoeplinger:
could we workaround it by manually adding the files to src/externalPackages/src/antlr4/runtime/CSharp in the VMR so that something is there that matches the inclusion glob when the next mirror happens?

adamzip:
> From what I can tell of the code that reads this file, we can't use more specific includes to re-include files that were previously excluded. As a result, we need to do a pretty wide set of excludes to only pull in the C# logic.

@akoeplinger This comment by Jeremy seems correct.
The inclusion rules are the first filter. Anything that is outside of the inclusion filter will definitely not make it into the codeflow.
Then the exclusion filter acts on top of that. Anything matching the exclusion rules also definitely won't be in the codeflow.

It doesn't go any further than that, because our service only applies all the inclusions first and then all the exclusions on top of that

adamzip:
It seems like one way or another, we'll either have to create inclusion rules that only includes what we want from antlr4, or exclusion rules that exclude everything we currently don't want, and I guess that either way the rules might have to be updated if new paths are added in antlr4

adamzip:
I suppose that this problem could be solved if instead of two separate lists for include and exclude, we'd have an ordered list that allows to mix included and excluded paths. then you'd have the full flexibility that git offers, and you'd be able to simply do something like this:
```JSON
[
    {
        "exclude": "src/externalPackages/src/antlr4/**",
    },
    {
        "include": "src/externalPackages/src/antlr4/runtime/CSharp/**"
    }
]
```

I'll bring this up with the team on Monday

akoeplinger:
@adamzip yup sounds good, whatever works and is easy to implement. do you think we could commit the files to the VMR manually in the meantime or will something break?

adamzip:
> @adamzip yup sounds good, whatever works and is easy to implement. do you think we could commit the files to the VMR manually in the meantime or will something break?

That sounds fine for now. If you add the antlr4 CSharp folder on the VMR side manually, the forward flow will be unblocked, and if you also have the antlr4 ready to forward flow from the SBRP side, I guess we should also add the long exclusion list that Jeremy intitially had

akoeplinger:
hmm yeah I guess we can just use the long exclusion list for now, then the next SBRP FF will add the CSharp files

akoeplinger:
https://github.com/dotnet/dotnet/pull/5279

src/source-build-reference-packages/.gitattributes
```diff
@@ -4,3 +4,6 @@
 *.patch text eol=lf
 
 *.sh text eol=lf
+
+# The CSharp runtime is the only folder from antlr4 we want in the VMR
+src/externalPackages/src/antlr4/runtime/CSharp/** vmr-keep
```
akoeplinger, line 9: I looked at that initially but my impression was it'd need to be in the .gitattributes of the submodule itself?

adamzip: From my understanding, if there are multiple .gitattributes files along a single path, then their values are combined.
I'll confirm this with a local test before we merge this PR.

https://git-scm.com/docs/gitattributes
> When deciding what attributes are assigned to a path, Git consults $GIT_DIR/info/attributes file (which has the highest precedence), .gitattributes file in the same directory as the path in question, and its parent directories up to the toplevel of the work tree (the further the directory that contains .gitattributes is from the path in question, the lower its precedence). Finally global and system-wide files are considered (they have the lowest precedence).

akoeplinger: @adamzip did the .gitattributes approach work out?

adamzip: No, you were right - only the .gitattributes that are inside the submodule repo itself count towards the patch.
I'll log an issue about a possible enhancement of adding our own .gitattributes when we create those patches, but for now the exclusion rules we have in SBRP are the only solution.

adamzip: https://github.com/dotnet/arcade-services/issues/6095

PR #5278, opened 2026-03-06
Add property to disable known pack version updates in tool

ViktorHofer:
Should fix the official build failures


ericstj:
@ViktorHofer do we want to do an official build to prove this, or just let it through and babysit?

ericstj:
Aside: I thought we were going to get a transport package from roslyn for this sort of shared infrastructure.

ViktorHofer:
> do we want to do an official build to prove this, or just let it through and babysit?

At this point, let's just get this in if the build succeeds to unbreak the official build. I'm pretty confident that this will fix the failures.

> Aside: I thought we were going to get a transport package from roslyn for this sort of shared infrastructure.

Nothing happened in that space, unfortunately. Aside: issues are also still in roslyn-analyzers.

ellahathaway:
@ViktorHofer - Windows_x86 leg failed so I disabled automerge.

ViktorHofer:
Yes, unrelated failure: https://github.com/dotnet/razor/issues/12758

PR #5279, opened 2026-03-06
Use full exclude list for antlr4 sources in source-mappings.json

akoeplinger:
https://github.com/dotnet/arcade-services/issues/6053
Unblocks the FF


adamzip (approved):

akoeplinger:
fast merge since this isn't tested in CI

PR #5314, opened 2026-03-09
vstest: Remove PATH search for dotnet executable

kasperk81:
Removed the logic for searching the PATH environment variable for the dotnet executable.

It will return null in worst case and Process.Start will find dotnet from PATH anyway.


kasperk81:
@nohwnd
@akoeplinger

akoeplinger:
This would make the method not return the full path anymore, not sure that makes sense (even if it might only be used in calls to Process.Start)

Also changes which don't touch multiple repos should be made in the respective individual repo github instead of in the VMR.

kasperk81:
@akoeplinger I left a comment in https://github.com/microsoft/vstest/pull/15442#discussion_r2895781652 which was fixed there, but https://github.com/dotnet/dotnet/pull/5183 was later merged without that fix.

This code is only used by `Process.Start`, which already performs the PATH lookup. That makes this logic redundant, since the OS maintains cached PATH lookups that `Process.Start` takes advantage of.

akoeplinger:
ok I see, I'd still rename the method then since it's not describing what it does but I'll leave it up to @nohwnd

src/vstest/src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -241,48 +241,12 @@ public static string CreateCommandLineArguments(ITestTask task)
...
-
-    /// <inheritdoc cref="ResolveDotnetPath(string)"/>
+    /// <returns>The resolved full path to the dotnet host.</returns>
     internal static string? ResolveDotnetPath()
     {
-        var dotnetExe = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "dotnet.exe" : "dotnet";
-
-        return ResolveDotnetPath(dotnetExe);
+        var dotnetHostPath = Environment.GetEnvironmentVariable("DOTNET_HOST_PATH");
+        return string.IsNullOrEmpty(dotnetHostPath) ? null : Path.GetFullPath(dotnetHostPath);
```
nohwnd:
```suggestion
        return StringUtils.IsNullOrEmpty(dotnetHostPath) ? null : Path.GetFullPath(dotnetHostPath);
```

src/vstest/src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
nohwnd:
```suggestion
```

src/vstest/src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -241,48 +240,12 @@ public static string CreateCommandLineArguments(ITestTask task)
     }
 
     /// <summary>
-    /// Resolves the full path to the dotnet executable by checking DOTNET_HOST_PATH,
-    /// current directory, and PATH in that order.
+    /// Resolves the full path to the dotnet host from the DOTNET_HOST_PATH environment variable, throwing if the variable is not set or empty.
```
akoeplinger: >  throwing if the variable is not set or empty

this doesn't seem to match with the implementation

kasperk81: comment updated

PR #5364, opened 2026-03-10
Use ResolveLinkTarget to replace realpath native call

kasperk81:


dsplaisted:
This looks like the same change as https://github.com/dotnet/sdk/pull/45637.

Probably we should use that PR instead of this one.  See https://github.com/dotnet/dotnet?tab=readme-ov-file#contribution

kasperk81:
it was just for testing the ci

PR #5389, opened 2026-03-11
[release/10.0.3xx] Reset razor 3xx on VMR to razor main branch

adamzip:
https://github.com/dotnet/arcade-services/issues/6058#issue-4033285066


src/source-manifest.json
```diff
@@ -34,13 +34,13 @@
       "barId": 302821,
```
jjonescz: Should the bar ID change too?

adamzip: Yes! thanks

PR #5397, opened 2026-03-11
Replace disabled feed for SB tests

mthalman:
Source-build tests for OmniSharp were taking a really long time attempting to restore packages because the NuGet.config was referencing a disabled feed. Updated to reference the active feed that provides the needed Microsoft.AspNetCore.OpenApi.10.0.6 package.

Fixes https://github.com/dotnet/source-build/issues/5514

### DIFF
--- a/eng/Version.Details.props
+++ b/eng/Version.Details.props
@@ -5,18 +5,18 @@ This file should be imported by eng/Versions.props
 -->
 <Project>
   <PropertyGroup>
-    <!-- _git/dotnet-dotnet dependencies -->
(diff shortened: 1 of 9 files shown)


akoeplinger:
great find, I wonder how we prevent this from happening again in the future

akoeplinger:
actually maybe we should just backport https://github.com/dotnet/dotnet/pull/4818 ?

mthalman:
> actually maybe we should just backport #4818 ?

No, I had tried that. It needs Microsoft.AspNetCore.OpenApi.10.0.6 which isn't on nuget.org.

PR #5414, opened 2026-03-12
Add exclusions for .pfx files in all SBRP submodules

adamzip:
https://github.com/dotnet/arcade-services/issues/5949

### DIFF
--- a/src/source-mappings.json
+++ b/src/source-mappings.json
@@ -94,9 +94,9 @@
             "defaultRemote": "https://github.com/dotnet/source-build-reference-packages",
             "exclude": [
               // Contains disallowed binaries
-              // 1ES Secret Scanning does not allow any PFX files
               "src/externalPackages/src/spectre-console/docs/**/*",
 
+              // 1ES Secret Scanning does not allow any PFX files
               // For submodule exclusions, we must use the full relative path of each submodule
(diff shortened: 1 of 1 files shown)


src/source-mappings.json
```diff
@@ -93,31 +93,44 @@
...
-                "src/externalPackages/src/antlr4/runtime/Go/**",
-                "src/externalPackages/src/antlr4/runtime/Java/**",
-                "src/externalPackages/src/antlr4/runtime/JavaScript/**",
-                "src/externalPackages/src/antlr4/runtime/Python3/**",
-                "src/externalPackages/src/antlr4/runtime/Swift/**"
+              // Contains disallowed binaries
+              // 1ES Secret Scanning does not allow any PFX files
+              "src/externalPackages/src/spectre-console/docs/**/*",
+
+              // For submodule exclusions, we must use the full relative path of each submodule
```
akoeplinger: I think the comment needs to be moved too
```suggestion
              "src/externalPackages/src/spectre-console/docs/**/*",

              // 1ES Secret Scanning does not allow any PFX files
              // For submodule exclusions, we must use the full relative path of each submodule
```

akoeplinger:
10.0 shouldn't need this since we ignored the .pfx files there

PR #5437, opened 2026-03-13
Fix Build.SourceVersionAuthor check for validation script

akoeplinger:
I noticed the check still failed in https://github.com/dotnet/dotnet/pull/5417, so the change in https://github.com/dotnet/dotnet/pull/4770#issuecomment-3887656953 didn't work.

This is because `Build.SourceVersionAuthor` looks at the git commit and that one uses `dotnet-maestro[bot]` as the author instead of `dotnet-bot`

### DIFF
--- a/eng/pipelines/pr.yml
+++ b/eng/pipelines/pr.yml
@@ -93,6 +93,7 @@ stages:
 
         - script: |
             set -ex
+            echo "Validating changes made by $(Build.SourceVersionAuthor)"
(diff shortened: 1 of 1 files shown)


akoeplinger:
hmm looks like I was wrong, Build.SourceVersionAuthor did print my github username and not the git author. No idea why the check doesn't work for dotnet-bot then...

akoeplinger:
Opened https://github.com/dotnet/dotnet/pull/5438 as an alternative

PR #5490, opened 2026-03-16
Enable runtime-async for all netcoreapp libraries

agocke:


agocke:
@mmitche huh, so interesting problem here. The offline source build is failing because crossgen2 is too old (doesn't have latest fixes). I see that the crossgen2 is preview3 so I assume it's pulling from unshipped bits. Do I have to do anything or will this fix itself by waiting for upstream from runtime? The change is from last week so it could just not have synced yet.

mmitche:
> @mmitche huh, so interesting problem here. The offline source build is failing because crossgen2 is too old (doesn't have latest fixes). I see that the crossgen2 is preview3 so I assume it's pulling from unshipped bits. Do I have to do anything or will this fix itself by waiting for upstream from runtime? The change is from last week so it could just not have synced yet.

Is it pulling crossgen from the SDK or from a package? If it's pacakge, then a new backflow will do it. If it's from the SDK, we'll need a new bootstrap and I don't know when the next one is scheduled.

agocke:
> Is it pulling crossgen from the SDK or from a package?

It's from the SDK (or at least, it's from a package that's tied to the SDK version). How would I set up a new bootstrap? Is there a process?

cc @jkoritzinsky

mmitche:
> > Is it pulling crossgen from the SDK or from a package?
> 
> It's from the SDK (or at least, it's from a package that's tied to the SDK version). How would I set up a new bootstrap? Is there a process?
> 
> cc @jkoritzinsky

https://github.com/dotnet/dotnet/blob/main/docs/VMR-re-bootstrapping.md

Documented here. There's a pipeline that can do it.

wtgodbe:
This was already done in dotnet/runtime, right?

Issue #5480, opened 2026-03-15
Internal pipeline: Validate Signing fails with "No space left on device" on Linux and Mac


steveisok:
## Summary

The "Validate Signing" task in internal unified-build pipeline runs out of disk space on Linux and Mac validation legs, causing builds to report `partiallySucceeded`. This has been recurring on `main` and `release/10.0.1xx` branches for at least 2 days.

## Error

```
/mnt/vss/_work/1/s/eng/signing-validation.proj(20,7): error : Signing validation failed: No space left on device
```

## Affected Builds

| Build | Branch | Date |
|-------|--------|------|
| [2926555](https://dev.azure.com/dnceng/internal/_build/results?buildId=2926555) | main | Mar 14 |
| [2926429](https://dev.azure.com/dnceng/internal/_build/results?buildId=2926429) | release/10.0.1xx | Mar 14 |
| [2926017](https://dev.azure.com/dnceng/internal/_build/results?buildId=2926017) | release/10.0.1xx | Mar 13 |
| [2925742](https://dev.azure.com/dnceng/internal/_build/results?buildId=2925742) | release/10.0.1xx | Mar 13 |

## Analysis

The signing-validation task iterates over all per-vertical artifact manifests and unpacks them for sign-check. With the growing number of verticals (30+ manifests observed in build 2926555 logs), the aggregate disk consumption exceeds the available space on the agent.

The failure occurs in both the "Validate Signing - Linux" and "Validate Signing - Mac" jobs within the same build, suggesting the issue is systematic rather than one-off.

## Impact

- Every internal rolling build on `release/10.0.1xx` reports `partiallySucceeded`
- Internal `main` builds are also affected
- The `partiallySucceeded` status masks real failures and makes pipeline health assessment unreliable

## Possible Mitigations

- Use larger disk agents for signing validation jobs
- Stream/validate manifests incrementally instead of unpacking all at once
- Clean up between vertical validations

/cc @dotnet/dotnet-engineering-services

akoeplinger:
I've not seen the Mac one run out of disk and it's not the case in any of the builds you linked

steveisok:
Good catch @akoeplinger — updated the title. The "No space left on device" errors were on the Linux validation legs. I incorrectly lumped Mac in.

akoeplinger:
https://github.com/dotnet/dotnet/pull/5525 fixes this but it showed a new issue on main: https://github.com/dotnet/dotnet/issues/5526

PR #5525, opened 2026-03-17
Run Validate Signing Linux job in a container

akoeplinger:
Run the ValidateSigning_Linux job inside the AzureLinux container instead of directly on the host agent. Add a disk space check step before artifact downloads on Linux.

There were a couple reports on FR recently about disk space issues when running directly on the build host, running inside a container works around this.

Fixes https://github.com/dotnet/dotnet/issues/5480

### DIFF
--- a/eng/pipelines/official.yml
+++ b/eng/pipelines/official.yml
@@ -137,6 +137,9 @@ extends:
       ${{ variables.ubuntuArmContainerName }}:
         image: ${{ variables.ubuntuArmContainerImage }}
         options: ${{ variables.defaultContainerOptions }}
+      ${{ variables.azurelinuxBuildContainerName }}:
+        image: ${{ variables.azurelinuxBuildContainerImage }}
+        options: ${{ variables.defaultContainerOptions }}
       ${{ variables.azurelinuxX64CrossContainerName }}:
         image: ${{ variables.azurelinuxX64CrossContainerImage }}
         options: ${{ variables.defaultContainerOptions }}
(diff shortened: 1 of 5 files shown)


akoeplinger:
For future readers: this wasn't actually related to the other disk size issue that was ongoing.

The reason why this works is because SignCheck uses Path.GetTempPath(): https://github.com/dotnet/dotnet/blob/f12904215d9817e5614391f7d2f8c9a6d198f5c7/eng/tools/tasks/Microsoft.DotNet.UnifiedBuild.Tasks/SigningValidation.cs#L58-L61

Outside the container we have /tmp explicitly mounted which has only 16GB free space:
```
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda3       252G   11G  231G   5% /
devtmpfs        4.0M     0  4.0M   0% /dev
tmpfs            16G  8.0K   16G   1% /dev/shm
tmpfs           6.3G  9.0M  6.3G   1% /run
tmpfs            16G  243M   16G   2% /tmp
/dev/sda2       452M   45M  373M  11% /boot
/dev/sdb1       196G  1.2G  185G   1% /mnt
/dev/ram11      2.8M   46K  2.4M   2% /mnt/storage
tmpfs           3.2G  4.0K  3.2G   1% /run/user/0
```

But inside the container the /tmp is mounted on `/` and has plenty of free space:
```
Filesystem      Size  Used Avail Use% Mounted on
overlay         252G  5.9G  236G   3% /
tmpfs            64M     0   64M   0% /dev
shm              64M     0   64M   0% /dev/shm
/dev/sdb1        98G  106M   93G   1% /__t
/dev/sda3       252G  5.9G  236G   3% /__a/externals
tmpfs           3.2G  9.0M  3.2G   1% /run/docker.sock
tmpfs           7.9G     0  7.9G   0% /proc/acpi
tmpfs           7.9G     0  7.9G   0% /proc/scsi
tmpfs           7.9G     0  7.9G   0% /sys/firmware
```

akoeplinger:
i.e. we need to keep using the container so we get the larger /tmp

PR #5543, opened 2026-03-17
Enable dynamic routing in official pipeline

mmitche:


mmitche:
I confirm this is the right approach from the TSA folks before merging.

PR #5595, opened 2026-03-19
Use sdk-task from VMR's eng/common instead of arcade

akoeplinger:
Reverts dotnet/dotnet#2447

The sdk-task script has been stable for some time, use the VMR copy instead. This ensures we're actually using the version of SignCheck defined in the VMR's nuget.config instead of the arcade one (which can be behind like we see right now due to https://github.com/dotnet/arcade/pull/16474 being blocked)


akoeplinger:
Fast merge since this isn't tested in public CI and we don't need to waste machine hours.

PR #5597, opened 2026-03-19
Remove auditSources from NuGet.Config

mmitche:
Removed auditSources section from Nuget.Client's NuGet.Config.


nkolev92:
I think we were able to get an exception for this. 

I know it's in our build and not the VMR, but this change would flow to NuGet and it's not something we'd want.

cc @zivkan @aortiz-msft

mmitche:
Closing in favor of disabling NuGet audit

PR #5600, opened 2026-03-20
Disable Nuget Audit in VMR builds

mmitche:


akoeplinger:
@ViktorHofer see the discussion on Teams. we need to get this in to fix the VMR build which got broken when we enabled CFSClean.

ViktorHofer:
Disabling this security feature doesn't feel right. If you can't remove the auditSource from nuget repo's NuGet.config for whatever reason, just remove the entry at build time. We already mutate each repo's NuGet.config.

akoeplinger:
I agree but we should unbreak the buuld first

ViktorHofer:
Ok. I'm just saying that this is the wrong course of action. There is no to do marker or any context on this PR suggesting that this is only for a short period of time.

ViktorHofer:
Cant you just set this property as a build argument only in nuget-client.proj?

akoeplinger:
ah I didn't see this was set for all projects, I agree we should only do it for nuget-client. I'd still prefer to get this PR in to fix the build and will file a follow-up PR

mmitche:
@ViktorHofer Is the concern about local builds? Because any official build will not be able to hit nuget.org.

mmitche:
> Disabling this security feature doesn't feel right. If you can't remove the auditSource from nuget repo's NuGet.config for whatever reason, just remove the entry at build time. We already mutate each repo's NuGet.config.

True, that seems like a better approach. But, either way, NuGet AUdit is not going to work in official builds against nuget.org, because that endpoint is blocked

mmitche:
@ViktorHofer I don't think it uses nuget.org by default if not specified, or the docs are wrong.

> If auditSources is not defined or is cleared without adding any sources, then packageSources will be used and warning NU1905 is suppressed.

https://learn.microsoft.com/en-us/nuget/concepts/auditing-packages#audit-sources

PR #5648, opened 2026-03-24
msbuild.dll instead of msbuild.exe

Nigusu-Allehu:


Nigusu-Allehu:
created for testing reasons. please ignore this PR

PR #5675, opened 2026-03-24
Fix typo: rename es-metatadata.yml to es-metadata.yml

danmoseley:
Fix typo in `src/runtime/src/libraries/es-metatadata.yml` — rename to `es-metadata.yml` (removing the extra "ta").

Every other `es-metadata.yml` across the VMR (26 files) and across 199+ internal AzDO repos uses the correct spelling.


danmoseley:
I don't know whether to put this in dotnet/runtime or not, but I think the files are actually "active" in this repo so here we are. I assume the rename will flow back.

danmoseley:
Without this I would imagine we're seeing code scannign issues get attributed to the wrong service tree...?

PR #5751, opened 2026-03-27
Prefer dotnet MSBuild.dll for worker nodes when hosted by dotnet.exe

rainersigwald:
When MSBuild is launched via dotnet.exe (e.g. dotnet build, dotnet msbuild), BuildEnvironmentHelper resolves CurrentMSBuildExePath to MSBuild.exe (the AppHost) because it prefers .exe over .dll. This causes all out-of-proc worker nodes to launch as MSBuild.exe AppHost processes, which seems to have regressed NuGet perf tests.

This change detects when the current process is dotnet.exe and the resolved MSBuild location is the AppHost, then substitutes MSBuild.dll so worker nodes are launched via dotnet.exe instead. This only applies to regular worker nodes (nodemode:1), not task host nodes (nodemode:2) which may need the AppHost for COM host object support.


rainersigwald:
Merged in MSBuild as https://github.com/dotnet/msbuild/pull/13452

PR #5769, opened 2026-03-28
Fix winforms CS0656: add Microsoft.CSharp reference for NetCurrent tests

steveisok:
## Problem

All rolling builds on `main` since Mar 25 fail with:

```
error CS0656: Missing compiler required member 'Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfo.Create'
```

in `NativeToManagedAdapterTests.cs` on `Windows_BuildTests_x64`. 3/3 consecutive rolling builds affected:
- [1353780](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1353780)
- [1354955](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1354955)
- [1356422](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1356422)

## Root Cause

[dotnet/winforms#14384](https://github.com/dotnet/winforms/pull/14384) multi-targeted the `System.Private.Windows.Core.Tests` project to include `net481`. It added `<Reference Include="Microsoft.CSharp" />` for `net481` (where it's a framework assembly), but **not** for `$(NetCurrent)-windows7.0`.

The test file uses `.Dynamic` TestAccessor (line 128+), which requires `Microsoft.CSharp.RuntimeBinder`. With `<DisableTransitiveFrameworkReferences>true</DisableTransitiveFrameworkReferences>` set in the project, the reference isn't available transitively.

## Fix

Add `<PackageReference Include="Microsoft.CSharp" />` for non-net481 targets, matching the existing pattern used for `System.Runtime.Serialization.Formatters`.


akoeplinger:
this PR got garbled and the description is wrong/misleading. the Microsoft.CSharp reference got added for net481 in the backflow https://github.com/dotnet/winforms/pull/14409 which made it into VMR main with https://github.com/dotnet/dotnet/pull/5709

the error message in the VMR build which failed before specifically complained about the net481 build so it should be fixed by that codeflow.

PR #5952, opened 2026-04-09
[release/10.0.3xx] Generate sarif file to avoid needing to update the SDK version on every branding update

mmitche:


marcpopMSFT:
@MichaelSimons didn't you make a change like this on the sdk repo side?

mmitche:
Will check this into sdk.

PR #6005, opened 2026-04-14
[release/10.0.3xx] Reset Vstest to rel/18.6 310166

dkurepa:
resets vstest contents to match the one at the sha that produced `310166`


marcpopMSFT:
Looks like vstest has moved to net10. Copilot suggested combining these lines together. Is this going to make it harder when we move to net12 though?

https://github.com/dotnet/dotnet/pull/6005/changes/d02ae1776eb59eeb9d2ca988430085c7a57e7cce

@dkurepa to review.

dkurepa:
> Looks like vstest has moved to net10. Copilot suggested combining these lines together. Is this going to make it harder when we move to net12 though?
> 
> [d02ae17](https://github.com/dotnet/dotnet/commit/d02ae1776eb59eeb9d2ca988430085c7a57e7cce)
> 
> @dkurepa to review.

I'm not sure, I'm afraid I don't have enough knowledge around this topic to answer that. @ViktorHofer would you be able to help here?

nohwnd:
Looks like the proposed change would search for our built artifacts always in net11.0 (netCurrent), but we build for net10.0 so it won't find them and fail.

We build for net10.0 unless we are in source build, where we build for netcurrent:

https://github.com/microsoft/vstest/blob/main/Directory.Build.targets#L20-L25

So the fix should look the same as it looks on main: 
https://github.com/dotnet/dotnet/blob/main/src/sdk/src/Layout/redist/targets/GenerateLayout.targets#L244-L245

 ```xml
   <Target Name="PublishTestCli">
    <PropertyGroup>
      <TestCliNuGetDirectoryTargetFramework>net10.0</TestCliNuGetDirectoryTargetFramework>
      <TestCliNuGetDirectoryTargetFramework Condition="'$(DotNetBuildSourceOnly)' == 'true'">$(NetCurrent)</TestCliNuGetDirectoryTargetFramework>
      <TestCliNuGetDirectory>$(NuGetPackageRoot)microsoft.testplatform.cli/$(MicrosoftTestPlatformCLIPackageVersion)/contentFiles/any/$(TestCliNuGetDirectoryTargetFramework)/</TestCliNuGetDirectory>
    </PropertyGroup>

```


We should also make some changes on vstest side to write the tfm matrix in the terms of arcade defined here, so we get automatically updated when moving up to net12: 

https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetFrameworkDefaults.props 


I am assuming that would allow us to produce net10 tfms up until net11 is published, and then arcade would update the definition for net minimum etc.

nohwnd:
That said I cannot think about a reason why we should not build for netcurrent in vmr as well,but that requires some small changes on vstest side as well.

marcpopMSFT:
> That said I cannot think about a reason why we should not build for netcurrent in vmr as well,but that requires some small changes on vstest side as well.

Well, I made it match main per your suggestion since it's still net10.

nohwnd:
was failing on system.collections.immutable, did same fix as we did in vstest main, probably happened after snap of 18.6 (what we are bringing here)

nohwnd:
@dkurepa builds but edits file that we should not edit, I guess you need to force merge it?

premun (approved):

PR #6037, opened 2026-04-15
Exclude workflows from the codeflow in all repos

dkurepa:
GitHub requires specific permissions to update workflows.
Rather than giving the Maestro app those permissions, we feel like it's a good idea to remove these from the VMR, since they're not used there anyway

Fsharp backflow is currently blocked on this. There's a possibility of this happening in more repos


src/source-mappings.json
jjonescz, line 1: now won't you need to give maestro permissions anyway, to *remove* the workflows?

dkurepa: Maestro won't try to remove them, this `exclusion` will affect the patch we create to flow the code, making it so it doesn't contain the workflows folder

jjonescz: Ah, you're right, that wouldn't make sense to remove them in the source repos, I got confused :)

T-Gro (approved):

dkurepa:
it looks like this is no longer needed

PR #6047, opened 2026-04-15
[release/10.0.1xx] Use `source-build-assets` repo

NikolaMilosavljevic:
After `source-build-reference-packages` repo was renamed to `source-build-assets`, changes are needed in VMR to allow proper consumption and flow.

This PR updates repo-related properties or values, as well as various other references in code and documentation. If something is not renamed, I'd prefer to prepare a follow up change in a new PR, to not delay this work any longer, considering timelines for next servicing release and code-complete date.

## Validation

[VMR build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2950877&view=results)


premun:
@NikolaMilosavljevic there are changes in an inlined submodule `MSBuildLocator`. I think we should not change that as once we would change the submodule in the SBA repo, it might fail to ingest as there would be conflicts.

If you need to have the text replacements done in `MSBuildLocator`, you'd ideally make those in that repo and then rewind the submodule in SBA and ingest that.

[Details about inlining](https://github.com/dotnet/dotnet/blob/main/docs/VMR-Full-Code-Flow.md#synchronizing-submodules).

premun:
Ah, the UI is showing it wrong, these are just regular renames with no changes. So I guess we can bypass-merge like this

premun (approved):

NikolaMilosavljevic:
Pushed the fix for License scanning as https://github.com/dotnet/dotnet/pull/6051 was merged this morning.

premun (approved):

NikolaMilosavljevic:
> Ah, the UI is showing it wrong, these are just regular renames with no changes. So I guess we can bypass-merge like this

Yeah, that's correct. I did not touch the submodule content, just moved it together with the repo.

PR #6090, opened 2026-04-16
[release/10.0.4xx] Initialize SDK band versions

marcpopMSFT:
Initialize SDK band versions on release/10.0.4xx.

Updated repositories:
- sdk: sdkminor 3 → 4, patch 0 → 00
- templating: patch 300 → 400
- sourcelink: patch 300 → 400

### DIFF
--- a/src/sdk/eng/Versions.props
+++ b/src/sdk/eng/Versions.props
@@ -6,8 +6,8 @@
   <PropertyGroup Label="Repo version information">
     <VersionMajor>10</VersionMajor>
     <VersionMinor>0</VersionMinor>
-    <VersionSDKMinor>3</VersionSDKMinor>
-    <VersionSDKMinorPatch>0</VersionSDKMinorPatch>
+    <VersionSDKMinor>4</VersionSDKMinor>
+    <VersionSDKMinorPatch>00</VersionSDKMinorPatch>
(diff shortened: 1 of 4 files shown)


src/sdk/eng/Versions.props
```diff
@@ -6,8 +6,8 @@
   <PropertyGroup Label="Repo version information">
     <VersionMajor>10</VersionMajor>
     <VersionMinor>0</VersionMinor>
-    <VersionSDKMinor>3</VersionSDKMinor>
-    <VersionSDKMinorPatch>0</VersionSDKMinorPatch>
+    <VersionSDKMinor>4</VersionSDKMinor>
+    <VersionSDKMinorPatch>00</VersionSDKMinorPatch>
```
mthalman, line 10: why is this changed to `00`?

marcpopMSFT: The mechanic did the change so I'm not sure. I'll change it back so it's consistent with prior previews. @mmitche on the mechanic behavior.?

mthalman (approved):

src/sdk/eng/Versions.props
mmitche: This isn't correct..should be 0. I'll debug this.

PR #6121, opened 2026-04-18
Use repo NuGet.Config to restore arcade

am11:
Before the recent change, we used to be able to build runtime repo without needing to have pwd==root_of_runtime_repo. Now it fails to restore Arcade SDK because repo's NuGet.Config is not respected. This PR restores the old behavior.


ViktorHofer (commented):
Thanks a lot. It would be good to search for the NuGet.config file with insensitive casing as some repos use a different format (i.e. nuget-client). https://github.com/dotnet/dotnet/blob/fa7dfcf7852b275589d02500e8e977b896151653/repo-projects/Directory.Build.props#L56-L57

am11:
Thanks, @ViktorHofer. I've made the search case-insensitive at depth 0 (only searching the repo root). I think NuGet officially supports three casings: NuGet.Config, NuGet.config, and nuget.config (https://github.com/NuGet/NuGet.Client/blob/de3b92c2768037cceac21a7bcc53318d42554428/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L29). Other variants will not work, but that error will be handled by the toolset.

Here, we've just made it permissive so the script can pass. If it makes sense, I can add upfront validation in the script for the supported casings, but I think that would introduce maintenance overhead when the list changes upstream.

ViktorHofer (approved):

ViktorHofer:
Thanks, this is good enough. Can you please make the same changes to src/arcade/eng/common? That is the actual outgoing location for eng/common updates. The current updates wouldn't flow to repos as they only touch files from the bootstrap layer.

ViktorHofer (approved):

PR #6139, opened 2026-04-20
Mark net10.0 as SupportedNETCoreAppTargetFramework

wtgodbe:
Work around net10.0 being marked as unsupported on VS17 in https://github.com/dotnet/sdk/pull/52235


vseanreesermsft (approved):

akoeplinger:
@wtgodbe @vseanreesermsft should we port this to main/10.0.2xx+ ?

wtgodbe:
> should we port this to main/10.0.2xx+ ?

It's not needed there, the underlying bug was already fixed in dotnet/sdk

PR #6009, opened 2026-04-14
Split SBRP license scanning into separate sub-scans

mthalman:
Split the source-build-reference-packages license scan into separate pipeline jobs - one per subdirectory under src/source-build-reference-packages/src/ plus a root scan with --ignore src for non-src content. This avoids the scancode timeout caused by processing too many files in a single run (dotnet/source-build#5470).

Fixes dotnet/source-build#5470

PR #6172, opened 2026-04-21
Fix license scan exclusion regex to match ignore patterns at any depth

mthalman:
The negative lookaheads for ignore patterns only checked the first path component after the repo root, but scancode `--ignore` matches directory names at any depth. For source-build-assets, exclusion paths like `src/source-build-assets/src/externalPackages/...` have an intermediate `src/` directory, so the lookahead `(?!externalPackages/)` didn't filter them. Changed to `(?!(?:.*/)?externalPackages/)` to match at any depth.

This fixes the issue with `source-build-assets` paths being [removed from the exclusions file](https://github.com/dotnet/dotnet/pull/6056).

Regression from https://github.com/dotnet/dotnet/pull/6009


ellahathaway (approved):

mthalman:
I wanted to wait for a successful SDK diff run here to validate this is really working before backporting. So let's leave those opened PRs for now.

PR #6221, opened 2026-04-23
Use repo NuGet.Config to restore arcade

am11:
Apply changes to runtime repo https://github.com/dotnet/dotnet/pull/6121 as it was most affected one. This will hopefully speed the flow.

### DIFF
--- a/src/runtime/eng/common/tools.ps1
+++ b/src/runtime/eng/common/tools.ps1
@@ -746,9 +746,19 @@ function InitializeToolset() {
   }
 
   $downloadArgs = @("package", "download", "Microsoft.DotNet.Arcade.Sdk@$toolsetVersion", "--prerelease", "--output", "$nugetCache")
-  if ($env:NUGET_CONFIG) {
+  $nugetConfig = $env:NUGET_CONFIG
+  if (-not $nugetConfig) {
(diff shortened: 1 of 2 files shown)


am11:
cc @ViktorHofer

ViktorHofer:
This change doesn't do anything. In the VMR, all repos that build after arcade use the src/arcade/eng/common layout so they already have the fix. Also, what flows to repos is src/arcade/eng/common.

am11:
Ah ok. So the next VMR sync in runtime will copy src/arcade/eng/common from VMR? I was confused because

https://github.com/dotnet/arcade/blob/main/eng/common/tools.sh and
https://github.com/dotnet/runtime/blob/main/eng/common/tools.sh

are out of sync (as-in runtime one seems "newer" with the broken state and arcade one doesn't have that change yet, let alone the fix for broken state)

ViktorHofer:
The dotnet -> arcade dependency flow is blocked on https://github.com/dotnet/arcade/pull/16726 but that doesn't matter. Runtime doesn't consume anything from that standalone repo, it gets the assets from the VMR.

The current state is due to VMR official builds failing for various reasons. People are working on it.

PR #6293, opened 2026-04-27
Remove templating

MichaelSimons:
Related to https://github.com/dotnet/templating/issues/10085

Templating was consolidated into the sdk.


eng/allowed-vmr-binaries.txt
```diff
@@ -119,9 +119,6 @@ src/source-build-assets/src/externalPackages/src/application-insights*/WEB/Src/W
 # symreader
 src/symreader/src/PdbTestResources/Resources/*
 
-# templating
```
NikolaMilosavljevic, line 122: Are we going to have these in `sdk`?

NikolaMilosavljevic (approved):

eng/allowed-vmr-binaries.txt
MichaelSimons: Yes, they were added right away then the src was migrated into sdk.  https://github.com/dotnet/dotnet/blob/main/eng/allowed-vmr-binaries.txt#L114

ViktorHofer (approved):

PR #6402, opened 2026-04-30
Add workflow to auto-apply milestones to codeflow PRs

wtgodbe:
## Summary

Adds a GitHub Action that automatically applies milestones to child-repo PRs when codeflow PRs merge into dotnet/dotnet. Triggered whenever `src/source-manifest.json` is modified on `main` or any `release/**` branch.

## How it works

1. **Diffs `source-manifest.json`** (merge commit vs parent) to find which repo changed and its old→new commit range
2. **Discovers child PRs** via GitHub GraphQL `associatedPullRequests` on each commit in the range (paginated for large ranges)
3. **Resolves the milestone name** based on the target branch (see corner cases below)
4. **Creates the milestone** in the child repo if it doesn't exist, then applies it to each discovered PR

### Release branches with suffix (e.g. `release/11.0.1xx-preview5`)
Milestone is extracted directly from the branch name: `11.0-preview5`.

### Main branch
Reads `PreReleaseVersionLabel` and `PreReleaseVersionIteration` from `eng/Versions.props` at the merge commit. The label determines behavior:

- **`alpha`**: Always maps to `preview1` (alpha builds ship as part of the first preview).
- **`preview`/`rc` with no matching release branch**: Normal case. Milestone is `{major}.{minor}-{label}{iteration}` (e.g. `11.0-preview5`, `11.0-rc1`).
- **`preview` with matching release branch exists**: Branch was just snapped but branding hasn't been bumped yet. We can't tell if the next milestone is another preview or rc1, so we **defer** by labeling the dotnet/dotnet PR with `needs-milestone`. The next run after branding is updated will process deferred PRs.
- **`rc` with matching release branch exists**: This is the **vNext snap** — main has become the next major version (e.g. main says `rc1`/`11.0` but `release/11.0.1xx-rc1` exists -> main is really `12.0`). Milestone is `{major+1}.0-preview1`.
- **`servicing`**: Skip (shouldn't happen on main).

### Bare release branches (e.g. `release/11.0.1xx`)
These are used during the RC->RTM phase. After snapping `release/11.0.1xx-rc1`, the `release/11.0.1xx` branch becomes the staging branch for rc2 and eventually RTM.

- **`VersionSDKMinorPatch` == 0 AND a sibling `rc1` or `rc2` branch exists**: Active pre-RTM branch. Milestone is read from the branch's own `eng/Versions.props` branding (e.g. `11.0-rc2`).
- **`VersionSDKMinorPatch` != 0**: Servicing branch, skip.
- **No sibling rc branches**: Not in the RC->RTM phase, skip.
- **Branded as `servicing` or empty**: Already shipped RTM, skip.

### Other release branch formats
Skipped (no-op).

## Deferred milestone processing

When the workflow can't determine the correct milestone (preview->rc ambiguity during branch snaps), it labels the dotnet/dotnet PR with `needs-milestone` instead. On subsequent runs against `main` where the milestone resolves successfully, the workflow processes all `needs-milestone`-labeled PRs. The label is only removed after successful processing; failures preserve it for retry.

## Cross-organization support

Repos can be in the `dotnet` or `microsoft` GitHub orgs. A GitHub App installed on both orgs provides separate tokens via `actions/create-github-app-token`. Repos in other orgs are skipped.

## Opt-out

Repos listed in `.github/milestone-config.json` are skipped. Currently opted out:
- Repos with independent versioning schemes (roslyn, msbuild, fsharp, razor, vstest, etc.)
- Repos with non-standard major versions (cecil, diagnostics, command-line-api, etc.)
- NuGet repos (ship on their own cadence)

## Dry-run mode

The workflow supports `workflow_dispatch` with a `dry-run` input (default: true) and a `pr-number` input for testing against any merged PR. In dry-run mode, all read/resolution logic runs normally but no milestones are created, applied, or labels changed.

## Error handling

- **Per-repo isolation**: A failure processing one child repo doesn't block others.
- **Per-PR isolation**: A failure milestoning one PR doesn't block others.
- **Milestone creation race**: Handles 422 "already exists" from concurrent runs.
- **Concurrency**: Workflow runs are serialized per target branch to prevent race conditions in deferred processing.
- **Pagination**: Both commit comparison (250/page) and milestone listing are fully paginated.

## Prerequisites

Before enabling, a GitHub App needs to be created and installed on the `dotnet` and `microsoft` orgs with the following permissions:
- **Issues**: Read & Write (create milestones, apply milestones to PRs, manage `needs-milestone` label)
- **Pull Requests**: Read (list files, get PR details)
- **Contents**: Read (read `source-manifest.json` and `eng/Versions.props` at specific commits)
- **Metadata**: Read (required by all GitHub Apps)

The app ID and private key go into repository secrets `MILESTONE_APP_ID` and `MILESTONE_APP_PRIVATE_KEY`.

## Files

- `.github/workflows/apply-milestones.yml` — the workflow
- `.github/milestone-config.json` — opt-out configuration

## Removed pre-existing automation

This PR also removes preview/rc milestone automation from child repo `resourceManagement.yml` policies that are now superseded by the centralized workflow:

- **aspnetcore**: Removed rules milestoning PRs merged to `main` (11.0-preview5) and `release/11.0-preview4` (11.0-preview4)
- **winforms**: Removed rule milestoning PRs merged to `main` (10.0 Preview7)
- **wpf**: Removed rule milestoning PRs merged to `main` (7.0 Preview4 — very stale)

Servicing milestone policies (e.g. aspnetcore's release/10.0 -> 10.0.9) are left intact since the new workflow only handles preview/rc/RTM milestones, not servicing.

### DIFF
--- a/.github/workflows/apply-milestones.yml
+++ b/.github/workflows/apply-milestones.yml
@@ -101,7 +101,7 @@ jobs:
 
             // --- Step 1: Resolve milestone name ---
 
-            async function resolveMilestone(branch) {
+            async function resolveMilestone(branch, mergeCommitSha) {
 
               // --- Preview/RC release branches ---
               // release/11.0.1xx-preview4 → 11.0-preview4
@@ -210,7 +210,28 @@ jobs:
                 return `${ver}-${fullLabel}`;
               }
 
-              // Branch exists → branding is stale. Deterministically increment.
+              // Branch exists → branding may be stale, OR the PR was merged just before the snap.
+              // Check if this PR's merge commit is in the release branch.
+              // If it is, the PR shipped in this preview — use current branding.
+              if (mergeCommitSha) {
+                try {
+                  const { data: comparison } = await github.rest.repos.compareCommits({
+                    owner: 'dotnet', repo: 'dotnet',
+                    base: mergeCommitSha,
+                    head: candidateBranch,
+                  });
+                  // If merge commit is behind or equal to the branch, it's included
+                  if (comparison.status === 'ahead' || comparison.status === 'identical') {
+                    core.info(`Merge commit ${mergeCommitSha.substring(0, 8)} is in ${candidateBranch} — using current branding`);
+                    return `${ver}-${fullLabel}`;
+                  }
+                } catch (e) {
+                  core.warning(`Could not check branch reachability: ${e.message}`);
+                }
+              }
+
+              // Merge commit is NOT in the release branch — branding is stale.
+              // Deterministically increment.
               // RC snap means main is now vNext
               if (label === 'rc') {
                 return `${major + 1}.0-preview1`;
@@ -462,13 +483,16 @@ jobs:
 
             // For workflow_dispatch, fetch the PR to get its base branch and checkout its merge commit
             let baseBranch;
+            let mergeCommitSha;
             if (context.payload.pull_request) {
               baseBranch = context.payload.pull_request.base.ref;
+              mergeCommitSha = context.payload.pull_request.merge_commit_sha;
             } else {
               const { data: pr } = await github.rest.pulls.get({
                 owner: 'dotnet', repo: 'dotnet', pull_number: prNumber,
               });
               baseBranch = pr.base.ref;
+              mergeCommitSha = pr.merge_commit_sha;
               if (!pr.merged) {
                 core.setFailed(`PR #${prNumber} has not been merged`);
                 return;
@@ -480,7 +504,7 @@ jobs:
 
             core.info(`Processing PR #${prNumber} against ${baseBranch}`);
 
-            const milestoneName = await resolveMilestone(baseBranch);
+            const milestoneName = await resolveMilestone(baseBranch, mergeCommitSha);
 
             if (!milestoneName) {
               core.info('Could not resolve milestone, skipping');
--- a/.github/workflows/apply-milestones.yml
+++ b/.github/workflows/apply-milestones.yml
@@ -98,6 +98,11 @@ jobs:
               return null;
             }
 
+            // "alpha" always maps to preview1
+            if (label === 'alpha') {
+              return { name: `${major}.${minor}-preview1`, deferred: false };
+            }
+
             // Combine label + iteration: "preview" + "5" → "preview5"
             const fullLabel = `${label}${iteration}`;
 
--- a/.github/workflows/apply-milestones.yml
+++ b/.github/workflows/apply-milestones.yml
@@ -23,6 +23,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
           fetch-depth: 2  # Need parent commit for diff
 
       - name: Generate dotnet org token
@@ -86,7 +87,7 @@ jobs:
               // Read patch version from eng/Versions.props on this branch
               const fs = require('fs');
               const versionsProps = fs.readFileSync('eng/Versions.props', 'utf8');
-              const patchMatch = versionsProps.match(/<VersionPatch>(\d+)<\/VersionPatch>/);
+              const patchMatch = versionsProps.match(/<VersionSDKMinorPatch>(\d+)<\/VersionSDKMinorPatch>/);
               const patch = patchMatch ? parseInt(patchMatch[1]) : -1;
 
               if (patch !== 0) {
@@ -270,23 +271,35 @@ jobs:
           // --- Step 3: Find PRs in commit range using GraphQL ---
 
           async function findPRsInRange(owner, repo, oldSha, newSha, octokit) {
-            // Get commits in range — handle large ranges by paginating
-            const { data: comparison } = await octokit.rest.repos.compareCommits({
-              owner, repo,
-              base: oldSha,
-              head: newSha,
-            });
+            // Collect all commits in range, paginating if > 250
+            const allCommits = [];
+            let currentBase = oldSha;
 
-            if (comparison.commits.length === 0) {
-              return [];
+            while (true) {
+              const { data: comparison } = await octokit.rest.repos.compareCommits({
+                owner, repo,
+                base: currentBase,
+                head: newSha,
+                per_page: 250,
+              });
+
+              if (comparison.commits.length === 0) break;
+
+              allCommits.push(...comparison.commits);
+
+              // If we got all commits, we're done
+              if (allCommits.length >= comparison.total_commits) break;
+
+              // Move the base forward to the last commit we got
+              currentBase = comparison.commits[comparison.commits.length - 1].sha;
             }
 
-            if (comparison.total_commits > 250) {
-              core.warning(`${owner}/${repo} has ${comparison.total_commits} commits in range — only first 250 will be processed`);
-            }
+            if (allCommits.length === 0) return [];
+
+            core.info(`${owner}/${repo}: ${allCommits.length} commits in range`);
 
             // Batch query associatedPullRequests via GraphQL
-            const commitAliases = comparison.commits.map((c, i) =>
+            const commitAliases = allCommits.map((c, i) =>
               `c${i}: object(oid: "${c.sha}") { ... on Commit { associatedPullRequests(first: 5) { nodes { number } } } }`
             );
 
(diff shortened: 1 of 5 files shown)


.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,539 @@
...
+
+            return milestone.number;
+          }
+
+          async function applyMilestone(owner, repo, prNumber, milestoneNumber, octokit) {
+            if (DRY_RUN) {
+              core.info(`[DRY RUN] Would apply milestone to ${owner}/${repo}#${prNumber}`);
+              return;
+            }
+
```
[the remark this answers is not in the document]
wtgodbe: Good catch. Added a .filter(i => i.pull_request) before iterating.

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,539 @@
...
+            let milestone = openMilestones.find(m => m.title === milestoneName);
+
+            if (!milestone) {
+              const closedMilestones = await octokit.paginate(octokit.rest.issues.listMilestones, {
+                owner, repo, state: 'closed', per_page: 100,
+              });
+              milestone = closedMilestones.find(m => m.title === milestoneName);
+            }
+
+            if (!milestone) {
```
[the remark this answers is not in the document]
wtgodbe: Fixed — retry path now uses octokit.paginate matching the primary lookup.

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,539 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
+          fetch-depth: 2  # Need parent commit for diff
+
+      - name: Generate dotnet org token
+        uses: actions/create-github-app-token@v1
+        id: dotnet-token
+        with:
```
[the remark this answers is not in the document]
wtgodbe: Good point — updated the PR description to list the full required permissions: Contents read, Pull Requests read, Issues read/write, and Metadata read.

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,539 @@
...
+
+            for (const issue of deferredPRs) {
+              core.info(`Processing deferred PR #${issue.number}`);
+              try {
+                await processOnePR(issue.number, milestoneName);
+
+                // Only remove label if processing fully succeeded
+                await github.rest.issues.removeLabel({
+                  owner: 'dotnet', repo: 'dotnet',
+                  issue_number: issue.number,
```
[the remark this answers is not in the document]
wtgodbe: Fixed — now stripping trailing .git after URL parsing.

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,539 @@
...
+                owner, repo,
+                base: currentBase,
+                head: newSha,
+                per_page: 250,
+              });
+
+              if (comparison.commits.length === 0) break;
+
+              allCommits.push(...comparison.commits);
+
```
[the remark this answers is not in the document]
wtgodbe: Bumped to first: 10. In practice a single commit in a child repo is almost always associated with 1 PR (squash merge) or occasionally 2-3 (merge commits). Full pagination would add significant query complexity for negligible benefit.

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,543 @@
...
+        id: microsoft-token
+        with:
+          app-id: ${{ secrets.MILESTONE_APP_ID }}
+          private-key: ${{ secrets.MILESTONE_APP_PRIVATE_KEY }}
+          owner: microsoft
+
+      - name: Apply milestones
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ steps.dotnet-token.outputs.token }}
```
mmitche: Why pass the dotnet-token and the microsoft token differently here?

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,543 @@
...
+        env:
+          MICROSOFT_TOKEN: ${{ steps.microsoft-token.outputs.token }}
+          DRY_RUN: ${{ github.event.inputs.dry-run || 'false' }}
+        script: |
+          const { Octokit } = require("@octokit/rest");
+          const microsoftOctokit = new Octokit({ auth: process.env.MICROSOFT_TOKEN });
```
mmitche: Suggest using a map here of org->auth.

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,543 @@
...
+
+          // Supported orgs for milestoning. Repos outside these orgs are skipped.
+          const SUPPORTED_ORGS = new Set(['dotnet', 'microsoft']);
+
+          // Repos that have opted out of auto-milestoning — loaded from .github/milestone-config.json
+          const fs = require('fs');
+          const config = JSON.parse(fs.readFileSync('.github/milestone-config.json', 'utf8'));
+          const OPTED_OUT_REPOS = new Set(config.optedOutRepos.map(r => r.toLowerCase()));
+
+          function getOctokit(owner) {
```
mmitche: See above.

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,543 @@
...
+            // These are used during the RC→RTM phase.
+            const bareMatch = branch.match(/^release\/(\d+\.\d+)\.(\d+)xx$/);
+            if (bareMatch) {
+              const ver = bareMatch[1];        // "11.0"
+              const featureBand = bareMatch[2]; // "1"
+
+              // Read patch version from eng/Versions.props on this branch
+              const fs = require('fs');
+              const versionsProps = fs.readFileSync('eng/Versions.props', 'utf8');
+              const patchMatch = versionsProps.match(/<VersionSDKMinorPatch>(\d+)<\/VersionSDKMinorPatch>/);
```
mmitche: Versions.props changes will break things here..best to guard.

.github/workflows/apply-milestones.yml
```diff
@@ -0,0 +1,543 @@
...
+                    branch: `release/${ver}.${featureBand}xx-${rcN}`,
+                  });
+                  if (rcN === 'rc1') rc1Exists = true;
+                  if (rcN === 'rc2') rc2Exists = true;
+                } catch (e) {
+                  if (e.status !== 404) throw e;
+                }
+              }
+
+              if (!rc1Exists && !rc2Exists) {
```
mmitche: What's the purpose of skipping in this case?

mmitche (commented):
I think the YAML can be a good deal simpler. it looks like it got overcomplicated. I think this rule set basically works:

- If on a release/<preview/rc> branch, then take the branding from the versions.props file in the branch.
- If on main or release/<non preview/rc> then take the branding from the pre-release info in the version.props file in the branch, adjusting + 1 if the preview/RC n branch exists. Create a small func that determines what + 1 will be (p7 -> rc1, rc2 -> ga)
- If the branch is branded ga/servicing, or the adjustment leads to you ga,. ignore.

.github/workflows/apply-milestones.yml
wtgodbe: Good point -- the asymmetry was because github-script injects the `github` octokit from `github-token`, while the microsoft one had to be manually constructed. Unified it: now both tokens are passed as env vars and there's an `orgClients` map of org->client. The `github` client from github-script is still used for dotnet org (since it also powers the `github.rest.repos.getBranch` calls for branch existence checks), but the auth pattern is consistent.

.github/workflows/apply-milestones.yml
wtgodbe: Done -- replaced SUPPORTED_ORGS set + separate microsoftOctokit variable with an `orgClients` map. `getOctokit(owner)` now does a map lookup and returns null for unconfigured orgs. Adding a new org is a one-liner in the map.

.github/workflows/apply-milestones.yml
wtgodbe: Fixed -- getOctokit now uses the orgClients map.

.github/workflows/apply-milestones.yml
wtgodbe: Removed

wtgodbe:
Holding off until I'm back from vacation 5/18

mmitche (approved):
I went through the logic and as far as I can tell, I think it's correctly identifying the correct milestone. The only place I wasn't completely positive was the rc2 timeframe, but it looks about right.

PR #6518, opened 2026-05-07
Reset nuget-client to build 310908

martinrrm:
Fixes VMR unsafe code flow PR


ViktorHofer (changes requested):
Many changes are undone here

premun:
@martinrrm I think you might have not committed the `src/source-manifest.json` file where the metadata is.

Can you update the PR description with the build ID / commit that you have used to do the reset?

nkolev92:
We don't need this.

PR #6570, opened 2026-05-12
Merging internal commits for release/10.0.3xx

vseanreesermsft:


src/sdk/NuGet.config
```diff
@@ -4,6 +4,8 @@
     <clear />
     <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
     <!--  Begin: Package sources from dotnet-dotnet -->
+    <add key="darc-int-dotnet-dotnet-b16286c" value="https://pkgs.dev.azure.com/dnceng/internal/_packaging/darc-int-dotnet-dotnet-b16286c2/nuget/v3/index.json" />
```
ericstj, line 7: @vseanreesermsft  should we be including these internal feeds?

ericstj (approved):
Assuming adding internal feeds is ok since I see instances already.

T-Gro (approved):

akoeplinger:
Build failure:
> src/sdk/src/Microsoft.DotNet.TemplateLocator/Microsoft.DotNet.TemplateLocator.csproj(0,0): error NU1102: (NETCORE_ENGINEERING_TELEMETRY=Restore) Unable to find package Microsoft.Deployment.DotNet.Releases with version (>= 2.0.0-rtm.1.26229.119)

I also see a lot of PublicAPI.Unshipped.txt‎ files changing file mode from 644 to 755, that seems unexpected

src/sdk/NuGet.config
mmitche: Yes, those are fine to include. They're disabled below.

mmitche:
The missing Deployment.Tasks.Releases package is interesting here. It's because it's a shipping package on an isolated feed that did not get released to nuget.org. I believe that is expected for this package.

The issue is that it's produced out of the 1xx build and so needs to be available when building 3xx. In 1xx, the deployment tools get built before sdk so the packages are live there. If the package was a non-shipping package, it would have ended up on the public dotnet10-transport feed, and that would have been fine.

Either something needs to change with deploiyment.tasks.releases (e.g. shipping to nuget.org on release days) @joeloff. Or..maybe it ends up non-shipping?

ericstj:
The 10.0.2xx PR has this same problem -- https://github.com/dotnet/dotnet/pull/6571/checks?check_run_id=75657706047

joeloff:
We periodically ship 1.0.x on NuGet. That package was never intended to ship on the same cadence as the rest, but SDK took a dependency on the deployment-tools repo and then bumped the version to 2.0.0 preview. and that has created a lot of issue.

joeloff:
@mmitche, can we move the package from https://pkgs.dev.azure.com/dnceng/internal/_packaging/darc-int-dotnet-dotnet-94ea8265/nuget/v3/index.json to dotnet10 for now and then figure out how to fix it. That way we can unblock the codeflow

joeloff:
Had a long chat with @ericstj need to run a few things by some other folks, but I think we have a plan for the long term to start versioning and shipping multiple releases from each of the servicing branches, so we'd have a 10.0.x and 11.0.x, etc. and  that way we don't need to do anything special and SDKs can consume it like they would any other package

NikolaMilosavljevic:
@vseanreesermsft can this be merged now?

NikolaMilosavljevic:
Is this ready for merging? It blocks https://github.com/dotnet/dotnet/pull/6567

PR #6572, opened 2026-05-12
Merging internal commits for release/10.0.1xx

vseanreesermsft:


ericstj (approved):

src/aspnetcore/src/Middleware/ResponseCaching/test/TestUtils.cs
```diff
@@ -301,6 +301,7 @@ internal class LoggedMessage
     internal static LoggedMessage ResponseNotCached => new LoggedMessage(27, LogLevel.Information);
     internal static LoggedMessage ResponseContentLengthMismatchNotCached => new LoggedMessage(28, LogLevel.Warning);
     internal static LoggedMessage ExpirationInfiniteMaxStaleSatisfied => new LoggedMessage(29, LogLevel.Debug);
+    internal static LoggedMessage RequestContainsInvalidCacheSymbols => new LoggedMessage(30, LogLevel.Warning);
```
DeagleGross:
```suggestion
    internal static LoggedMessage RequestContainsInvalidCacheSymbols => new LoggedMessage(30, LogLevel.Debug);
```

ericstj:
@DeagleGross -- was there a problem with the codeflow here?  If you're making an additional change like this you'll need to make sure that's going to work for tests (which won't build or run here) as well as ensure the corresponding changes are made to other active codebases branches (which won't ever see flow from this branch).

ericstj (approved):
I chatted with @DeagleGross about these - he'll make sure to flow the same fix to all the right places.  He's tested it independently and confirmed it's working and necessary for aspnetcore tests (once they backflow)

PR #6590, opened 2026-05-13
Set _IsPublishing=true and add ProjectReference for PublishDotnetAot

dsplaisted:
## Summary

Follow-up to #6576. Two related changes to `PublishDotnetAot` in `src/sdk/src/Layout/redist/`:

1. **Pass `_IsPublishing=true` as a global property** to the `<MSBuild Targets="Publish">` call on `dotnet-aot.csproj`. Per @MichalStrehovsky's feedback on #6576, running the Publish target without `_IsPublishing=true` has been a source of subtle issues — it's the property the SDK uses to gate Publish-aware behaviors (suppressing apphost generation when `PublishAot=true`, several `NETSdkError` conditions, `_RuntimeIdentifierUsesAppHost` adjustments, etc.).

2. **Add a `ProjectReference` from `redist.csproj` to `dotnet-aot.csproj`** (`ReferenceOutputAssembly=false`, `SkipGetTargetFrameworkProperties=true`, `Private=false`). This sequences the outer solution build's evaluation of `dotnet-aot.csproj` to complete before redist's `GenerateSdkLayout` target runs, eliminating the race surface re-introduced by change (1).

cc @ericstj @MichalStrehovsky @NikolaMilosavljevic

### DIFF
--- a/src/sdk/src/Layout/redist/redist.csproj
+++ b/src/sdk/src/Layout/redist/redist.csproj
@@ -75,8 +75,15 @@
          Without an explicit reference, the outer Build (writing to bin\dotnet-aot\...) and the
          Publish call's internal Build (writing to the same paths) can run concurrently and
          race over shared bin\ and obj\ directories. This ProjectReference sequences the outer
-         Build to complete before redist's GenerateSdkLayout target runs. -->
-    <ProjectReference Include="$(RepoRoot)src\Cli\dotnet-aot\dotnet-aot.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" Private="false" />
+         Build to complete before redist's GenerateSdkLayout target runs.
+
+         Conditioned on the same TargetRid prefix check as the PublishDotnetAot target itself;
+         when NativeAOT publish doesn't run (e.g. source-build with TargetRid like centos.10-x64),
+         pulling dotnet-aot.csproj into the restore graph would demand the ILCompiler RID-specific
+         runtime pack, which isn't available from source-built feeds (NU1100). -->
+    <ProjectReference Include="$(RepoRoot)src\Cli\dotnet-aot\dotnet-aot.csproj"
+                      Condition="'$(TargetRid)' == '$(HostRid)' and ($(TargetRid.StartsWith('win')) or $(TargetRid.StartsWith('linux')) or $(TargetRid.StartsWith('osx')))"
+                      ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" Private="false" />
 
     <ProjectReference Include="$(RepoRoot)template_feed\*\*.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" Private="false" />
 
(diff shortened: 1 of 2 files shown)


ViktorHofer (approved):

ericstj:
Does this project ever have a non-publishing mode?  If it doesn't, can it just always have the publish property set?  The risk for using global-properties like this is it can cause a rebuild if for any reason some target input changes - or you can get different results from build targets (intentional or not) that mean the output of build and publish differ.  This can cause cases like shipping something different than test.   In either case, if user's need to set this maybe it should be a public property - or we need a documented/designed way to do a publish via project-reference with minimal rebuild and no public properties.  Best practice design for such systems would not need to replay any targets from build and would make all behavior differences in the added targets required for publish.

nagilson (approved):

nagilson:
If we set _IsPublishing then the build, restore and other commands/ actions on the project would have the publish specific settings also run (e.g. build would use publish trimmed) - it's a bit of a weird pattern. If we could block building the project then this could be interesting. I'd also see if @rainersigwald has ideas.

dsplaisted:
@ViktorHofer @nagilson @ericstj I had to make a fix so this needs another approval.

As far as setting _IsPublishing directly in the project, I think that would probably work.  It might be something we should look into but I prefer not to do it here.

ViktorHofer (approved):
Please log a follow-up issue to follow the pattern used in finalzer.csproj for publishing as part of the build.

src/sdk/src/Layout/redist/redist.csproj
```diff
@@ -69,6 +69,22 @@
     <ProjectReference Include="$(RepoRoot)src\Dotnet.Watch\dotnet-watch\dotnet-watch.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" Private="false" />
     <ProjectReference Include="$(RepoRoot)src\Dotnet.Format\dotnet-format\dotnet-format.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" Private="false" />
 
+    <!-- The PublishDotnetAot target in GenerateLayout.targets calls <MSBuild Targets="Publish"> on
+         dotnet-aot.csproj. That call passes _IsPublishing=true, which makes it a different
+         BuildManager project instance than the outer build's evaluation of dotnet-aot.csproj.
+         Without an explicit reference, the outer Build (writing to bin\dotnet-aot\...) and the
+         Publish call's internal Build (writing to the same paths) can run concurrently and
```
rainersigwald, line 76: This seems reasonable, but I don't think I fully understand why the solution is "sequence the overwrite" rather than "avoid the overwrite by getting a distinct output path in the publish case". Not blocking but please consider that.

dsplaisted: It sounds like the guidance from Viktor is to set _IsPublishing=true via evaluation so there's no overwriting and no separate output path.  I've filed https://github.com/dotnet/sdk/issues/54324 for that.

dsplaisted:
> Please log a follow-up issue to follow the pattern used in finalzer.csproj for publishing as part of the build.

Thanks, I've filed https://github.com/dotnet/sdk/issues/54324.  Please take a look to verify I understood things correctly :-)

PR #6594, opened 2026-05-13
Remove vestigial test_templates at sdk's old TemplateEngine.TestTemplates path

dsplaisted:
## Summary

Remove the unreferenced `src/sdk/test/TemplateEngine/Microsoft.TemplateEngine.TestTemplates/` directory from the VMR. These files are leftovers from when `dotnet/templating` was forward-flowing into the VMR and the test templates lived at that path in dotnet/sdk.

## Background

* `dotnet/sdk` PR [#53733](https://github.com/dotnet/sdk/pull/53733) (merged 2026-04-07 by @MichaelSimons) moved the test template fixtures from `test/TemplateEngine/Microsoft.TemplateEngine.TestTemplates/` to `test/TestAssets/TestPackages/TemplateEngine/`.
* `dotnet/templating` has since been consolidated into `dotnet/sdk` (README + default branch now `archive`) and is no longer in `src/source-manifest.json`.
* The VMR's copy of the sdk source mirrored the new location but never deleted the old-location files. They're now vestigial.

## Why this matters

Both `sdk.slnx` and `TemplateEngine.slnf` reference only the new location (`test/TestAssets/TestPackages/...`). The old-location files are unreferenced inside the VMR.

But every backflow from this VMR to dotnet/sdk re-introduces them in dotnet/sdk (because dotnet/sdk's main no longer has them after #53733). The duplicated content then gets picked up by `test/UnitTests.proj`'s `**\*.Tests.csproj` glob, which only excludes `TestAssets\**\*.Tests.csproj` — not the old path. Result on every codeflow PR: `NETSDK1138` (net7.0 EOL) and `NU1008` (inline Version + CPM) on `MyTestProject.Tests.csproj`. Blocked dotnet/sdk PR #54292.

Deleting the old path here makes future backflows clean and removes the recurring failure.

## What this PR does

`git rm -r src/sdk/test/TemplateEngine/Microsoft.TemplateEngine.TestTemplates/` — 324 files, 6193 deletions, no additions.

I verified there are no references to this path anywhere in the VMR (only the NEW location is referenced in sdk.slnx / TemplateEngine.slnf).

cc @MichaelSimons @marcpopMSFT


marcpopMSFT (approved):

marcpopMSFT:
CC @premun as it feels weird that this is needed to keep the vmr from bringing the files back during codeflow.

premun:
We will have a look. I think someone recently reverted an SDK flow in the VMR so this might be an unwanted outcome

premun:
@marcpopMSFT we've started seeing some weirdness after Monday's rollout. Will roll back and investigate

PR #6620, opened 2026-05-14
Configure nuget audit to not block CI builds but still fail PR builds

marcpopMSFT:
- Remove nuget audit disablement that was put in place for servicing
- Check what fails in the PR and CI builds
- See if we can configure WarnNotAsError for these warnings in the official build

### DIFF
--- a/eng/tools/Directory.Build.props
+++ b/eng/tools/Directory.Build.props
@@ -9,4 +9,10 @@
     <RestoreSources Condition="Exists('$(SharedComponentsArtifactsPath)')">$(RestoreSources);$(SharedComponentsArtifactsPath)</RestoreSources>
   </PropertyGroup>
 
+  <!-- Disable NuGet audit in official builds. NuGet Restore does not respect
+       WarningsNotAsErrors, so audit findings become errors via TreatWarningsAsErrors. -->
+  <PropertyGroup Condition="'$(OfficialBuild)' == 'true'">
+    <NuGetAudit>false</NuGetAudit>
+  </PropertyGroup>
+
 </Project>
--- a/eng/tools/Directory.Build.props
+++ b/eng/tools/Directory.Build.props
@@ -9,8 +9,9 @@
     <RestoreSources Condition="Exists('$(SharedComponentsArtifactsPath)')">$(RestoreSources);$(SharedComponentsArtifactsPath)</RestoreSources>
   </PropertyGroup>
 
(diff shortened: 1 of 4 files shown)


vseanreesermsft (approved):

src/sdk/Directory.Build.props
```diff
@@ -79,20 +79,16 @@
     <GenerateProgramFile>false</GenerateProgramFile>
   </PropertyGroup>
 
-  <!-- TODO: Re-enable NuGet Audit and fix alerts: https://github.com/dotnet/sdk/issues/51465 -->
-  <PropertyGroup>
+  <!-- Disable NuGet audit in official builds. NuGet Restore does not respect
```
nkolev92: NuGet Restore does respect WarningsNotAsErrors, it's MSBuild that has bugs. 

https://github.com/dotnet/msbuild/issues/10801

eng/tools/Directory.Build.props
```diff
@@ -9,13 +9,10 @@
...
-  <!-- Suppress known NuGet audit vulnerabilities in VMR tooling projects.
-       These are compile-only references to Microsoft.Build packages whose transitive
-       dependency on System.Security.Cryptography.Xml has no patched upstream version yet. -->
-  <ItemGroup>
-    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-37gx-xxp4-5rgx" />
-    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-w3x6-4m5h-cxqf" />
-    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-g4vj-cjjj-v7hg" />
-  </ItemGroup>
+  <!-- Disable NuGet audit in official builds. NuGet Restore does not respect
+       WarningsNotAsErrors, so audit findings become errors via TreatWarningsAsErrors. -->
```
nkolev92: same here: https://github.com/dotnet/msbuild/issues/10801. 

NuGet.Client does actually use WarningsNotAsErrors, but we had to stop using MSBuildTreatWarningsAsErrors to stop hitting the msbuild bug.

src/sdk/Directory.Build.props
marcpopMSFT: Ahh, I wasn't sure why it wasn't working, just that it I didn't work. I'll get the comment updated that way we have accurate tracing and will more likely switch this to warnings not as errors once the msbuild bug is fixed.

wtgodbe (approved):

PR #5256, opened 2026-03-05
WIP: Set SOURCE_DATE_EPOCH env var during the build

akoeplinger:
NuGet is adding support for deterministic builds (https://github.com/NuGet/Home/pull/14785) to make .NET more reproducible for https://github.com/dotnet/source-build/issues/4963

This is a WIP PR to check if setting the SOURCE_DATE_EPOCH env var (which is already supported by various other tools like clang) causes any issues right now.

PR #6622, opened 2026-05-14
Add reproducibility validation test for SDK tarball

### LINKED ISSUE #5486: Define a reproducibility validation test for the SDK tarball
## Summary

Add a test that validates the reproducibility of the .NET SDK tarball build by comparing the output of two builds of the same commit. The test asserts the two tarballs are identical (or differ only in known, baselined ways).

## Background

The parent issue (dotnet/source-build#4963) tracks making .NET's build deterministic/reproducible. A key part of this effort is having a test that continuously validates reproducibility so regressions are caught. There is currently no test that validates **byte-for-byte reproducibility** of the build output.

### Test Location

The test should live in `test/Microsoft.DotNet.Tests/` alongside existing VMR-level tests.

### Test Class

Create a new `ReproducibilityTests` class that:

1. **Accepts paths to two SDK tarballs** built from the same source commit.
2. **Extracts both tarballs** and compares their contents.
3. **Performs byte-level comparison** of all files, identifying any files that differ between the two builds.
4. **Uses exclusion system** to filter out known non-reproducible files.
5. **Uses baseline comparison system** to compare the list of differing files against a baseline. Any new non-reproducible file causes the test to fail; any file that becomes reproducible causes the baseline to be stale (and should be updated).
6. **Generates updated baseline/exclusion files** in the logs directory so that CI can publish them for easy baseline maintenance.

### Conditional Test Execution

The test should be conditional on both tarball paths being provided.

### Obtaining the Two Builds

The priority right now is to test reproduceability for source-only builds so these will be the build types that are targeted for this workflow.

The test pipeline follows the same pattern as `eng/pipelines/templates/jobs/sdk-diff-tests.yml` to locate and download artifacts from a prior `dotnet-unified-build` run:

1. **Find the associated `dotnet-unified-build` run.** Use the same logic as `sdk-diff-tests.yml`:
   - If triggered by a pipeline resource, use the triggering pipeline's run ID and commit SHA.
   - Otherwise, query for the latest completed run on the current branch.

2. **Download the SDK tarball** from the identified `dotnet-unified-build` run using the existing `download-vmr-artifact.yml` template. This becomes the "first build" tarball.

3. **Checkout the VMR at the same commit** as the `dotnet-unified-build` run.

4. **Run a second build** of the SDK at that commit, producing a second SDK tarball. This build should use the same build flags/configuration as the original.

5. **Run the `ReproducibilityTests`** passing both tarball paths.

mthalman:
## Summary

Adds an automated test that validates the SDK tarball produced by the VMR build is byte-reproducible: building the same source twice should produce identical artifacts. The test fails the pipeline whenever a new non-deterministic file appears that isn't either a known root-cause-justified exclusion or already in the recorded baseline.

## Architecture

The pipeline (eng/pipelines/unified-build-reproducibility-tests.yml) builds the VMR twice within a single pipeline run and compares the two SDK tarballs. Building twice inside the same pipeline run ensures both halves share the same pipeline-resolved container image digest (so the same clang/LLD toolchain), the same `Build.SourceVersion`, and the same `Build.BuildNumber` - eliminating environmental drift as a source of false-positive diffs.

The stage template (eng/pipelines/templates/stages/reproducibility-tests.yml) defines three stages:

* `Build1` - source-build leg + Microsoft leg (Linux x64).
* `Build2` - identical to `Build1`, runs in parallel.
* `Compare` - downloads both stages' SDK tarballs and runs the test job.

The `reproducibility-test` job template (eng/pipelines/templates/jobs/reproducibility-test.yml) downloads the two SDK tarballs from the current pipeline's `Build1` and `Build2` stages and runs the test against both tarballs.

## Test implementation

ReproducibilityTests.cs reads both tarballs, computes a `SHA256` hash for each file in each tarball, and compares the two hash maps. Files matching patterns in the exclusions file are skipped before comparison; the remaining diffs are formatted as a sorted list and compared against a recorded baseline file. The test fails when the diff list doesn't match the baseline.

Fixes https://github.com/dotnet/source-build/issues/5486


mthalman:
@omajid - PTAL

test/Microsoft.DotNet.Tests/Microsoft.DotNet.Tests.csproj
ViktorHofer, line 1: This test project is quite overloaded already and receives a ton of different parameters. Consider adding a specialized test project for this instead and have a test utilities library to share common test code.

eng/pipelines/unified-build-reproducibility-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+    buildSourceOnly: true
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  parameters:
+    settings:
+      networkIsolationPolicy: Permissive, CFSClean, CFSClean2
+    pool:
+      name: NetCore1ESPool-Svc-Internal
+      image: 1es-ubuntu-2204
```
akoeplinger:
```suggestion
      name: $(DncEngInternalBuildPool)
      image: build.azurelinux.3.amd64
```

eng/pipelines/unified-build-reproducibility-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+      os: linux
+    sdl:
+      sourceAnalysisPool:
+          name: NetCore1ESPool-Svc-Internal
```
akoeplinger:
```suggestion
          name: $(DncEngInternalBuildPool)
```

eng/pipelines/unified-build-reproducibility-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+  - pipeline: dotnet-unified-build
+    source: dotnet-unified-build
+    # For releases branches only run on internal/release branches because that's where dependencies flow.
+    # Previews don't have internal/release branches so they must be run from non-internal release branches.
+    trigger:
+      branches:
+        include:
+        - refs/heads/release/*.0.1xx-preview*
+        - refs/heads/internal/release/*.0.?xx*
+        - refs/heads/internal/release/*.0.???
```
akoeplinger, line 26: can you explain how this works when we have both schedule trigger and this one here?

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,166 @@
...
+      echo "##vso[task.setvariable variable=SecondSdkTarballPath]$(Pipeline.Workspace)/RebuildArtifacts/$second_sdk_tarball"
+    displayName: Locate second SDK tarball
+
+  - script: |
+      eng/common/build.sh -bl --projects $(Build.SourcesDirectory)/test/Microsoft.DotNet.Tests/Microsoft.DotNet.Tests.csproj --restore
+
+      source ./eng/common/tools.sh
+      InitializeDotNetCli true
+
+      echo "##vso[task.setvariable variable=DotNetPath]$_InitializeDotNetCli"
```
akoeplinger: nit: you don't need this, you can just use `eng/common/dotnet.sh test` in the next step

ViktorHofer: Agreed. This step could be a one liner and the next step could also just use eng/common/dotnet.sh.

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,166 @@
...
+    displayName: Restore tests
+    workingDirectory: $(Build.SourcesDirectory)
+
+  - script: >
+      $(DotNetPath)/dotnet test
```
ViktorHofer: I think this will re-restore the project (unnecessary work). Not sure why you need the separate restore step above. I would either combine them or pass `--no-restore` in here.

eng/pipelines/templates/stages/reproducibility-tests.yml
```diff
@@ -0,0 +1,132 @@
...
+      artifactsRid: ${{ variables.centOSStreamX64Rid }}
+      targetArchitecture: x64
+      buildSourceOnly: true
+      enablePoison: false
+      buildFromArchive: false
+      runOnline: false
+      runTests: false
+      withPreviousSDK: false
+      createSourceArtifact: false
+      brandingType: ${{ variables.brandingType }}
```
omajid, line 43: Should we set [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) or [`/p:DeterminsticTimestamp=...`](https://github.com/NuGet/Home/blob/dev/accepted/2026/deterministic-pack-revisited.md) to rule out time-related differences and force both set of builds to the same wall-clock-time?

---

`/p:DeterministicTimestamp` uses `SOURCE_DATE_EPOCH` (if available) under the hood, so we don't need to set both.

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,166 @@
...
+- name: publishTestResultsPr
+  type: boolean
+  default: false
+
+- name: pullRequestTitle
+  type: string
+  default: 'Update Reproducibility Test Baselines and Exclusions'
+
+jobs:
+- job: Reproducibility_${{ parameters.buildType }}_${{ parameters.build1JobName }}_${{ parameters.architecture }}
```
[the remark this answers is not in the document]
mthalman: The names are within Azure DevOps limits and I want them to align with the naming conventions of the dotnet-unified-build pipeline.

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,166 @@
...
+  # Download the first SDK tarball from this pipeline's Build1 stage.
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      buildType: current
+      artifact: ${{ parameters.build1JobName }}_Artifacts
+      ${{ if eq(parameters.buildType, 'Msft') }}:
+        patterns: '**/assets/Release/Sdk/*/dotnet-sdk-+([0-9]).+([0-9]).+([0-9])?(-@(alpha|preview|rc|rtm)*)-${{ parameters.targetRid }}.tar.gz'
+      ${{ else }}:
+        patterns: '**/assets/Release/dotnet-sdk-+([0-9]).+([0-9]).+([0-9])?(-@(alpha|preview|rc|rtm)*)-${{ parameters.targetRid }}.tar.gz'
+      downloadPath: $(Pipeline.Workspace)/Artifacts
```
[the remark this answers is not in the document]
mthalman: This pattern is copied from sdk-diff-tests.yml and is empirically working — DownloadPipelineArtifact@2's minimatch supports the extglob syntax in practice.

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,166 @@
...
+    displayName: Move first build artifacts to root
+
+  - script: |
+      first_sdk_tarball=$(find "$(Pipeline.Workspace)/Artifacts" -name "dotnet-sdk-*-${{ parameters.targetRid }}.tar.gz" -exec basename {} \;)
+
+      if [[ -z "$first_sdk_tarball" ]]; then
+        echo "First SDK tarball not found in '$(Pipeline.Workspace)/Artifacts'. The Build1 stage's '${{ parameters.build1JobName }}' job might have failed to publish the expected artifact."
+        exit 1
+      fi
+
```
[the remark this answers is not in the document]
mthalman: This follows the same pattern as sdk-diff-tests.yml, and only one SDK tarball is produced per RID per build.

eng/pipelines/templates/jobs/reproducibility-test.yml
mthalman: Fixed in 9c3684e

eng/pipelines/templates/jobs/reproducibility-test.yml
mthalman: Fixed in 9c3684e

eng/pipelines/templates/stages/reproducibility-tests.yml
mthalman: We'll need to wait for https://github.com/NuGet/NuGet.Client/pull/7020 to flow into the VMR and be released in a new SDK version that the VMR references in order to make this possible.

eng/pipelines/unified-build-reproducibility-tests.yml
mthalman: The two triggers cover disjoint branch sets, mirroring the pattern in [source-build-sdk-diff-tests.yml](https://github.com/dotnet/dotnet/blob/main/eng/pipelines/source-build-sdk-diff-tests.yml):

- **Cron**: runs on `main` and pre-release feature-band branches (`release/*.0.4xx*`) on a daily cadence. These don't yet have `internal/release/*` mirrors so the pipeline-resource trigger wouldn't fire on them.
- **Pipeline-resource**: runs reactively on post-release branches (`release/*.0.1xx-preview*` and `internal/release/*`) every time `dotnet-unified-build` completes its build stages.

Per the comment in `source-build-sdk-diff-tests.yml`, the feature-band cron entries exist to cover the pre-release time period and can be removed once the band ships, since the pipeline-resource trigger handles the corresponding `internal/release/*` branches after release.

eng/pipelines/unified-build-reproducibility-tests.yml
mthalman: Fixed in a43b153

eng/pipelines/unified-build-reproducibility-tests.yml
mthalman: Fixed in a43b153

test/Microsoft.DotNet.Tests/Microsoft.DotNet.Tests.csproj
mthalman: Fixed in 9490784

test/Microsoft.DotNet.Reproducibility.Tests/ReproducibilityTests.cs
```diff
@@ -0,0 +1,246 @@
...
+            exclusionsFileName: "SourceBuiltReproducibilityExclusions.txt",
+            baselineFileName: "SourceBuiltReproducibilityBaseline.diff",
+            updatedBaselineFileName: "UpdatedSourceBuiltReproducibilityBaseline.diff");
+    }
+
+    /// <summary>
+    /// Compares two SDK tarballs and asserts they are identical, modulo files listed in
+    /// the exclusions file. Any new non-reproducible file causes the test to fail; any file
+    /// that becomes reproducible causes the baseline to be stale.
+    /// </summary>
```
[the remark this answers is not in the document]
mthalman: Fixed in 9f4203e

test/Microsoft.DotNet.Reproducibility.Tests/ReproducibilityTests.cs
```diff
@@ -0,0 +1,246 @@
...
+        Dictionary<string, string> hashes1 = null!;
+        Dictionary<string, string> hashes2 = null!;
+
+        Parallel.Invoke(
+            () => hashes1 = ComputeTarballFileHashes(tarballPath1),
+            () => hashes2 = ComputeTarballFileHashes(tarballPath2));
+
+        OutputHelper.WriteLine($"  Tarball 1 files: {hashes1.Count}");
+        OutputHelper.WriteLine($"  Tarball 2 files: {hashes2.Count}");
+
```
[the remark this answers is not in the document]
mthalman: Fixed in 08ec137

eng/pipelines/templates/stages/reproducibility-tests.yml
```diff
@@ -0,0 +1,132 @@
...
+stages:
+
+# Two parallel build stages, each producing a full set of SDK tarballs from the same
+# pipeline-resolved commit and the same pipeline-resolved container image digest.
+# Same-pipeline-run guarantees the same toolchain (clang/LLD, etc.) for both stages,
+# so any difference in the regular-file contents between their outputs is real
+# non-determinism, not toolchain drift.
+
+- stage: Build1
+  variables:
```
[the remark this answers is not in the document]
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/6622/commits/db1e1098a495af6021be6aa1d8c775ca40447f72

eng/pipelines/templates/stages/reproducibility-tests.yml
```diff
@@ -0,0 +1,132 @@
...
+      targetOS: linux
+      targetArchitecture: x64
+      runTests: false
+      brandingType: ${{ variables.brandingType }}
+
+# Compare stage: download both stages' SDK tarballs and run the reproducibility tests.
+- stage: Compare
+  dependsOn:
+  - Build1
+  - Build2
```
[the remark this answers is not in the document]
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/6622/commits/db1e1098a495af6021be6aa1d8c775ca40447f72

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,154 @@
...
+  steps:
+
+  # Download the first SDK tarball from this pipeline's Build1 stage.
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      buildType: current
+      artifact: ${{ parameters.build1JobName }}_Artifacts
+      ${{ if eq(parameters.buildType, 'Msft') }}:
+        patterns: '**/assets/Release/Sdk/*/dotnet-sdk-+([0-9]).+([0-9]).+([0-9])?(-@(alpha|preview|rc|rtm)*)-${{ parameters.targetRid }}.tar.gz'
+      ${{ else }}:
```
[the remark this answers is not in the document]
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/6622/commits/db1e1098a495af6021be6aa1d8c775ca40447f72

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,154 @@
...
+    displayName: Locate second SDK tarball
+
+  - script: >
+      eng/common/dotnet.sh test
+      $(Build.SourcesDirectory)/test/Microsoft.DotNet.Reproducibility.Tests/Microsoft.DotNet.Reproducibility.Tests.csproj
+      --logger:'trx;LogFileName=$(Agent.JobName)_ReproducibilityTests.trx'
+      --logger:'console;verbosity=detailed'
+      -c Release
+      -bl:$(Build.SourcesDirectory)/artifacts/log/Debug/ReproTests_$(date +"%m%d%H%M%S").binlog
+      -flp:LogFile=$(Build.SourcesDirectory)/artifacts/logs/ReproTests_$(date +"%m%d%H%M%S").log
```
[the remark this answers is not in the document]
mthalman: Fixed in 84af90f

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -183,9 +188,11 @@ jobs:
         image: ${{ parameters.container.image }}
         options: $(defaultContainerOptions)
 
-  ${{ if gt(length(parameters.reuseBuildArtifactsFrom), 0) }}:
+  ${{ if or(gt(length(parameters.reuseBuildArtifactsFrom), 0), gt(length(parameters.dependsOn), 0)) }}:
     dependsOn:
     - ${{ parameters.reuseBuildArtifactsFrom }}
+    - ${{ each dep in parameters.dependsOn }}:
+      - ${{ dep }}
```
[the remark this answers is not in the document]
mthalman: Fixed in 12f2a53

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,154 @@
...
+      echo "Second SDK tarball: $second_sdk_tarball"
+      echo "##vso[task.setvariable variable=SecondSdkTarballPath]$(Pipeline.Workspace)/RebuildArtifacts/$second_sdk_tarball"
+    displayName: Locate second SDK tarball
+
+  - script: >
+      eng/common/dotnet.sh test
+      $(Build.SourcesDirectory)/test/Microsoft.DotNet.Reproducibility.Tests/Microsoft.DotNet.Reproducibility.Tests.csproj
+      --logger:'trx;LogFileName=$(Agent.JobName)_ReproducibilityTests.trx'
+      --logger:'console;verbosity=detailed'
+      -c Release
```
ViktorHofer: That's the default in the VMR already.

eng/pipelines/templates/jobs/reproducibility-test.yml
```diff
@@ -0,0 +1,154 @@
...
+      -bl:$(Build.SourcesDirectory)/artifacts/log/Release/ReproTests_$(date +"%m%d%H%M%S").binlog
+      -flp:LogFile=$(Build.SourcesDirectory)/artifacts/log/ReproTests_$(date +"%m%d%H%M%S").log
+      -clp:v=m
```
ViktorHofer: Is it necessary to define that logger configuration in CI only? Wouldn't it be better to make them apply to local and CI tests? I would also remove /clp:v=m (default). The binlog parameter seems to be the only setting that should be CI only.

eng/pipelines/templates/stages/reproducibility-tests.yml
omajid: Do we have a tracking issue for that? Alternatively, we could just set these settings right now and benefit when the fixes flow in.

omajid (approved):

eng/pipelines/templates/stages/reproducibility-tests.yml
mthalman: Fixed in 2b8a7663355

eng/pipelines/templates/jobs/reproducibility-test.yml
mthalman: Fixed in 2cbe27dfd4b

eng/pipelines/templates/jobs/reproducibility-test.yml
mthalman: Fixed in 2cbe27dfd4b

test/Microsoft.DotNet.Reproducibility.Tests/ReproducibilityTests.cs
```diff
@@ -0,0 +1,254 @@
...
+            fileHashes[relativePath] = hash;
+        }
+
+        return fileHashes;
+    }
+
+    private static string ComputeHash(Stream? dataStream)
+    {
+        if (dataStream is null)
+            return string.Empty;
```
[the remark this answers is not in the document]
mthalman: null is a valid value for tar entries in some cases

eng/pipelines/templates/stages/reproducibility-tests.yml
```diff
@@ -0,0 +1,136 @@
+parameters:
+# This is not expected to be passed in but rather just an object variable reused below
+- name: pool_Linux
+  type: object
+  default:
+    name: NetCore1ESPool-Svc-Internal
```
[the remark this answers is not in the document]
mthalman: Fixed in https://github.com/dotnet/dotnet/pull/6622/commits/a28ef8baaae7ea3199a48d45660876c3e8f3984a

omajid:
One thing that we have seen in the past is that rebuilding dotnet on different days produces artifacts with different versions ([example from a while ago](https://github.com/dotnet/source-build/issues/2852)). Is it possible to move the clock forward by a few days between the two builds?

eng/pipelines/templates/jobs/vmr-build.yml
```diff
@@ -420,11 +437,24 @@ jobs:
         sbomEnabled: false
 
   steps:
+  - ${{ if eq(parameters.setDeterministicTimestamp, 'true') }}:
+    # Compute SourceDateEpoch (Unix seconds since epoch) from System.PipelineStartTime
+    # and export it via SOURCE_DATE_EPOCH on every build script invocation in this job
+    # so that every NuGet pack invocation stamps the same wall-clock time into its
+    # .nupkg ZIP envelope. System.PipelineStartTime is set once per pipeline run and
+    # is identical across all stages/jobs in the run, so two independent builds in
+    # the same run get matching package envelopes.
```
akoeplinger: > System.PipelineStartTime is set once per pipeline

this is actually not true, it is set to the job start time, see https://developercommunity.visualstudio.com/t/pipelinestarttime-is-not-consistent-in-a-pipeline/1558186

that's why I decided to base it off of Build.BuildNumber in https://github.com/dotnet/dotnet/pull/5256

mthalman: Well that's a horribly named variable.

mthalman: Fixed in 2daf5b8f61a — switched to Build.BuildNumber which is consistent across all stages/jobs in the same pipeline run.

mthalman:
> One thing that we have seen in the past is that rebuilding dotnet on different days produces artifacts with different versions ([example from a while ago](https://github.com/dotnet/source-build/issues/2852)). Is it possible to move the clock forward by a few days between the two builds?

I've added that here: https://github.com/dotnet/dotnet/pull/6622/commits/112678d5e6fa95ee9562d6e8a726f78031972a36. It didn't produce any additional diffs.

mthalman:
Ready for another review

akoeplinger (approved):

PR #6689, opened 2026-05-19
 Advance src/msbuild on release/10.0.4xx to msbuild/main HEAD

OvesN:
Solves problem with conflict resolving in this PR: https://github.com/dotnet/msbuild/pull/13791
 
 ## What happened
 
 After re-pointing backflow subscription `91fa59f1` (VMR → msbuild/main) from channel `.NET 11.0.1xx SDK` to `.NET 10.0.4xx SDK` on May 15, the backflow opened [dotnet/msbuild#13791](https://github.com/dotnet/msbuild/pull/13791) as an "unsafe self-corrective" PR — because VMR's `src/msbuild/` on `release/10.0.4xx` was still pinned at the May-4 snapshot (`87b0cc3`) while msbuild/main had advanced ~80 commits to `d7f2e70` (May 15).
 
 Forward-flow subscription `3fc72f68` (msbuild VS 18.8 → VMR release/10.0.4xx) has not opened a PR since https://github.com/dotnet/dotnet/pull/6415 merged on May 5, so the VMR side never caught up.
 
 ## What this PR does
 
 Runs `darc vmr reset msbuild:911bea0b57d3613eb9c29f49ff9858d03884c397` to realign `src/msbuild/` with the current msbuild/main HEAD.
 
 Verified no VMR-side edits to `src/msbuild/` since May 5:

gh api 'repos/dotnet/dotnet/commits?path=src/msbuild&sha=release/10.0.4xx&since=2026-05-05T15:06:18Z'

 (only result is the May-5 forward-flow merge itself), so this reset doesn't overwrite any unique VMR work.
 
 After merge, https://github.com/dotnet/msbuild/pull/13791 will be closed.


premun (approved):

premun:
The validation leg will go red but I will force-merge after.

ViktorHofer (commented):
Looks good. Question: How did we accumulate such a big delta? I would have assume that release/10.0.4xx and main in the VMR should be on msbuild's main HEAD.

JanProvaznik:
> Looks good. Question: How did we accumulate such a big delta? I would have assume that release/10.0.4xx and main in the VMR should be on msbuild's main HEAD.

subscriptions were misconfigured such that VMR was taking msbuild VS 18.7 when main was going to VS 18.8 for a while

ViktorHofer:
I have seen `darc vmr reset`s like every other month so it seems like we are doing something wrong here and we need to update our process. Would you agree with that? This seems to be a perfect example for some form of automation.

premun:
FYI - if you'd just retarget subscriptions and start flowing from a different branch, Maestro can now detect this and issue a PR that tries to fix this. It's not a reset but an unsafe flow which gives a bit better results.

We are considering adding a `darc` command which would create a "reset PR" too so that it's a bit faster to do resets than manually.

JanProvaznik:
> I have seen `darc vmr reset`s like every other month so it seems like we are doing something wrong here and we need to update our process. Would you agree with that? This seems to be a perfect example for some form of automation.

1 of the process problems is that the vmr release/10.0.4xx branch starts existing manually and we have to manually go and verify our flows make sense in that world. + there are special cases for preview channels/branches that I don't think we can script
2 is that the double flow `main<->main`, `main<->prereleased10.0.x00` we thought was supported is not supported so we ran into that 3x this year

@OvesN do you have the diagram for expected MSBuild codeflow from our session, will you publish it into our docs?
I think we can script the happy path for "do all darc updates at the end of the month". But after this we still need to regenerate the diagram and inspect "does everything make sense?"

premun:
1. When the VMR snaps the 4xx branch, it's a centralized process that is tied to a tool that also opens a PR against the Maestro configuration repo to remap subscriptions. So if there would be an action to perform for MSBuild, it could easily happen in a coordinated fashion.

OvesN:
@premun Can this Pr be merged?

PR #6697, opened 2026-05-19
Fix intermittent circular dependency in sharedfx.targets (MSB4006)

wtgodbe:
## Summary

Fixes an intermittent `MSB4006: There is a circular dependency in the target dependency graph` build failure in shared framework projects, and fixes a latent platform manifest bug exposed by the fix.

## Problem 1: Circular dependency (MSB4006)

`sharedfx.targets` contains self-referencing `<MSBuild Projects="$(MSBuildProjectFullPath)">` calls in `_AddFilesToNuGetPackage` and `GetFilesToPublish`. When these nested calls produce **identical global properties** to the parent build, the MSBuild scheduler treats them as the same build and detects a false circular dependency.

`_AddFilesToNuGetPackage` (line 540) uses `RemoveProperties="NoBuild"`, but when `NoBuild` isn't set in the outer build, removing it is a no-op — the nested call has identical global properties to the parent.

### Fix

Add differentiating dummy properties (`_GettingFilesForPackage=true`, `_GettingFilesForPublish=true`) to the nested MSBuild calls so they always have distinct global property sets from the parent build.

This is the same pattern used in the workaround at dotnet/aspnetcore#65667 (which added `_PublishingToLayout=true` to the caller side for `GetFilesToPublish`). This PR fixes both targets at the source in arcade and removes the aspnetcore workaround.

### Evidence

- **Failing build**: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1426800 (aspnetcore CI, Linux x64)
- **Error**: `sharedfx.targets(540,5): error MSB4006: There is a circular dependency in the target dependency graph involving target "_AddFilesToNuGetPackage"`
- **MSBuild issue**: https://github.com/dotnet/msbuild/issues/13599

## Problem 2: Missing platform manifest entry (exposed by fix)

With the circular dependency fix, the nested MSBuild call in `_AddFilesToNuGetPackage` now correctly runs as a **distinct build** instead of reusing cached results from the parent. This causes `_ValidateAllFilesInTemplatedPlatformManifest` to evaluate freshly and catch a pre-existing bug: `Microsoft.NETCore.App.Runtime.CoreCLR.r2r.dylib` (the composite R2R Mach-O image produced by `Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj` on Apple mobile platforms) was missing from the `PlatformManifestFileEntry` list.

The manifest already had `Microsoft.NETCore.App.r2r.dylib` (from the Mono sfxproj), but not the CoreCLR variant. This bug was latent — previously hidden because MSBuild reused the parent build's already-evaluated targets, skipping fresh validation.

### Fix

Add `Microsoft.NETCore.App.Runtime.CoreCLR.r2r.dylib` to the `PlatformManifestFileEntry` list in `src/runtime/src/installer/pkg/sfx/Microsoft.NETCore.App/Directory.Build.props`.

## Changes

| Commit | Component | Change |
|--------|-----------|--------|
| 1 | src/arcade | Add `_GettingFilesForPackage` and `_GettingFilesForPublish` differentiating properties to nested MSBuild calls in sharedfx.targets |
| 2 | src/aspnetcore | Remove `_PublishingToLayout` workaround from sfxproj (no longer needed) |
| 3 | src/runtime | Add missing `Microsoft.NETCore.App.Runtime.CoreCLR.r2r.dylib` platform manifest entry |

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
+++ b/src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
@@ -537,8 +537,12 @@
 
   <Target Name="_AddFilesToNuGetPackage" BeforeTargets="_GetPackageFiles">
 
+    <!-- Set _GettingFilesForPackage=true to differentiate global properties for this nested MSBuild call
+         and avoid an intermittent circular dependency (MSB4006) when the scheduler builds the same project
+         concurrently with identical global properties. See https://github.com/dotnet/msbuild/issues/13599 -->
     <MSBuild Projects="$(MSBuildProjectFullPath)"
           Targets="_GetAllSharedFrameworkFiles"
+          Properties="_GettingFilesForPackage=true"
           RemoveProperties="NoBuild">
       <Output TaskParameter="TargetOutputs" ItemName="_FilesToPackage" />
     </MSBuild>
@@ -570,8 +574,11 @@
 
   <Target Name="GetFilesToPublish">
 
+    <!-- Set _GettingFilesForPublish=true to differentiate global properties for this nested MSBuild call
+         and avoid an intermittent circular dependency (MSB4006). -->
     <MSBuild Projects="$(MSBuildProjectFullPath)"
           Targets="_GetAllSharedFrameworkFiles"
+          Properties="_GettingFilesForPublish=true"
           RemoveProperties="OutputPath;SymbolsOutputPath">
       <Output TaskParameter="TargetOutputs" ItemName="_FilesToPackage" />
     </MSBuild>
--- a/src/aspnetcore/src/Framework/App.Runtime/src/Microsoft.AspNetCore.App.Runtime.sfxproj
+++ b/src/aspnetcore/src/Framework/App.Runtime/src/Microsoft.AspNetCore.App.Runtime.sfxproj
@@ -136,10 +136,9 @@
   </Target>
 
   <Target Name="PublishToSharedLayoutRoot" BeforeTargets="Build">
-    <!-- Set _PublishingToLayout=true to differentiate global properties for this nested MSBuild call and avoid an intermittent circular dependency involving GetFilesToPublish. -->
     <MSBuild Projects="$(MSBuildProjectFullPath)"
              Targets="PublishToDisk"
-             Properties="OutputPath=$(SharedFrameworkLayoutRoot);_PublishingToLayout=true" />
+             Properties="OutputPath=$(SharedFrameworkLayoutRoot)" />
   </Target>
 
   <Target Name="ReturnProductVersion" Returns="$(Version)" />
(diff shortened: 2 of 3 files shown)


lewing (approved):

ViktorHofer (approved):

ViktorHofer:
cc @jkoritzinsky

wtgodbe:
Needs a fix:

> Microsoft.NETCore.App.Runtime.CoreCLR.r2r.dylib. Add these file names with extensions to the 'PlatformManifestFileEntry' item group for the runtime pack and corresponding ref pack to include them in the platform manifest.
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets(303,5): error : The following files are missing entries in the templated manifest: [/Users/runner/work/1/s/src/runtime/src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj]
  /Users/runner/work/1/s/artifacts/source-built-sdks/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets(303,5): error : Microsoft.NETCore.App.Runtime.CoreCLR.r2r.dylib. Add these file names with extensions to the 'PlatformManifestFileEntry' item group for the runtime pack and corresponding ref pack to include them in the platform manifest. [/Users/runner/work/1/s/src/runtime/src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj]

Looking

wtgodbe:
Pushed https://github.com/dotnet/dotnet/pull/6697/commits/7080fe50632c82ef9a16982590e1e6fa457bb7ac - CC @radekdoulik, @hoyosjs - see the "Problem 2" section in the PR description

akoeplinger (approved):

omajid:
I think this PR has introduced a new problem.

When building the VMR, I get:

```

  dotnet/artifacts/source-built-sdks/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets(198,5): error : The process cannot access the file 'dotnet/src/runtime/artifacts/obj/coreclr-pack/Release/net11.0/fedora.43-x64/Microsoft.NETCore.App.deps.json' because it is being used by another process. [dotnet/src/runtime/src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj]
  dotnet/src/runtime/src/installer/pkg/sfx/bundle/Microsoft.NETCore.App.Bundle.bundleproj(44,5): error MSB4181: The "MSBuild" task returned false but did not log an error.
```

Claude says:

  ---                                                                                                                                                         
###  Before the PR   
                                                                                                                                                              
  Both nested <MSBuild> calls had identical global properties to their parent. MSBuild's scheduler sees two build requests for the same project with the same  global properties and treats them as the same build instance — it deduplicates them, serializes them, or serves a cached result. Only one instance of `_GenerateSharedFrameworkDepsFile` ever runs. No file contention.
                                                                                                                                                              
  The downside: the deduplication/caching interaction occasionally triggered MSB4006 (false-positive "circular dependency" from MSBuild's scheduler when two nodes raced to claim the same project). This was the bug the PR was fixing. It was intermittent.
                                                                                                                                                              
###  After the PR    

  The two nested calls now have different global properties. MSBuild's scheduler treats them as entirely separate project instances — they get separate cache entries, and can be scheduled on separate worker nodes concurrently. Both execute `_GenerateSharedFrameworkDepsFile` → `File.Create(depsFilePath)` with `FileShare.None` on the same path → "file in use by another process".                                                                                         
                  
  The MSB4006 is gone. In its place is a file-lock crash. This failure is more deterministic — it happens whenever the scheduler places the two instances on  concurrent nodes, which at `-maxcpucount` (the default) on a multi-core machine happens reliably.

PR #6718, opened 2026-05-20
Move nuget and msbuild forward to latest 18.0 versions

marcpopMSFT:


lewing (approved):

marcpopMSFT:
This fix doesn't work as that msbuild has a transitive dependency that'll be flagged by scanners. I'll try a more targeted CPM fix.

joeloff (approved):

wtgodbe (approved):

PR #6352, opened 2026-04-28
Modify BuildArgs to disable NgenOptimization

rainersigwald:
Updated BuildArgs to disable NgenOptimization unconditionally.

PR #6719, opened 2026-05-20
Disable IBC by default for release/10.0.1xx

akoeplinger:
https://github.com/dotnet/dotnet/pull/6352 only disabled it for msbuild but we need the same for roslyn so disable it globally.


vseanreesermsft (approved):

akoeplinger:
fast merge, this is not tested in public PRs so we don't need to waste CI resources.

PR #6720, opened 2026-05-20
Make bundled tools from dotnet/aspnetcore native AOT

DamianEdwards:
This updates the tools bundled into the SDK from dotnet/aspnetcore to be published native AOT. Adds coarse-grained smoke tests that exercise the main functionality of the tools after they've been published.

Tools effected:
- `dotnet user-secrets`
- `dotnet user-jwts`
- `dotnet dev-certs`

The published size of each tool locally on `win-x64` is:

Tool | Published size
-- | --
dotnet-user-secrets | 3.94 MB
dotnet-user-jwts | 5.36 MB
dotnet-dev-certs | 3.04 MB


src/sdk/src/Layout/redist/targets/BundledDotnetTools.targets
```diff
@@ -1,25 +1,69 @@
...
+    <DotnetUserSecretsPackageRid Condition="'$(DotnetUserSecretsPackageRid)' == '' and '$(DotNetBuildSourceOnly)' != 'true' and $(ProductMonikerRid.StartsWith('win')) and '$(TargetArchitecture)' == 'x86'">win-x86</DotnetUserSecretsPackageRid>
+    <DotnetUserSecretsPackageRid Condition="'$(DotnetUserSecretsPackageRid)' == '' and '$(DotNetBuildSourceOnly)' != 'true' and $(ProductMonikerRid.StartsWith('win')) and '$(TargetArchitecture)' == 'arm64'">win-arm64</DotnetUserSecretsPackageRid>
+    <DotnetUserSecretsPackageRid Condition="'$(DotnetUserSecretsPackageRid)' == '' and '$(DotNetBuildSourceOnly)' != 'true' and $(ProductMonikerRid.StartsWith('osx')) and '$(TargetArchitecture)' == 'x64'">osx-x64</DotnetUserSecretsPackageRid>
+    <DotnetUserSecretsPackageRid Condition="'$(DotnetUserSecretsPackageRid)' == '' and '$(DotNetBuildSourceOnly)' != 'true' and $(ProductMonikerRid.StartsWith('osx')) and '$(TargetArchitecture)' == 'arm64'">osx-arm64</DotnetUserSecretsPackageRid>
+  </PropertyGroup>
```
baronfel, line 7: This looks sketchy - we should already have a 'known SDK target RID' that could be used here instead of this logic.

src/sdk/src/Layout/redist/targets/BundledDotnetTools.targets
```diff
@@ -1,25 +1,69 @@
...
       <RestoredPackagePath>$(NuGetPackageRoot)$([MSBuild]::ValueOrDefault('%(PackagePathRelativeToPackageRoot)', '').ToLowerInvariant())/</RestoredPackagePath>
-      <NupkgPathRelativeToPackageRoot>%(Identity)/%(Version)/%(Identity).%(Version).nupkg</NupkgPathRelativeToPackageRoot>
       <RestoredNupkgPath>$(NuGetPackageRoot)$([MSBuild]::ValueOrDefault('%(NupkgPathRelativeToPackageRoot)', '').ToLowerInvariant())</RestoredNupkgPath>
     </BundledDotnetTool>
   </ItemGroup>
 
   <ItemGroup>
-    <PackageDownload Include="@(BundledDotnetTool)" Version="[%(Version)]" />
+    <PackageDownload Include="%(BundledDotnetTool.PackageIdToRestore)" Version="[%(BundledDotnetTool.Version)]" />
   </ItemGroup>
```
baronfel: We should move computation of the PackageDownloads to a Target, so that we can use the SelectRuntimeIdentifierSpecificItems Task to choose for us instead of doing any hard-coding.

baronfel (commented):
Some thoughts on how we should do this better

src/aspnetcore/src/Tools/Directory.Build.props
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <ToolSharedSourceRoot>$(MSBuildThisFileDirectory)Shared\</ToolSharedSourceRoot>
     <Nullable>disable</Nullable>
+    <BundledToolTargetRuntimeIdentifiers>linux-x64;linux-arm;linux-arm64;win-x64;win-x86;win-arm64;osx-x64;osx-arm64</BundledToolTargetRuntimeIdentifiers>
```
DamianEdwards: Would be nice if there were a known property for the set of RIDs we need to target to match what we ship the SDK for. It's possible this list is missing some, e.g. I think @MichalStrehovsky said we should have the linux musl variants here too?

baronfel: Start with linux-musl-x64 and linux-musl-arm64 IMO.

MichalStrehovsky: I used the list of downloads from: https://dotnet.microsoft.com/en-us/download/dotnet/10.0. we also have arm32 musl.

src/aspnetcore/src/Tools/NativeAot/test/NativeAotToolScenarioTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        var userSecretsProperty = string.IsNullOrEmpty(userSecretsId) ? string.Empty : $"<UserSecretsId>{userSecretsId}</UserSecretsId>";
+        var projectPath = Path.Combine(projectDirectory, "TestProject.csproj");
+
+        File.WriteAllText(
+            projectPath,
+            $$"""
+            <Project Sdk="Microsoft.NET.Sdk">
+              <PropertyGroup>
+                <OutputType>Exe</OutputType>
+                <TargetFramework>net11.0</TargetFramework>
```
BrennanConroy: Can we avoid hardcoding 11?

src/aspnetcore/src/Tools/Shared/SecretsHelpers/ProjectIdResolver.cs
```diff
@@ -130,16 +131,26 @@ public string Resolve(string project, string configuration)
...
+        if (!string.IsNullOrEmpty(assemblyLocation))
         {
-                Path.Combine(AppContext.BaseDirectory, "assets"),
-                Path.Combine(assemblyDir, "assets"),
-                AppContext.BaseDirectory,
-                assemblyDir,
-            };
+            var assemblyDir = Path.GetDirectoryName(assemblyLocation);
+            // Preserve search order from previous versions before the check of the assembly location value was added.
+            searchPaths.Insert(1, Path.Combine(Path.GetDirectoryName(assemblyDir), "assets"));
```
BrennanConroy: Looks like this accidentally added a second `GetDirectoryName` call. I assume it wasn't on purpose.

```suggestion
            searchPaths.Insert(1, Path.Combine(assemblyDir, "assets"));
```

src/aspnetcore/src/Tools/Shared/TestHelpers/NativeAotToolRunner.cs
```diff
@@ -0,0 +1,173 @@
...
+    public static async Task<NativeAotToolResult> RunAsync(
+        string toolName,
+        IEnumerable<string> arguments,
+        ITestOutputHelper output,
+        string? workingDirectory = null,
+        IReadOnlyDictionary<string, string>? environmentVariables = null,
+        TimeSpan? timeout = null)
+    {
+        var toolPath = GetPublishedToolPath(toolName);
+        var startInfo = new ProcessStartInfo
```
BrennanConroy, line 31: Not necessary, but can this make use of the new Process APIs?
https://devblogs.microsoft.com/dotnet/process-api-improvements-in-dotnet-11/

src/aspnetcore/src/Tools/Directory.Build.targets
```diff
@@ -11,6 +11,31 @@
     <TargetLatestDotNetRuntime Condition=" '$(IsServicingBuild)' == 'true' ">false</TargetLatestDotNetRuntime>
     <!-- Tool projects publish before packing. Packages should contain the latest bits. -->
     <UseLatestPackageReferences>true</UseLatestPackageReferences>
+
+    <!-- AOT publish is not supported in source-only builds, so disable it to avoid build errors. -->
+    <PublishAot Condition=" '$(PublishAot)' == 'true' AND '$(DotNetBuildSourceOnly)' == 'true'">false</PublishAot>
```
BrennanConroy: nit: 
```suggestion
    <PublishAot Condition=" '$(DotNetBuildSourceOnly)' == 'true'">false</PublishAot>
```

src/sdk/src/Cli/dotnet/CommandFactory/CommandResolution/DotnetToolsCommandResolver.cs
```diff
@@ -26,10 +31,25 @@ public DotnetToolsCommandResolver(string? dotnetToolPath = null)
...
         var version = packageId.GetDirectories()[0];
-        var dll = version.GetDirectories("tools")[0]
+        var toolDirectory = version.GetDirectories("tools")[0]
             .GetDirectories()[0] // TFM
-            .GetDirectories()[0] // RID
-            .GetFiles($"{arguments.CommandName}.dll")[0];
+            .GetDirectories()[0]; // RID
+
+        var toolSettingsFile = Path.Combine(toolDirectory.FullName, ToolSettingsFileName);
+        if (File.Exists(toolSettingsFile))
```
BrennanConroy: Someone more familiar with this area should review this

BrennanConroy (approved):
Generally good, some other folks should look closer at the msbuild and tool invocation change.

src/aspnetcore/src/Tools/Shared/TestHelpers/NativeAotToolRunner.cs
DamianEdwards: Yeah I thought that and then decided it was a perfect candidate for a follow up :smile:

src/aspnetcore/src/Tools/Shared/SecretsHelpers/ProjectIdResolver.cs
DamianEdwards: Ah seems the copilot review found that too

src/aspnetcore/src/Tools/Directory.Build.targets
DamianEdwards: Updated

src/aspnetcore/src/Tools/Shared/SecretsHelpers/ProjectIdResolver.cs
DamianEdwards: Updated

src/sdk/src/Cli/dotnet/CommandFactory/CommandResolution/DotnetToolsCommandResolver.cs
DamianEdwards: @baronfel

baronfel: I think this isn't the right place. I think we have a different resolver for the built-in tools, this one is for local tools more generally. Can check later tonight or tomorrow.

DamianEdwards: OK agent is looking at this again

DamianEdwards: I checked this path. `DotnetToolsCommandResolver` is the resolver for SDK-bundled tools under `sdk/<version>/DotnetTools`: `Program.ExecuteExternalCommand` prefixes unknown commands to `dotnet-*`, then `DefaultCommandResolverPolicy` tries `DotnetToolsCommandResolver` before `LocalToolsCommandResolver`. `LocalToolsCommandResolver` is the manifest/cache-based resolver for repo/local tools.

So I think this is the right place for bundled `dotnet user-secrets`/`dotnet dev-certs`/`dotnet user-jwts` execution. The built SDK smoke path also goes through this layout and `dotnet user-secrets --help` resolves from the SDK `DotnetTools` directory with the native executable layout.

DamianEdwards: @baronfel who should review this from the SDK side?

baronfel: @dsplaisted or @joeloff would be my go-tos for toll resolution

DamianEdwards: FYI I downloaded the CI build and verified that the tools are laid out in the SDK hive as expected and that the tool resolution works:

image

```
> C:\Users\dedward\Downloads\dotnet-sdk-11.0.100-ci-win-x64\dotnet.exe user-secrets list --file .\usersecrets.cs
MySecret = This is a secret!
```

DamianEdwards: Also verified that dispatch to the other bundled tools, e.g. `dotnet format`, `dotnet watch`, still works

baronfel (approved):
The tool resolver changes are good on the SDK side of things, thanks for plumbing this through!

src/sdk/src/Layout/redist/targets/BundledDotnetTools.targets
```diff
@@ -1,28 +1,108 @@
 <Project>
 
+  <UsingTask TaskName="Microsoft.NET.Build.Tasks.SelectRuntimeIdentifierSpecificItems" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
+
+  <PropertyGroup>
+    <AspNetCoreRidSpecificToolPackageRids>linux-x64;linux-arm;linux-arm64;linux-musl-x64;linux-musl-arm;linux-musl-arm64;win-x64;win-x86;win-arm64;osx-x64;osx-arm64</AspNetCoreRidSpecificToolPackageRids>
```
am11, line 6: When we are building .NET for community platforms like linux-riscv64 or freebsd-arm64 (currently, there are nine of them), it should account for it. With the same token, those platforms tools aren't required in MSFT SDK.

IOW, we are missing "current target RID"  in the list.

DamianEdwards: Opened follow-up PR: #6797

lewing:
## 🚨 NativeAOT linker failure on musl/cross-build legs in VMR rolling builds

Since this PR merged, **all internal `main` rolling builds are failing** on 6 Linux legs with:

```
clang: error: invalid linker name in argument '-fuse-ld=bfd'
  [src/aspnetcore/src/Tools/dotnet-user-secrets/src/dotnet-user-secrets.csproj]

microsoft.dotnet.ilcompiler/11.0.0-preview.5.26272.119/build/Microsoft.NETCore.Native.targets(370,5):
  error MSB3073: The command "clang" ... -fuse-ld=bfd ... --target=aarch64-linux-musl ... exited with code 1.
```

### Affected legs

Linux_Alpine_arm64, Linux_Alpine_x64, Linux_arm64, Linux_arm, Linux_Pgo_x64, Linux_Pgo_arm64 — all cross-compilation or musl targets.

### Bisect

Last green: build 2981555 (sha `6fa530d`, queued May 22 07:06 UTC)
First red: build 2982298 (sha `821c533`, queued May 23 02:13 UTC) — first commit is this PR.

### Cause

The NativeAOT ILCompiler passes `-fuse-ld=bfd` (GNU linker) when linking `dotnet-user-secrets` for `linux-musl-arm64`. But the VMR cross-compilation containers (AzureLinux) and Alpine containers don't have `ld.bfd` installed — they ship `ld.lld` (LLVM) or `ld.mold`.

### Possible fixes

1. **Runtime**: Change NativeAOT linker selection to prefer `lld` over `bfd` on musl targets (runtime team, `Microsoft.NETCore.Native.targets`)
2. **Infra**: Install `binutils` (provides `ld.bfd`) in the cross-compilation container images (`@dotnet/dnceng`)
3. **Revert** this PR until the linker issue is resolved

cc @DamianEdwards @MichalStrehovsky @am11

am11:
> **Runtime**: Change NativeAOT linker selection to prefer `lld` over `bfd` on musl targets (runtime team, `Microsoft.NETCore.Native.targets`)

This is our go-to way for official builds. You can just copy this target and it should work https://github.com/dotnet/runtime/blob/2dce83ad91efd1efbf974eb636a18f9b0fcdcb17/eng/toolAot.targets#L28C1-L48C12 (uses a script from arcade's eng/common so it's available everywhere)

PS -
lld is LLVM's linker, the default linker on linux is bfd. bfd comes from GNU binutils package, that is target specific (e.g. to build arm64 on x64, we need cross-arm64 binutils). lld, OTOH, is multi-target, so you can even link linux-riscv64 binary on win-x64 using lld.exe on windows and it would work.

lewing:
### Additional details on the `-fuse-ld=bfd` root cause

The NativeAOT linker flavor selection lives in `src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets`:

```xml
<LinkerFlavor Condition="... and '$(_targetOS)' == 'freebsd'">lld</LinkerFlavor>
<LinkerFlavor Condition="... and '$(_linuxLibcFlavor)' == 'bionic'">lld</LinkerFlavor>
<LinkerFlavor Condition="... and '$(_targetOS)' == 'android'">lld</LinkerFlavor>
<LinkerFlavor Condition="... and '$(_targetOS)' == 'linux'">bfd</LinkerFlavor>  ← catches all linux including musl
```

**There is no musl-specific override.** All `linux-musl-*` targets fall through to the generic `linux` → `bfd` default, but the VMR cross-compilation containers (AzureLinux) don't have `ld.bfd` installed.

The runtime's own cross-build docs (`src/coreclr/nativeaot/docs/containers.md`) document the workaround: pass `-p:LinkerFlavor=lld`. The runtime's own `eng/toolAot.targets` also auto-detects LLD when available.

The fix would be adding a musl override in `Microsoft.NETCore.Native.Unix.targets`:
```xml
<LinkerFlavor Condition="'$(LinkerFlavor)' == '' and '$(_linuxLibcFlavor)' == 'musl'">lld</LinkerFlavor>
```

**Short-term workaround** for this PR: set `<LinkerFlavor>lld</LinkerFlavor>` in the aspnetcore tool csproj files for musl RIDs, or pass `/p:LinkerFlavor=lld` on cross-build legs.

Note: this PR was authored directly in `dotnet/dotnet` (not flowed from aspnetcore/runtime), so it was only tested by VMR PR validation — which does not include any Linux cross-build or musl legs. That's the coverage gap that allowed this to slip through.

PR #6732, opened 2026-05-20
Update just the Cryptography.xml transitive dependency

marcpopMSFT:
MSbuild transitively pulls in 8.0.0. Updating to 8.0.3 as the least significant change but open to going to a higher version.


mthalman (approved):

joeloff (approved):

marcpopMSFT:
Interesting, BuildComparer ius using a 9 RC2 version and changevalidation is on 10.0.0 from darclib. This time I'll just lift to the latest 10 version as I started with trying the least impactful version update.

marcpopMSFT:
Confirmed that this worked. We can still decide to also take the nuget/msbuild change but we need this for sure.

marcpopMSFT:
Still TBD is why the PR builds don't flag the nuget issue but the CI builds do

wtgodbe (approved):

PR #6740, opened 2026-05-21
Fix deps.json file race in sharedfx.targets using isolated output directory

JanProvaznik:
## Summary

Fix file race in `sharedfx.targets` between Pack and Publish paths that write `deps.json` to the same `IntermediateOutputPath`.

## Root Cause

Two MSBuild self-calls in `sharedfx.targets` trigger `_GenerateSharedFrameworkDepsFile`:
1. **Pack path** (`_AddFilesToNuGetPackage`): `RemoveProperties="NoBuild"`
2. **Publish path** (`GetFilesToPublish`): `RemoveProperties="OutputPath;SymbolsOutputPath"`

Different `RemoveProperties` = different global property dictionaries = different MSBuild configurations = **no deduplication** = both run in parallel and write `.deps.json` to the same directory.

## Fix

Pass a custom property `_DepsFileIntermediateDir` (not `IntermediateOutputPath`) from the Publish self-call. The `_GenerateSharedFrameworkDepsFile` target uses this property (with fallback to `IntermediateOutputPath`) to write deps.json to a unique subdirectory.

**Why not IntermediateOutputPath directly?** Setting `IntermediateOutputPath` as a global property on the MSBuild task propagates it to ALL P2P references (`ResolveReferences` is in the dependency chain). This caused `ILLink.Tasks.csproj` to build with the wrong intermediate path and fail when copying its output to a location locked by other nodes.

## Impact

- Fixes `deps.json` file race that caused `Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj` to fail
- No propagation to P2P references (fixes `ILLink.Tasks.dll` copy failure on Windows)
- No change to build ordering or parallelism semantics

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
+++ b/src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
@@ -572,7 +572,8 @@
 
     <MSBuild Projects="$(MSBuildProjectFullPath)"
           Targets="_GetAllSharedFrameworkFiles"
-          RemoveProperties="OutputPath;SymbolsOutputPath">
+          RemoveProperties="OutputPath;SymbolsOutputPath"
+          Properties="IntermediateOutputPath=$(IntermediateOutputPath)publish/">
       <Output TaskParameter="TargetOutputs" ItemName="_FilesToPackage" />
     </MSBuild>
 
--- a/src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
+++ b/src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
@@ -185,9 +185,11 @@
           DependsOnTargets="GetFilesToPackage"
           Condition="'$(PlatformPackageType)' == 'RuntimePack'">
     <PropertyGroup>
+      <_DepsFileOutputDir>$([MSBuild]::ValueOrDefault('$(_DepsFileIntermediateDir)', '$(IntermediateOutputPath)'))</_DepsFileOutputDir>
       <_TargetRootForFilesForDepsFile Condition="'$(HostJsonTargetPath)' == '$(_DefaultHostJsonTargetPath)'">runtimes/$(RuntimeIdentifier)</_TargetRootForFilesForDepsFile>
       <_TargetRootForFilesForDepsFile Condition="'$(HostJsonTargetPath)' != '$(_DefaultHostJsonTargetPath)'">$(HostJsonTargetPath)</_TargetRootForFilesForDepsFile>
     </PropertyGroup>
+    <MakeDir Directories="$(_DepsFileOutputDir)" Condition="'$(_DepsFileIntermediateDir)' != ''" />
     <ItemGroup>
       <_FilesForDepsFile Include="@(FilesToPackage)"
                          Condition="'%(FilesToPackage.IsSymbolFile)' != 'true' and
@@ -201,7 +203,7 @@
                                      SharedFrameworkPackName="$(PackageId)"
                                      Version="$(Version)"
                                      Files="@(_FilesForDepsFile)"
-                                     IntermediateOutputPath="$(IntermediateOutputPath)"
+                                     IntermediateOutputPath="$(_DepsFileOutputDir)"
                                      SharedFrameworkDepsNameOverride="$(SharedFrameworkHostFileNameOverride)"
                                      RuntimeIdentifierGraph="$([MSBuild]::ValueOrDefault('$(RuntimeIdentifierGraphPath)', '$(BundledRuntimeIdentifierGraphFile)'))"
                                      IncludeFallbacksInDepsFile="$(IncludeFallbacksInDepsFile)">
@@ -573,7 +575,7 @@
     <MSBuild Projects="$(MSBuildProjectFullPath)"
           Targets="_GetAllSharedFrameworkFiles"
           RemoveProperties="OutputPath;SymbolsOutputPath"
-          Properties="IntermediateOutputPath=$(IntermediateOutputPath)publish/">
+          Properties="_DepsFileIntermediateDir=$(IntermediateOutputPath)publish/">
       <Output TaskParameter="TargetOutputs" ItemName="_FilesToPackage" />
     </MSBuild>
 


src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/installer.singlerid.targets
rainersigwald, line 40: I'm pretty confused by this fix proposal--in the binlog from https://dev.azure.com/dnceng-public/public/_build/results?buildId=1428236&view=artifacts&pathAsName=false&type=publishedArtifacts I don't see the `GenerateInstallers` target called, nor any import of this file.

JanProvaznik:
You're correct — I analyzed the binlog from build 1428236 (the original failing CentOS source-build) and confirmed:

1. **No \\GenerateInstallers\\ target** is called in that build
2. **No \\installer.singlerid.targets\\ import** is present
3. **\\_GenerateSharedFrameworkDepsFile\\ IS called** 3 times on CoreCLR.sfxproj (twice succeeds, third fails with IOException)

**The actual race in source-build is at the traversal level:**

- \\Build.proj\\ (Traversal SDK) builds all \\@(ProjectToBuild)\\ in parallel, which includes both \\CoreCLR.sfxproj\\ and \\Bundle.bundleproj\\ (via \\Subsets.props\\ lines 738, 743)
- CoreCLR.sfxproj's own build calls \\_AddFilesToNuGetPackage\\ → nested MSBuild(self, \\RemoveProperties=""NoBuild""\\) → \\_GenerateSharedFrameworkDepsFile\\
- Bundle.bundleproj calls CoreCLR.sfxproj with \\GetFilesToPublish\\ → nested MSBuild(self, \\RemoveProperties=""OutputPath;SymbolsOutputPath""\\) → \\_GenerateSharedFrameworkDepsFile\\
- These nested calls have **different global property sets** (different properties removed), so MSBuild treats them as distinct instances and doesn't serialize them
- Both write to the same \\deps.json\\ path via \\File.Create()\\ with exclusive lock → IOException

**The fix has two parts:**

1. \\BuildInParallel=""false""\\ in \\installer.singlerid.targets\\ — needed for non-source-build scenarios where \\GenerateInstallers\\ IS called (Windows/macOS installer generation). The CI 7/8 green confirmed this works.

2. **New commit**: Make \\GenerateSharedFrameworkDepsFile\\ task resilient to concurrent writes — write to temp file, atomic move, retry on IOException. Since all invocations produce identical output (same inputs, deterministic), reusing another writer's result is safe.

The second fix addresses the source-build scenario you identified.

JanProvaznik (commented):
no this is wrong, the build definition should be right, task should not be edited

src/arcade/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets
```diff
@@ -201,7 +203,7 @@
                                      SharedFrameworkPackName="$(PackageId)"
                                      Version="$(Version)"
                                      Files="@(_FilesForDepsFile)"
-                                     IntermediateOutputPath="$(IntermediateOutputPath)"
+                                     IntermediateOutputPath="$(_DepsFileOutputDir)"
```
JanProvaznik, line 206: @rainersigwald The error in the original failing build (1428236) directly references this file:

```
artifacts/source-built-sdks/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets(198,5): error : The process cannot access the file '...Microsoft.NETCore.App.deps.json' because it is being used by another process.
```

The file is imported via `<Sdk Name="Microsoft.DotNet.SharedFramework.Sdk" />` in `Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj`. The race occurs between two self-calls on the sfxproj:

1. **Pack path**: `_AddFilesToNuGetPackage` (line 542) → self-call with `RemoveProperties="NoBuild"` → triggers `_GenerateSharedFrameworkDepsFile` 
2. **Publish path**: `GetFilesToPublish` (line 575) → self-call with `RemoveProperties="OutputPath;SymbolsOutputPath"` → triggers `_GenerateSharedFrameworkDepsFile`

Different `RemoveProperties` = different global property dictionaries = different MSBuild configurations (see `ConfigurationMetadata.cs:162-170`) = no deduplication = both run `File.Create()` on the same deps.json.

The binlog you'd need to check is the per-project binlog for `Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj` (or look at the runtime.binlog that covers the installer subset), not the top-level VMR traversal binlog. The `GenerateInstallers` target you mention is unrelated — the race is in `_GenerateSharedFrameworkDepsFile` which runs during both Pack and Publish.

**Updated fix approach (v2)**: Instead of passing `IntermediateOutputPath` directly (which propagates to P2P refs and caused ILLink.Tasks.dll failures on Windows), we now pass a custom property `_DepsFileIntermediateDir` that only `_GenerateSharedFrameworkDepsFile` reads. This avoids P2P propagation while still isolating the deps.json write location.

JanProvaznik:
@rainersigwald The error in the original failing build (1428236) directly references this file:

```
artifacts/source-built-sdks/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets(198,5): error : The process cannot access the file '...Microsoft.NETCore.App.deps.json' because it is being used by another process.
```

The file is imported via `<Sdk Name="Microsoft.DotNet.SharedFramework.Sdk" />` in `Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj`. The race occurs between two self-calls on the sfxproj:

1. **Pack path**: `_AddFilesToNuGetPackage` (line 542) → self-call with `RemoveProperties="NoBuild"` → triggers `_GenerateSharedFrameworkDepsFile` 
2. **Publish path**: `GetFilesToPublish` (line 575) → self-call with `RemoveProperties="OutputPath;SymbolsOutputPath"` → triggers `_GenerateSharedFrameworkDepsFile`

Different `RemoveProperties` = different global property dictionaries = different MSBuild configurations (see `ConfigurationMetadata.cs:162-170`) = no deduplication = both run `File.Create()` on the same deps.json.

The binlog you'd need to check is the per-project binlog for `Microsoft.NETCore.App.Runtime.CoreCLR.sfxproj` (or look at the runtime.binlog that covers the installer subset), not the top-level VMR traversal binlog. The `GenerateInstallers` target you mention is unrelated — the race is in `_GenerateSharedFrameworkDepsFile` which runs during both Pack and Publish.

**Updated fix approach (v2)**: Instead of passing `IntermediateOutputPath` directly (which propagates to P2P refs and caused ILLink.Tasks.dll failures on Windows), we now pass a custom property `_DepsFileIntermediateDir` that only `_GenerateSharedFrameworkDepsFile` reads. This avoids P2P propagation while still isolating the deps.json write location.

PR #6779, opened 2026-05-22
Add macOS signing for libdotnet-aot.dylib and use ExecutableType meta…

mmitche:
…data

The new NativeAOT dotnet-aot shared library (libdotnet-aot.dylib) was missing from the signing configuration, causing notarization failures on macOS.

Also switch existing Mach-O FileSignInfo entries from using Condition="'$(TargetOS)' == 'osx'" to ExecutableType="MachO" metadata, which lets the signing tool match by binary type directly.


akoeplinger:
as discussed on Teams this is not the right fix, we're already signing .dylib

PR #6788, opened 2026-05-22
Add copilot instructions and binlog investigation skill

marcpopMSFT:
## Summary

Adds two new files to improve Copilot's ability to diagnose build issues in the VMR:

### .github/copilot-instructions.md
VMR-specific context that Copilot loads automatically:
- Documents the Arcade SDK / `SkipArcadeSdkImport` relationship
- Lists key Arcade-derived properties (`OfficialBuild` from `OfficialBuildId`)
- Explains NuGet audit configuration and why it's disabled in official builds
- Build command guidance

### .github/skills/binlog-investigation/SKILL.md
Diagnostic strategies for binlog investigation:
- **Strategy 1**: Start from errors, narrow to failing project first (avoid unscoped queries on large VMR binlogs)
- **Strategy 2**: Trace why a property has an unexpected value — check evaluation global properties, search property assignments, verify import chain
- **Strategy 3**: NuGet restore/audit error investigation patterns
- **Strategy 4**: Directory.Build.props chain analysis

## Motivation

While investigating [NU1903 errors in official builds](https://github.com/dotnet/dotnet/pull/6729), it took significant time to determine that `SkipArcadeSdkImport=true` on two projects was preventing `OfficialBuild` from being set, causing the `NuGetAudit=false` condition to not fire. These instructions and strategies encode the diagnostic patterns that would have led to a faster diagnosis.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,9 +77,42 @@ Repos are categorized in `eng/VmrLayout.props` as `SharedRepositoryReference` it
 - **Intermediate objects**: `artifacts/obj/`
 - **Test results**: `artifacts/TestResults/`
 
+## Arcade SDK and SkipArcadeSdkImport
+
+The VMR uses the [Microsoft.DotNet.Arcade.Sdk](https://github.com/dotnet/arcade) for shared build infrastructure. Arcade provides the common `eng/common/` scripts, signing, packaging, and CI patterns used across all .NET repos. The `eng/common/` directory at the VMR root (and within each `src/<repo>/`) is effectively read-only — it is sourced from arcade and overwritten during re-bootstrap or code flow. Do not modify these files directly; changes to shared build scripts must be made in the [dotnet/arcade](https://github.com/dotnet/arcade) repository.
+
+The root `Directory.Build.props` imports the Arcade SDK conditionally:
+
+```xml
+<Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk"
+        Condition="'$(SkipArcadeSdkImport)' != 'true'" />
+```
+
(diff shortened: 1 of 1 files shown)


marcpopMSFT:
I had copilot suggest these changes as I found it was spinning this morning investigating a binlog, I got frustrated and investigated myself, and ended up solving the issue faster than it had even starting 10 minutes late. I need to test if this set of changes will actually speed up its investigation.

ViktorHofer (changes requested):
We already have an AGENTS.md file which works with both copilot cli, claude code and others. It's discouraged to still use copilot-instructions.md unless you really only want to add copilot cli specific instructions.

We don't need to copy skills directly anymore. Marketplace with plugins are now natively supported: https://github.com/dotnet/runtime/blob/b88ce03bc28304f2b6cfd27c29923e142a9040c6/.github/copilot/settings.json#L11

steveisok:
> Marketplace with plugins are now natively supported: https://github.com/dotnet/runtime/blob/b88ce03bc28304f2b6cfd27c29923e142a9040c6/.github/copilot/settings.json#L11

That does not apply to CCA. If the goal is to help CCA investigate binlogs too, I think you'll still need to inline in AGENTS.md

marcpopMSFT:
Looking back through my copilot CLI session, it realized that the skill was in another repo but still chose to copy it in here without asking. Additionally, it looks like I was in a different branch that didn't have the agents.md file when the copilot started which is why it didn't modify that one. Let me see if it can fix this.

marcpopMSFT:
@ViktorHofer let's sync offline on how best to update arcade-skills and test locally. I want to eventually test the prompt I used to investigate the issue with and without my changes to see if it truly does help as copilot was definitely lost for quite a while (though it did figure it out eventually...I was just impacient).

PR #6797, opened 2026-05-23
Include current RID for ASP.NET Core bundled tools

DamianEdwards:
Addresses feedback from https://github.com/dotnet/dotnet/pull/6720#pullrequestreview-4350211177.

This appends the current product RID (`ProductMonikerRid`, derived from `TargetRid`) to the ASP.NET Core RID-specific bundled tool package list only when it is not already present. This lets community platform builds select their matching native AOT tool packages while preserving the existing official SDK RID list without duplicates.

@am11 @baronfel

### DIFF
--- a/src/sdk/src/Layout/redist/targets/BundledDotnetTools.targets
+++ b/src/sdk/src/Layout/redist/targets/BundledDotnetTools.targets
@@ -3,9 +3,8 @@
   <UsingTask TaskName="Microsoft.NET.Build.Tasks.SelectRuntimeIdentifierSpecificItems" AssemblyFile="$(MicrosoftNETBuildTasksAssembly)" />
 
   <PropertyGroup>
-    <AspNetCoreRidSpecificToolPackageRids>linux-x64;linux-arm;linux-arm64;linux-musl-x64;linux-musl-arm;linux-musl-arm64;win-x64;win-x86;win-arm64;osx-x64;osx-arm64</AspNetCoreRidSpecificToolPackageRids>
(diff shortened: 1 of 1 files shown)


src/sdk/src/Layout/redist/targets/BundledDotnetTools.targets
```diff
@@ -4,6 +4,8 @@
 
   <PropertyGroup>
     <AspNetCoreRidSpecificToolPackageRids>linux-x64;linux-arm;linux-arm64;linux-musl-x64;linux-musl-arm;linux-musl-arm64;win-x64;win-x86;win-arm64;osx-x64;osx-arm64</AspNetCoreRidSpecificToolPackageRids>
```
am11: nit: leading/trailing semicolons allow for reliable `Contains` condition below:
```suggestion
    <AspNetCoreRidSpecificToolPackageRids>;linux-x64;linux-arm;linux-arm64;linux-musl-x64;linux-musl-arm;linux-musl-arm64;win-x64;win-x86;win-arm64;osx-x64;osx-arm64;</AspNetCoreRidSpecificToolPackageRids>
```

DamianEdwards: Done in f08997d1f17 — the RID list is now delimiter-wrapped directly, and the current RID append check uses that list without the helper property.

DamianEdwards:
ping @lewing @akoeplinger for review

akoeplinger (approved):

PR #6798, opened 2026-05-23
Use lld for ASP.NET Core AOT musl tools

DamianEdwards:
Fixes the VMR rolling build failure reported after #6720 where NativeAOT publishing for the ASP.NET Core bundled tools selected the Linux default `bfd` linker for `linux-musl-*` RIDs, but the cross-build containers do not provide `ld.bfd`.

This scopes the workaround to the ASP.NET Core tool NativeAOT publishes by setting `LinkerFlavor=lld` only when `RuntimeIdentifier` starts with `linux-musl-`. Non-musl Linux publishes continue to use the existing NativeAOT default.

Tagging @lewing and @am11 from the #6720 discussion.

> [!NOTE]
> This PR description was generated by GitHub Copilot.

Validation:
- Restored the repo-local ASP.NET Core SDK with `src\aspnetcore\restore.cmd`.
- Evaluated `LinkerFlavor` via MSBuild for `dotnet-user-secrets`, `dotnet-user-jwts`, and `dotnet-dev-certs`: `linux-musl-arm64=lld`; `linux-x64=bfd`.

### DIFF
--- a/src/aspnetcore/src/Tools/Directory.Build.targets
+++ b/src/aspnetcore/src/Tools/Directory.Build.targets
@@ -16,6 +16,8 @@
   <PropertyGroup Condition=" '$(PublishAot)' == 'true' ">
     <InvariantGlobalization>true</InvariantGlobalization>
     <OptimizationPreference>Size</OptimizationPreference>
+    <!-- NativeAOT defaults Linux to bfd, but the VMR musl cross-build images provide lld instead. -->
+    <LinkerFlavor Condition=" '$(LinkerFlavor)' == '' and $(RuntimeIdentifier.StartsWith('linux-musl-')) ">lld</LinkerFlavor>
     <StackTraceSupport>false</StackTraceSupport>
     <EventSourceSupport>false</EventSourceSupport>
     <MetricsSupport>false</MetricsSupport>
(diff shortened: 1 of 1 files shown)


lewing:
This fixes the 3 Alpine/musl legs but **5 of 8 failing legs use non-musl RIDs** and will still fail:

| Leg | RID | Fixed? |
|---|---|---|
| Linux_Alpine_arm64 | `linux-musl-arm64` | ✅ |
| Linux_Alpine_x64 | `linux-musl-x64` | ✅ |
| Linux_Alpine_arm | `linux-musl-arm` | ✅ |
| Linux_arm64 | `linux-arm64` | ❌ |
| Linux_arm | `linux-arm` | ❌ |
| Linux_Pgo_x64 | `linux-x64` | ❌ |
| Linux_Pgo_arm64 | `linux-arm64` | ❌ |
| Linux_x64 | `linux-x64` | ❌ |

These are cross-compiled inside AzureLinux containers that also don't have `ld.bfd`, but the RID check `$(RuntimeIdentifier.StartsWith('linux-musl-'))` doesn't match them.

The runtime solved this with environment detection instead of RID matching — the `LocateNativeCompiler` target in [`eng/toolAot.targets`](https://github.com/dotnet/runtime/blob/main/eng/toolAot.targets#L28-L48) runs `init-compiler.sh` to probe which linker is actually available, then sets `LinkerFlavor` accordingly. That handles both musl and glibc cross-containers.

am11:
As a side note, we could dedup this target in arcade/eng/common/native and reuse it in sdk, runtime and maybe more repos under dotnet/ org in the future. It is basically compatible with dotnet engineering infrastructure.

As an end-user of PublishAot, I may just install cross binutils for cross publishing or just use the actual linux-arm64 machine etc. to build native stuff on the target platform (which is publicly documented).

lewing (approved):

akoeplinger (approved):

PR #6830, opened 2026-05-26
Retain sccsid symbol in native libs

### LINKED ISSUE #128575: source control version string missing in native libraries
```
$ strings /usr/lib64/dotnet/shared/Microsoft.NETCore.App/10.*/libSystem.Net.Security.Native.so | grep '@(#)'
@(#)Version 10.0.726.21808 @Commit: b16286c2284fecf303dbc12a0bb152476d662e44
```
vs
```
$ /tmp/dotnet/dotnet --version
11.0.100-preview.5.26272.112
$ strings /tmp/dotnet/shared/Microsoft.NETCore.App/11.*/libSystem.Net.Security.Native.so | grep '@(#)'
```
This probably gets stripped due to changes in https://github.com/dotnet/runtime/pull/128232.

@elinor-fung @jkotas @am11 ptal.

cc @omajid

am11:
Fixes https://github.com/dotnet/runtime/issues/128575.


akoeplinger (approved):
is this because we upgraded to some newer llvm? I wonder if/how far back we need to backport it

am11:
We started using `--gc-sections` linker flag (aggressing trimming of binaries) in https://github.com/dotnet/runtime/pull/128232 and @tmds caught it in time.

akoeplinger:
ok makes sense. there's another one in https://github.com/dotnet/runtime/blob/4e4d02afbe0b8144c8b3c8f1fdf25cacf5663b3c/src/coreclr/inc/corhlprpriv.h#L185-L188 and https://github.com/dotnet/diagnostics/blob/edf66fc88f26651024498cb7443eded5fa805439/src/shared/inc/corhlprpriv.h#L185-L189, not sure if they should be fixed in the same PR

am11:
Those are probably not even needed, lets see: https://github.com/dotnet/diagnostics/pull/5856 😅

janvorli:
> Those are probably not even needed, lets see: [dotnet/diagnostics#5856](https://github.com/dotnet/diagnostics/pull/5856)

The build of some legs fails due to the removal of the used attribute

src/runtime/src/coreclr/inc/corhlprpriv.h
```diff
@@ -182,11 +182,6 @@ class CQuickMemoryBase
         _Alloc<TRUE /*bGrow*/, TRUE /*bThrow*/>(iItems);
     }
 
-#ifdef __GNUC__
-    // This makes sure that we will not get an undefined symbol
```
jkotas, line 186: Some of the builds are failing due to this change
```
undefined symbol: CQuickMemoryBase<512ul, 128ul>::ReSizeNoThrow(unsigned long)
```
I think the proper fix is to move the template implementation from the .cpp file to here. It can be done in dotnet/runtime.

am11:
> > Those are probably not even needed, lets see: [dotnet/diagnostics#5856](https://github.com/dotnet/diagnostics/pull/5856)
> 
> The build of some legs fails due to the removal of the used attribute

@janvorli, interestingly, https://github.com/dotnet/runtime/issues/128575 is only reported for shared libs from `src/native/libs`, coreclr libs are retaining sccsid even without this fix (although copilot's PR https://github.com/dotnet/runtime/pull/128232 unconditionally applied the change). I haven't figured out why `--gc-sections` is less diligent in coreclr subset; sounds like it's leaving some stuff on the table (or perhaps linker has different dead code analyzer for C vs. C++)? My premise of last commit based on @akoeplinger's comment was if it didn't failed in `corhlprpriv.h` without `retain` attribute, then `used` isn't needed; but after a careful look, it's actually difference in compiler toolchain analysis of dead code.

src/runtime/src/coreclr/inc/corhlprpriv.h
am11: Opened https://github.com/dotnet/runtime/pull/128582.

jkotas (approved):
Thank you!

PR #6875, opened 2026-05-27
Port SuppressValidation fix for WiX ICE failures to preview5

mmitche:
Cherry-pick of #6398 from preview4.

Ports the fix that adds `-sval` (suppress all ICE validation) to WiX `light.exe` invocations. Without this, MSI repackaging after signing fails with `LGHT0217` errors when the Windows Installer service is unavailable on the build agent.

**Changes:**
- `LinkerToolTask.cs`: Replace `-sice:` with `-sval`
- `CreateLightCommandPackageDrop.cs`: Add `SuppressValidation` property and `-sval` flag support
- `wix.targets`: Set `SuppressValidation=true`
- `workloads.csproj`: Set `SuppressValidation=true`

Fixes the build failure in build [#2984751](https://dev.azure.com/dnceng/internal/_build/results?buildId=2984751).

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix/wix.targets
+++ b/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix/wix.targets
@@ -376,6 +376,7 @@
       NoLogo="true"
       Cultures="en-us"
       InstallerFile="$(_OutInstallerFile)"
+      SuppressValidation="true"
       WixExtensions="@(WixExtensions)"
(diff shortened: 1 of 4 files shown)


rbhanda (approved):

NikolaMilosavljevic (approved):

mmitche:
@joeloff The patterns on which this happens are pretty weird. We've had a bunch of successful builds, then nothing will pass. There's no change between what was in the previous successful build and now that would affect workloads, but workload builds started failing.

@wtgodbe These are worth tracking.

PR #6976, opened 2026-06-02
Add SkipUnchanged param to GenerateFileFromTemplate & ingest in aspnetcore

### LINKED ISSUE #66972: [Infrastructure]: Build incrementality issues
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Our builds don't seem to be properly incremental today.

At least working with `src/Validation` and running `build.cmd` multiple times without making any code changes would invoke Csc for some of the projects.

I think the first area to investigate is all our usages of Arcade's `GenerateFileFromTemplate`. While the task itself [doesn't support](https://github.com/dotnet/arcade/issues/16019) an option to skip writing when the file content matches, the targets where we invoke this task ourselves should have proper Inputs/Outputs so that the target is skipped altogether when not needed.

From a second build's binlog, I can see:

> Input file "C:\Users\ygerges\source\repos\aspnetcore\artifacts\bin\GenerateFiles\Directory.Build.targets" is newer than output file "C:\Users\ygerges\source\repos\aspnetcore\artifacts\obj\RepoTasks\Debug\RepoTasks.dll".

> Input file "C:\Users\ygerges\source\repos\aspnetcore\artifacts\bin\GenerateFiles\Directory.Build.targets" is newer than output file "C:\Users\ygerges\source\repos\aspnetcore\artifacts\obj\Microsoft.Extensions.Validation.Localization\Debug\net11.0\Microsoft.Extensions.Validation.Localization.xml".

(etc, for many projects. I haven't looked exactly where the Directory.Build.targets is written but I'm guessing it's one of the `GenerateFileFromTemplate` invocations that we have)

This is currently hurting productivity a lot, for me at least :smile:

wtgodbe:
Prevents unnecessary re-generation of unchanged template files - re-generating those files updates the timestamp, which can cascade into re-running compilation targets & add extra build time.

Fixes https://github.com/dotnet/aspnetcore/issues/66972

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Build.Tasks.Templating/src/GenerateFileFromTemplate.cs
+++ b/src/arcade/src/Microsoft.DotNet.Build.Tasks.Templating/src/GenerateFileFromTemplate.cs
@@ -49,6 +49,16 @@ namespace Microsoft.DotNet.Build.Tasks.Templating
         public string[] Properties { get; set; }
 
         /// <summary>
+        /// When <see langword="true"/>, the output file is only written if it does not already exist
+        /// or if its current contents differ from what the task would write. Defaults to <see langword="false"/>.
+        /// <para>
+        /// Enabling this preserves the output file's timestamp when its contents are unchanged, which can
+        /// significantly speed up incremental builds by avoiding unnecessary invalidation of downstream targets.
+        /// </para>
+        /// </summary>
+        public bool SkipUnchanged { get; set; }
+
+        /// <summary>
         /// The destination for the generated file resolved by this task.
         /// </summary>
         [Output]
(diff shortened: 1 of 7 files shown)


ViktorHofer (approved):
Build perf will be slower in the non-incremental build case due to the comparison but that is likely neglectable. Basically, the same recommendations that we provide for the `SkipUnchangedFiles` parameter of the `Copy` task apply here. LGTM

src/arcade/src/Microsoft.DotNet.Build.Tasks.Templating/src/GenerateFileFromTemplate.cs
```diff
@@ -68,12 +78,63 @@ public override bool Execute()
...
+            {
+                Directory.CreateDirectory(directory);
+            }
+
+            File.WriteAllBytes(ResolvedOutputPath, resultBytes);
 
             return !Log.HasLoggedErrors;
         }
 
+        private static bool FileContentsMatch(string path, byte[] expectedBytes)
```
akoeplinger, line 103: FYI we have a slightly more advanced version of this logic in https://github.com/dotnet/runtime/blob/main/src/tasks/Common/Utils.cs#L217 which uses Spans for the comparison of the buffers. Probably overkill for the usage here but just wanted to mention it

PR #6987, opened 2026-06-03
Add nuget.org as NuGet audit source for PR builds

marcpopMSFT:
## Summary

Follow-up to #6753. The prior change disabled NuGet audit for official builds to work around dotnet/msbuild#10801, but it also effectively disabled audit for PR validation because the Azure DevOps feeds in NuGet.config don't serve vulnerability advisory data.

This PR adds `RestoreAuditSources` pointing to `https://api.nuget.org/v3/index.json` for non-official builds so that NuGet audit can fetch advisories during PR validation.

## Testing

The second commit intentionally downgrades `System.Security.Cryptography.Xml` to 10.0.0 (which has known vulnerabilities) to verify the audit source works. If the PR build flags NU1903/NU1904 warnings, the configuration is working correctly. That commit should be reverted before merge.

## Changes

1. **Directory.Build.props**: Add `RestoreAuditSources` property targeting nuget.org, conditioned on non-official builds
2. **eng/Versions.props**: Temporary downgrade of System.Security.Cryptography.Xml for validation


Directory.Build.props
```diff
@@ -78,6 +78,10 @@
     <NuGetAudit Condition="'$(OfficialBuild)' == 'true'">false</NuGetAudit>
     <!-- Explicitly set NuGetAuditModel level as it's currently disabled in the product. -->
     <NuGetAuditMode>all</NuGetAuditMode>
+    <!-- Configure nuget.org as the audit source for non-official builds (PR validation).
+         The Azure DevOps feeds in NuGet.config don't serve vulnerability data, so we need
+         to explicitly point NuGet at nuget.org for audit advisory lookups. -->
+    <NuGetAuditSources Condition="'$(OfficialBuild)' != 'true'">https://api.nuget.org/v3/index.json</NuGetAuditSources>
```
akoeplinger: I think we could use data.nuget.org as the audit source: https://learn.microsoft.com/en-us/nuget/concepts/auditing-packages#audit-sources

nuget-client is using it: https://github.com/NuGet/NuGet.Client/blob/dev/NuGet.Config#L12-L15

ViktorHofer (changes requested):
For others, we have an offline chat about this. `NuGetAuditSources` isn't a property that NuGet reads so this doesn't work. We recommended to use data.nuget.org in NuGet.config and switching the NuGetAudit property to false for official builds.

marcpopMSFT:
Looks like it works as expected but the offline build can't access the audit source and fails. I think we need to suppress that warning in that build or disable audit in that build (assuming disabling audit is enough to take care of that warning message).

ViktorHofer:
https://github.com/dotnet/dotnet/blob/0068a027f7fc8202436e4fe85d185bc9607df490/repo-projects/Directory.Build.props#L225 -> NuGet Audit is disabled when building source-only (not sure why). The SB offline errors are from the VMR orchestrator - so the outer repo - dotnet/dotnet itself. I think you want to disable it in a file that is imported by both the repo infra and Arcade SDK infra. eng/Versions.props should work.

marcpopMSFT:
I've confirmed that this is the minimum set of changes needed to have audit flag in PR builds, not fail in offline builds because of access to nuget.org, and not fail in CI. I ran both PR and CI with commit 24d1b19f1add3931ee3dfa5fac5738067218ff63 and the PR build only failed with expected audit errors and CI didn't fail.

ViktorHofer (commented):
Good progress. This only enables NuGet Audit in the VMR orchestrator - the non-interesting part of the VMR. Is that intentional?

ViktorHofer (approved):

marcpopMSFT:
@ViktorHofer The orchestrator was the one being flagged before so that's what I targeted.  I think I should merge this and then do any follow-ups as a separate PR. Are there additional places you think I should target that would be the most valuable? I wasn't planning on doing every nuget.config in the repo.

PR #7004, opened 2026-06-03
Share NativeAotSupported check via arcade

am11:
Shared with runtime and SDK.


am11:
cc @akoeplinger please run linux-musl type of leg against this PR so we can see cross-build is intact.

src/sdk/eng/common/native/LocateNativeCompiler.targets
```diff
@@ -0,0 +1,27 @@
+<Project>
```
akoeplinger, line 1: this one should be removed

src/runtime/eng/common/native/NativeAotSupported.props
```diff
@@ -0,0 +1,30 @@
+<Project>
+  <PropertyGroup>
+
+    <!-- Reject unsupported OS via RID prefix match -->
+    <_NativeAotSupportedOS Condition="
+      !$(TargetRid.StartsWith('browser-')) and
```
akoeplinger: this uses TargetRid instead of TargetOS in runtime, I don't remember if there were any gotchas with that?

src/sdk/eng/common/native/LocateNativeCompiler.targets
am11: This is used under sdk/ in this PR.

src/runtime/eng/common/native/NativeAotSupported.props
am11: Subsets.props uses both split form (TargetOS/Architecture) and joined form (TargetRid), in SDK which we are sharing with only have the merged form (TargetRid), so I used the common prop.

src/sdk/eng/common/native/LocateNativeCompiler.targets
akoeplinger: oh sorry I thought this got merged as part of https://github.com/dotnet/dotnet/commit/a1e77c9a2090b736f3358ce912aab5fecb1bd849

src/runtime/eng/common/native/NativeAotSupported.props
akoeplinger: Subsets.props only uses TargetRid in one place.

We need to be careful to make sure this has the same behavior, given that TargetRid goes through portable/non-portable/distro RID logic. I don't have time right now to think this through :D

am11: switched to TargetOS and TargetArchitecture.

am11:
cc @baronfel, @MichalStrehovsky, similar to @DamianEdwards' aspnetcore approach https://github.com/dotnet/dotnet/pull/6798 but for SDK. We are sharing them via arcade to stay consistent.

Single point to determine:

* List of NativeAOT supported platforms (majority of platforms are these days; remaining ones are 'port in progress')
* Cross-build linker flags compatible with .NET engineering infrastructure

baronfel:
cc @JeremyKuhne for visibility - we'll likely want to adopt this pattern in MSBuild as well

src/arcade/eng/common/native/NativeAotSupported.props
```diff
@@ -0,0 +1,30 @@
+<Project>
+  <PropertyGroup>
+
+    <!-- Reject unsupported OS via RID prefix match -->
```
MichalStrehovsky, line 4: Any reason why one is a reject list and the other is allow list?

I think reject list is better because it will keep shrinking. Allow list will just have an unbounded growth.

The fact we have an allow list in https://github.com/dotnet/dotnet/blob/a81a7ebcc8f81b179d47839ba75a3f4758efb6b5/src/runtime/src/coreclr/CMakeLists.txt#L158 is probably the reason why these are not in sync in general. As people keep adding platform support, they forget about native AOT. We're approaching a point where it might not be possible to get a working SDK without native AOT support so people will not be able to cut corners here anymore (we never had a reject/allow list for R2R, even though R2R is also technically an optional technology).

am11: More operating systems are allowed, so it was less typing this way I think. Architecture checks are a bit more involved but they can also be streamlined on parity. We used to have the full RID list but it was refactored / split later to OS and arch. The combination of two implies "we are happy" with the entire package and can use it even during the product build (for crossgen2, ilcompiler, cdac, xunitlogger etc.); not just "AOT runtime components compile" because creation of package and managed tools that goes in it are controlled by this `NativeAotSupported` property.

am11: The reject list of arch is longer as expected, I have pushed the change nonetheless.

src/runtime/src/coreclr/CMakeLists.txt
```diff
@@ -154,8 +154,9 @@ if (NOT CLR_CMAKE_TARGET_BROWSER OR CLR_CROSS_COMPONENTS_BUILD)
 endif ()
 
 if(NOT CLR_CROSS_COMPONENTS_BUILD)
-  # NativeAOT only buildable for a subset of CoreCLR-supported configurations
-  if(CLR_CMAKE_HOST_ARCH_ARM64 OR CLR_CMAKE_HOST_ARCH_AMD64 OR CLR_CMAKE_HOST_ARCH_ARM OR CLR_CMAKE_HOST_ARCH_LOONGARCH64 OR CLR_CMAKE_HOST_ARCH_RISCV64 OR (CLR_CMAKE_HOST_ARCH_I386 AND CLR_CMAKE_HOST_WIN32))
+  # NativeAOT is buildable for all CoreCLR-supported configurations except a known reject list.
+  # Keep this in sync with the NativeAotSupported reject list in eng/common/native/NativeAotSupported.props.
+  if(NOT (CLR_CMAKE_HOST_ARCH_ARMV6 OR CLR_CMAKE_HOST_ARCH_S390X OR CLR_CMAKE_HOST_ARCH_POWERPC64 OR CLR_CMAKE_HOST_ARCH_WASM OR CLR_CMAKE_HOST_ARCH_MIPS64 OR (CLR_CMAKE_HOST_ARCH_I386 AND NOT CLR_CMAKE_HOST_WIN32)))
```
MichalStrehovsky: Does this list only need to include architectures that rest of CoreCLR can build with?

E.g. I don't expect happiness when building CoreCLR for PowerPC. If someone does a bringup of PowerPC with CoreCLR on the future, I prefer they also do native AOT, same way they would do e.g. ReadyToRun or the debugger infra (these are all "skippable components"). If they want to make a call not to support PowerPC with native AOT, they can add it to the list then.

src/arcade/eng/common/native/NativeAotSupported.props
```diff
@@ -10,14 +10,15 @@
...
-    <!-- Allow architectures via RID suffix match -->
+    <!-- Reject unsupported architectures via RID suffix match -->
     <_NativeAotSupportedArch Condition="
-      '$(TargetArchitecture)' == 'x64' or
-      '$(TargetArchitecture)' == 'arm64' or
-      '$(TargetArchitecture)' == 'arm' or
-      '$(TargetArchitecture)' == 'loongarch64' or
-      '$(TargetArchitecture)' == 'riscv64' or
-      ('$(TargetOS)' == 'windows' and '$(TargetArchitecture)' == 'x86')
+      '$(TargetArchitecture)' != 'armel' and
```
MichalStrehovsky: I thought armel works, Samsung did work on that long time ago: https://github.com/dotnet/corert/pulls?q=is%3Apr+armel+is%3Aclosed

MichalStrehovsky: (If we're not sure, we can just assume it works and see if anyone complains. I think this list could match the list in CMakeLists since I don't expect Mono's support matrix is very relevant at this point, except probably s390x.)

src/runtime/src/coreclr/CMakeLists.txt
```diff
@@ -154,8 +154,9 @@ if (NOT CLR_CMAKE_TARGET_BROWSER OR CLR_CROSS_COMPONENTS_BUILD)
 endif ()
 
 if(NOT CLR_CROSS_COMPONENTS_BUILD)
-  # NativeAOT only buildable for a subset of CoreCLR-supported configurations
-  if(CLR_CMAKE_HOST_ARCH_ARM64 OR CLR_CMAKE_HOST_ARCH_AMD64 OR CLR_CMAKE_HOST_ARCH_ARM OR CLR_CMAKE_HOST_ARCH_LOONGARCH64 OR CLR_CMAKE_HOST_ARCH_RISCV64 OR (CLR_CMAKE_HOST_ARCH_I386 AND CLR_CMAKE_HOST_WIN32))
+  # NativeAOT is buildable for all CoreCLR-supported configurations except a known reject list.
+  # Keep this in sync with the NativeAotSupported reject list in eng/common/native/NativeAotSupported.props.
+  if(NOT (CLR_CMAKE_HOST_ARCH_ARMV6 OR CLR_CMAKE_HOST_ARCH_S390X OR CLR_CMAKE_HOST_ARCH_WASM OR CLR_CMAKE_HOST_ARCH_MIPS64 OR (CLR_CMAKE_HOST_ARCH_I386 AND NOT CLR_CMAKE_HOST_WIN32)))
```
MichalStrehovsky: Does CoreCLR really build with ARMV6/S390X/MIPS64?

am11: Yes, `git grep -iE 's390x|mips64|armv6' -- ':/src/runtime/src/coreclr'`


all mono targets are using coreclr-pal at least, and they are gradually moving to coreclr since mono is on its way to retirement in runtime repo.

MichalStrehovsky: There are no hits for mips64 under src/coreclr, except for ELF.h but that doesn't count. So mips64 should not be here.

I do not think `build.sh clr` work for armv6. What was brought up is basically clr.iltools and maybe some other small bits: https://github.com/dotnet/runtime/pull/62594#discussion_r791382856. I don't see us excluding things that don't work: one needs to build the right subset. So armv6 should also be out.

I'm pretty sure the same applies to s360x.

am11: What about these:

```sh
src/runtime/src/coreclr/inc/volatile.h:#error The Volatile type is currently only defined for GCC when targeting x86, AMD64, ARM, ARM64, LOONGARCH64, RISCV64, PPC64LE, or S390X CPUs
src/runtime/src/coreclr/inc/volatile.h:#if defined(HOST_ARMV6)
src/runtime/src/coreclr/inc/volatile.h:// DMB ISH not valid on ARMv6
src/runtime/src/coreclr/jit/CMakeLists.txt:  elseif((TARGETDETAILS_ARCH STREQUAL "armv6") OR (TARGETDETAILS_ARCH STREQUAL "armv6l"))
src/runtime/src/coreclr/jit/CMakeLists.txt:    set(JIT_ARCH_SOURCES ${JIT_ARMV6_SOURCES})
src/runtime/src/coreclr/jit/CMakeLists.txt:    set(JIT_ARCH_HEADERS ${JIT_ARMV6_HEADERS})
src/runtime/src/coreclr/jit/CMakeLists.txt:  elseif(TARGETDETAILS_ARCH STREQUAL "s390x")
src/runtime/src/coreclr/jit/CMakeLists.txt:    set(JIT_ARCH_SOURCES ${JIT_S390X_SOURCES})
src/runtime/src/coreclr/jit/CMakeLists.txt:    set(JIT_ARCH_HEADERS ${JIT_S390X_HEADERS})
src/runtime/src/coreclr/jit/CMakeLists.txt:set( JIT_ARMV6_SOURCES
src/runtime/src/coreclr/jit/CMakeLists.txt:set( JIT_S390X_SOURCES
src/runtime/src/coreclr/jit/CMakeLists.txt:set ( JIT_ARMV6_HEADERS
src/runtime/src/coreclr/jit/CMakeLists.txt:set ( JIT_S390X_HEADERS
src/runtime/src/coreclr/jit/CMakeLists.txt:convert_to_absolute_path(JIT_ARMV6_SOURCES ${JIT_ARMV6_SOURCES})
src/runtime/src/coreclr/jit/CMakeLists.txt:convert_to_absolute_path(JIT_ARMV6_HEADERS ${JIT_ARMV6_HEADERS})
src/runtime/src/coreclr/jit/CMakeLists.txt:convert_to_absolute_path(JIT_S390X_SOURCES ${JIT_S390X_SOURCES})
src/runtime/src/coreclr/jit/CMakeLists.txt:convert_to_absolute_path(JIT_S390X_HEADERS ${JIT_S390X_HEADERS})
src/runtime/src/coreclr/jit/CMakeLists.txt:elseif(CLR_CMAKE_TARGET_ARCH_ARMV6)
src/runtime/src/coreclr/jit/CMakeLists.txt:  set(JIT_ARCH_SOURCES ${JIT_ARMV6_SOURCES})
src/runtime/src/coreclr/jit/CMakeLists.txt:  set(JIT_ARCH_HEADERS ${JIT_ARMV6_HEADERS})
src/runtime/src/coreclr/jit/CMakeLists.txt:elseif(CLR_CMAKE_TARGET_ARCH_S390X)
src/runtime/src/coreclr/jit/CMakeLists.txt:  set(JIT_ARCH_SOURCES ${JIT_S390X_SOURCES})
src/runtime/src/coreclr/jit/CMakeLists.txt:  set(JIT_ARCH_HEADERS ${JIT_S390X_HEADERS})
src/runtime/src/coreclr/jit/CMakeLists.txt:    if (NOT ARCH_TARGET_NAME STREQUAL s390x AND NOT ARCH_TARGET_NAME STREQUAL armv6 AND NOT ARCH_TARGET_NAME STREQUAL ppc64le AND NOT ARCH_TARGET_NAME STREQUAL riscv64 AND NOT ARCH_TARGET_NAME STREQUAL wasm)
```

MichalStrehovsky: The volatile.h hit is from the PR I mentioned above that only brought up clr.iltools (the PR specifically says only subset of clr can be built and even mentions build breaks if more than that tried to be built).

The s390x in the jit is from https://github.com/dotnet/runtime/commit/54f78b8a2169e2d988c6aee34a6b9a9d744ed754#diff-88a7756f71b2a189c45e9cd0a135ea65d2054e1b88b983f029738478a8114412. That one is also `This suffices to make clr.iltools and clr.paltests buildable`.

I would prefer restricting this list as much as possible even at the risk of breaking something vs building a cargo cult around a build of CoreCLR that don't exist and nobody uses and we're now handling. We did some level of due diligence.

am11: > I would prefer restricting this list as much as possible even at the risk of breaking something

Ok. If this breaks community pipeline in dotnet/runtime (which has armv6, s390x and ppc64le), then they can fix it. This pipeline is broken from time to time, so it's not critical.

MichalStrehovsky (approved):
Thank you! This is great. I expect the worst thing this could cause is an issue during CMake configure, but looking at the ARMv6 or s390 clr.iltools bringups, I expect it to be minor or nothing at all.

@akoeplinger I have not done changes to eng/common since the monorepo got introduced. Is this good to go from flow perspective? One thing I'm not sure about is the eng/common/native/LocateNativeCompiler.target being only present in sdk, we might need a copy in the other directories.

am11:
eng/common brings significant portion of shared engineering infra  and it is on auto sync. I’ve only added files to SDK and runtime which I needed to successfully compile.

MichalStrehovsky:
> eng/common brings significant portion of shared engineering infra and it is on auto sync. I’ve only added files to SDK and runtime which I needed to successfully compile.

The specific concern I have is that LocateNativeCompiler.targets only exists in sdk/eng/common/native/LocateNativeCompiler.targets but it's missing in arcade's eng/common. AFAIK arcade's eng/common is the source of truth and automation will make sure the subrepos match arcade. In the past it wasn't sufficient to put a file into a subrepo and expect it to land in arcade (e.g. yours https://github.com/dotnet/arcade/pull/16756 that mirrored a file from dotnet/runtime's copy).

am11:
It was sync'd with arcade as well https://github.com/dotnet/arcade/blob/56539035afb7a5b9ef1be4ec42aa654863995f08/eng/common/native/LocateNativeCompiler.targets. You are actually raising a good point; I had the same question and @ViktorHofer explained it to me. My understanding is, either:

* Update eng/common in dotnet/arcade repo, it will flow to the consumer repos
* or update src/arcade/eng/common in VMR (e.g. if we want to test stuff simultaneously across the repos like this PR) and it will flow to arcade along with other repos.

Note that VMR's top-level eng/common is not the source-of-truth; arcade is as before.

MichalStrehovsky:
> It was sync'd with arcade as well https://github.com/dotnet/arcade/blob/56539035afb7a5b9ef1be4ec42aa654863995f08/eng/common/native/LocateNativeCompiler.targets

Ah, that's the key part I was missing, thanks!

src/arcade/eng/common/native/NativeAotSupported.props
ViktorHofer, line 1: Beware that only a small subset of repositories that use Arcade need this and even a smaller subset of that use the TargetOS and TargetArchitecture properties that are being used here. We would like to eventually standardize these input controls (properties) in the Arcade SDK and automatically set them. Can't find a tracking issue.

am11: NativeAOT is gradually making its way to other repos. Recently (in past few weeks) aspnetcore and SDK tools were onboarded to PublishAot plan, I imagine other repos will follow the course.

ViktorHofer: Why doesn't this logic then live in the shipping SDK?

am11: This is compatible with .NET engineering infra where we use LLVM's linker (lld) on all Unix platforms (except Apple platforms). In shipping configuration, we use platform defaults (GNU's bfd linker on Linux, lld on FreeBSD, ld64 on macOS and so on).

In dotnet prereqs containers, we build our own LLVM toolchain with custom configurations https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/57f70a1e66e1d5a0568b1ad30a98995d9d7d68fb/src/azurelinux/3.0/net11.0/crossdeps-builder/amd64/Dockerfile#L83 and prefer using it over GNU/gcc toolchain, which has some conditions in their GPL license which are not well-compatible with our MIT license (they aren't exactly incompatible, but aren't welcoming; e.g. they give "exemption" for non GPL license software so we have to be very careful what we can or can't use from GNU toolchain).

PR #7005, opened 2026-06-03
Replace `darc vmr scan-cloaked-files` with an in-VMR tool

akoeplinger:
Replaces the `darc vmr scan-cloaked-files` invocation in the VMR scan pipeline with a dedicated in-VMR tool at `eng/tools/CloakedFileScan/`, modeled after `eng/tools/ChangeValidation/`.

The new tool consumes `Microsoft.DotNet.DarcLib` and instantiates `VmrCloakedFileScanner` directly, producing identical scan results (git-diff against the empty tree with cloaked path filters from `src/source-mappings.json`) and emitting AzDO `##vso[task.logissue type=error]` annotations for any matches.

### Why

This lets us drop the pinned `microsoft.dotnet.darc` global tool. Removed:
- `microsoft.dotnet.darc` entry from `.config/dotnet-tools.json` (now empty)
- `Microsoft.DotNet.Darc` `<Dependency>` from `eng/Version.Details.xml`

`Microsoft.DotNet.DarcLib` is kept — it's already used by `ChangeValidation` and is now also used by the new tool.

### Pipeline change

`eng/pipelines/templates/stages/vmr-scan.yml` no longer needs to bootstrap the .NET CLI separately or run `dotnet tool restore`; it just runs the project via `eng/common/dotnet.sh run`.

### Verification

Built and run locally against this VMR:
- Clean tree → `0 cloaked files`, exit 0.
- Staged a test file matching a cloak rule (`src/roslyn/.../ObfuscatedCloakTest.dll`) → tool reported it with the expected AzDO error annotation, exit 1.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,12 +1,5 @@
 {
   "version": 1,
   "isRoot": true,
-  "tools": {
-    "microsoft.dotnet.darc": {
-      "version": "1.1.0-beta.26268.2",
-      "commands": [
-        "darc"
-      ]
-    }
-  }
+  "tools": {}
 }
--- a/eng/tools/CloakedFileScan/Program.cs
+++ b/eng/tools/CloakedFileScan/Program.cs
@@ -1,0 +1,62 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Maestro.Common;
+using Microsoft.DotNet.DarcLib.Helpers;
+using Microsoft.DotNet.DarcLib.VirtualMonoRepo;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace CloakedFileScan;
+
+internal static class Program
+{
+    private static void LogInfo(string message) => Console.WriteLine(message);
+    private static void LogError(string message) => Console.WriteLine($"##vso[task.logissue type=error] {message}");
+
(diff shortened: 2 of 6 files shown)


.config/dotnet-tools.json
```diff
@@ -1,12 +1,5 @@
...
   "isRoot": true,
-  "tools": {
-    "microsoft.dotnet.darc": {
-      "version": "1.1.0-beta.26268.2",
-      "commands": [
-        "darc"
-      ]
-    }
-  }
+  "tools": {}
```
premun, line 4: This file can go?

premun (approved):

.config/dotnet-tools.json
akoeplinger: yeah, but I wasn't sure whether the `"isRoot": true,` isn't important for something somewhere

eng/tools/CloakedFileScan/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+    internal static async Task<int> Main()
+    {
+        var pm = new ProcessManager(NullLogger<ProcessManager>.Instance, "git");
+        var vmrPath = pm.FindGitRoot(AppContext.BaseDirectory);
+
+        LogInfo($"VMR root: {vmrPath}");
+
+        var serviceProvider = RegisterServices(vmrPath);
+        await using var scope = serviceProvider.CreateAsyncScope();
+
```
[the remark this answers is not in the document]
akoeplinger: this is fine in this case

akoeplinger:
Not tested in public PR builds, fast merge.

PR #7076, opened 2026-06-05
[release/10.0.4xx] Quote paths with spaces in MSBuild bootstrap InstallDotNetCoreTask

lewing:
Cherry-pick of #7075 to release/10.0.4xx to unblock currently-broken Windows builds.

All Windows legs on 10.0.4xx official and PR builds are failing with:
\\\
Install-scripts was not executed successfully
\\\

Root cause: 1ES agents were updated with SDK 10.0.300 system-wide at \C:\Program Files\dotnet\, causing \DotNetRoot\ to resolve to a path with spaces. The unquoted path breaks PowerShell's \-File\ argument parsing.

See #7075 for full details.


JanProvaznik:
this is not enough, it'll also fail in sdk; fixing it in our codeflow https://github.com/dotnet/dotnet/pull/7061

lewing:
> this is not enough, it'll also fail in sdk; fixing it in our codeflow #7061

every branch is failing

PR #7124, opened 2026-06-08
Revert "Aggregate ASP.NET Core tool executables (#7020)"

lewing:
This reverts commit 7150f20253ebab0d6d013a3e9d36a8b7e1e60cf9 (PR #7020).

## Problem

The aspnetcoretools aggregation introduced by #7020 is causing **ILC SourceLinkWriter crashes on Windows**, breaking all main branch Windows vertical builds (both official and PR CI).

**Error:** `ILCompiler.SourceLinkWriter..ctor` at `SourceLinkWriter.cs:27` (`File.CreateText`) throws a file-locked exception:
```
aspnetcoretools.sourcelink because it is being used by another process
```

**Impact:**
- All main branch internal builds failing since June 7 (builds 2994281, 2994408)
- All main branch PR CI builds failing on Windows_x64 and Windows_x86 legs
- Code flow PRs #7107 (efcore), #7115 (sdk), #7114 (emsdk), #7118 (node) are all blocked

## Root Cause Analysis

The crash has **three contributing factors**:

### 1. Multi-RID parallel ILC writes to same file
`aspnetcoretools.csproj` sets `RuntimeIdentifiers` (potentially multiple RIDs) + `NativeLib=Shared`. MSBuild creates parallel build iterations per RID, and all ILC processes race on writing `aspnetcoretools.sourcelink` via `File.CreateText()` — which on Windows uses exclusive file locking.

### 2. Test project triggers overlapping publishes
`Microsoft.AspNetCore.Tools.NativeAot.Tests.csproj` has a `PublishNativeAotToolsForTests` target that runs `BeforeTargets="Build"` and calls `MSBuild Targets="Publish"` on `aspnetcoretools.csproj`. This can overlap with the main build's publish of aspnetcoretools, creating concurrent ILC processes writing to the same intermediate files.

### 3. Individual tools still have `PublishAot=true`
The three tool `.csproj` files (`dotnet-dev-certs`, `dotnet-user-jwts`, `dotnet-user-secrets`) still declare `PublishAot=true` independently. This causes double ILC compilation — once via the aggregate project and once individually — creating additional file contention.

### Why Windows-only?
Linux/macOS don't enforce exclusive file locking on `File.CreateText`, so concurrent writes succeed silently. Windows does enforce it, causing `IOException` at `SourceLinkWriter.cs:27`.

## Timeline
- Build 2994109 (June 6, **before** #7020 merge) — no SourceLinkWriter crash, aspnetcoretools built successfully
[...]

PR #7020, opened 2026-06-04
Aggregate ASP.NET Core tool executables

DamianEdwards:
Updates the tools bundled into the SDK from `dotnet/aspnetcore` to use the new preview "aggregate executable shims" feature to reduce the layout size on disk.

A new `aspnetcoretools.csproj` was added to produce the aggregated executables and the SDK tool layout logic has been updated to resolve `dotnet user-secrets`, `dotnet user-jwts`, and `dotnet dev-certs` from the aggregate tool shims folder.

Source build doesn't build the aggregated shims or native aot, preferring the pre-existing layout.

@baronfel @MichalStrehovsky @BrennanConroy

Closes dotnet/aspnetcore#66976

PR #7131, opened 2026-06-08
Restore aggregate ASP.NET Core tool executables

DamianEdwards:
Re-lands the aggregate ASP.NET Core tool executable changes from #7020 after the revert in #7124.

#7124 identified Windows ILC `SourceLinkWriter` file-lock failures caused by overlapping NativeAOT publishes writing the same `aspnetcoretools.sourcelink` intermediate. This re-land keeps the aggregate executable layout, but gives the test-triggered aggregate publish its own `NativeIntermediateOutputPath` so it cannot race with the product publish on the same SourceLink/intermediate files.

Validation performed locally on Windows:
- Restored and published `src\aspnetcore\src\Tools\NativeAot\aspnetcoretools\aspnetcoretools.csproj` for `win-x64`.
- Ran `PublishNativeAotToolsForTests` on `src\aspnetcore\src\Tools\NativeAot\test\Microsoft.AspNetCore.Tools.NativeAot.Tests.csproj` for `win-x64`.
- Confirmed the product and test publishes now produce distinct `aspnetcoretools.sourcelink` paths under separate `artifacts\obj` directories.

Note: the full documented ASP.NET Core `eng\build.cmd` path is currently blocked before reaching these projects by an unrelated GenerateFiles task/package mismatch where the resolved `Microsoft.DotNet.Build.Tasks.Templating` task does not expose the `SkipUnchanged` parameter used by source.


src/sdk/test/dotnet.Tests/CommandFactoryTests/GivenADotnetToolsCommandResolver.cs
```diff
@@ -84,4 +84,40 @@ public void ItReturnsAnExecutableCommandSpecWhenExecutableExists()
...
+                });
+
+                result.Should().NotBeNull();
+                result.Path.Should().Be(executablePath);
+                result.Args.Should().Be("--help");
+            }
+            finally
+            {
+                Directory.Delete(dotnetToolPath, recursive: true);
+            }
```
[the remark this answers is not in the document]
DamianEdwards: Addressed by making the temp directory cleanup best-effort: the test now guards with Directory.Exists and ignores cleanup failures, while staying independent of SdkTest/redist setup. I also rebuilt dotnet.Tests and ran the focused aggregate tool resolver test via dotnet exec successfully.

src/aspnetcore/src/Tools/NativeAot/test/Microsoft.AspNetCore.Tools.NativeAot.Tests.csproj
```diff
@@ -22,25 +23,14 @@
...
   <Target Name="PublishNativeAotToolsForTests"
     BeforeTargets="Build"
     Condition="'$(DotNetBuildSourceOnly)' != 'true'">
+    <!-- Keep test publishes from racing with product publishes on NativeAOT intermediate files. -->
     <MSBuild
-      Projects="%(NativeAotToolUnderTest.ProjectPath)"
+      Projects="..\aspnetcoretools\aspnetcoretools.csproj"
       Targets="Publish"
-      Properties="Configuration=$(Configuration);RuntimeIdentifier=$(NativeAotToolPublishRuntimeIdentifier);SelfContained=true;PublishAot=true;PublishDir=$([MSBuild]::NormalizeDirectory('$(NativeAotToolPublishRoot)', '%(NativeAotToolUnderTest.Identity)'))" />
+      Properties="Configuration=$(Configuration);RuntimeIdentifier=$(NativeAotToolPublishRuntimeIdentifier);SelfContained=true;PublishAot=true;PublishDir=$(NativeAotToolPublishRoot);NativeIntermediateOutputPath=$(NativeAotToolPublishIntermediateRoot)" />
```
dsplaisted: It looks like this is just redirecting the native output, so you still could have races and file locks for the normal build part of the project.

The ideal solution is probably to depend on the published assets of the product build here, though might be complicated to do.

Alternatively, you could try overriding IntermediateOutputPath and OutputPath to avoid collisions with the product build, but then you need to do a separate restore, which is also problematic.

DamianEdwards: Do sufficient legs run on PR here to get confidence of reliability? There doesn't seem to be a great path forward right now if this change isn't suffiicient.

dsplaisted: > Do sufficient legs run on PR here to get confidence of reliability?

Probably not.  This fix probably improves things so it might be enough, but if you merge it you might end up with VMR build failures.

We went through similar issues in the SDK repo.  Here's copilot's recommendations based on that experience:

I'd remove the race by either eliminating the *concurrency* or eliminating the *sharing*. Two options, roughly in order of preference:

**1. Add a real dependency edge so the test consumes/orders after the product publish (preferred).**
This is the "depend on the published assets" idea, and it turned out to be the clean fix for us. The pattern:
- Have `aspnetcoretools.csproj` publish-as-part-of-build (set `_IsPublishing=true` and a `PublishOnBuild` target that runs `Publish` after `Build`), so building that project always produces the published native output.
- Have the test project take a `ProjectReference` to `aspnetcoretools.csproj` (`ReferenceOutputAssembly=false`, `SkipGetTargetFrameworkProperties=true`, `Private=false`) and point `HelixContent` at *its* publish output instead of re-publishing.

That removes the second publish entirely (no duplicate ILC cost) and makes ordering deterministic instead of timing-dependent. If reusing the product publish dir as-is is awkward, even just the `ProjectReference` edge (without consuming the output) gets you most of the way: it sequences the test publish after the product publish so they no longer run concurrently, and the file lock can't happen.

We did this for the SDK's `dotnet-aot` AOT layout in dotnet/dotnet#6590 (added `_IsPublishing=true` + a `ProjectReference` to sequence the publish before the consumer). Related: dotnet/dotnet#6576 moved that same publish from a separate-process `dotnet publish` to an in-process `<MSBuild Targets="Publish">` so the BuildManager reuses cached project results instead of rebuilding/clobbering shared outputs — same family of problem, worth a skim.

**2. If you'd rather keep the test's own publish, fully isolate it.**
Override `IntermediateOutputPath` **and** `OutputPath` (in addition to `PublishDir`), not just the native subdir — then the two publishes share no writable path. The catch is this forces a separate restore for the test publish, and Arcade's centralized restore writes a RID-less `project.assets.json`, so you have to make sure the RID-specific target exists: set `RuntimeIdentifier` (singular — `RuntimeIdentifiers` plural does **not** create RID targets in the lock file). We worked through that one in dotnet/sdk#54222. It works, but it's more moving parts than the dependency-edge approach.

My vote is option 1 — it's the version that actually removes the race rather than narrowing it, and it's less to maintain.

Links:
- dotnet/dotnet#6590 — `_IsPublishing=true` + `ProjectReference` sequencing for the SDK AOT publish
- dotnet/dotnet#6576 — in-process `<MSBuild Targets="Publish">` instead of a separate publish process
- dotnet/sdk#54222 — the RID-less centralized-restore / `RuntimeIdentifier` (singular) gotcha

DamianEdwards: Implemented option 1 in d4ac4fbcda6:

- `aspnetcoretools.csproj` now publishes as part of its own build via a guarded `Publish` MSBuild call with `_IsPublishing=true`.
- The NativeAOT test project no longer runs a second publish. It has a sequence-only `ProjectReference` to `aspnetcoretools.csproj` and consumes the aggregate project's publish output from the product artifact path.

Local validation:
- `dotnet build .\src\aspnetcore\src\Tools\NativeAot\aspnetcoretools\aspnetcoretools.csproj -c Release -p:DotNetBuild=true -p:RuntimeIdentifier=win-x64 -p:TargetRuntimeIdentifier=win-x64 -p:ArtifactsBinDir=<fresh-root> -p:ArtifactsObjDir=<fresh-root> -bl:...`
- Confirmed the build runs `PublishAspNetCoreToolsOnBuild`, passes `_IsPublishing=true`, and produces the aggregate publish output under `artifacts\validation-aggregate\bin\aspnetcoretools\Release\net11.0\win-x64\publish\`.

The full ASP.NET Core repo build entry point still hits the pre-existing local `GenerateFiles`/`SkipUnchanged` task mismatch before reaching this change, so I used targeted project validation for this fix.

DamianEdwards: @dsplaisted ready for re-review

dsplaisted: @DamianEdwards This is better but I think it could still hit double-build problems, as the global `_IsPublishing` property creates a new build with the same output paths.

What we've done in other situations is to set `_IsPublishing` statically:

```xml
 <PropertyGroup>
   <_PublishWasInvokedDirectly Condition="'$(_IsPublishing)' == 'true'">true</_PublishWasInvokedDirectly>
   <_IsPublishing>true</_IsPublishing>            <!-- static, not a global property -->
 </PropertyGroup>
 
 <Target Name="PublishOnBuild"
         AfterTargets="Build"
         DependsOnTargets="Publish"               <!-- runs in the SAME instance -->
         Condition="'$(_PublishWasInvokedDirectly)' != 'true'" />
```

If you can do that, it would probably be better.  You would also need to always have a RuntimeIdentifier set for this to work, otherwise PublishAOT will fail.

dsplaisted: Also, I'm not sure I'm the right one to sign off on this overall.  I was just commenting on a piece of it that I had some insight into.

I am excited to see this land!

DamianEdwards: Addressed in 09155305f0f: `aspnetcoretools.csproj` now sets `_IsPublishing` statically, records direct publish via `_PublishWasInvokedDirectly`, and has `PublishAspNetCoreToolsOnBuild` depend on `Publish` so it runs in the same project instance rather than re-entering with global properties. It also defaults `RuntimeIdentifier` when NativeAOT publish is enabled.

Validation:
- `dotnet build .\src\aspnetcore\src\Tools\NativeAot\aspnetcoretools\aspnetcoretools.csproj -c Release -p:DotNetBuild=true -p:RuntimeIdentifier=win-x64 -p:TargetRuntimeIdentifier=win-x64 ...`
- `dotnet build .\src\aspnetcore\src\Tools\NativeAot\test\Microsoft.AspNetCore.Tools.NativeAot.Tests.csproj -c Release -p:DotNetBuild=true -p:RuntimeIdentifier=win-x64 -p:TargetRuntimeIdentifier=win-x64 ...`

The repo bootstrap still hits the pre-existing local `GenerateFileFromTemplate`/`SkipUnchanged` task mismatch, so I generated the missing build artifacts from the repo templates before running the focused builds.

DamianEdwards: Thx, updated.

DamianEdwards:
@lewing ready for your review

src/aspnetcore/src/Tools/NativeAot/aspnetcoretools/aspnetcoretools.csproj
```diff
@@ -0,0 +1,49 @@
...
+    <TargetFramework>$(DefaultNetCoreTargetFramework)</TargetFramework>
+    <OutputType>Library</OutputType>
+    <PackageId>aspnetcoretools</PackageId>
+    <PackAsTool>true</PackAsTool>
+    <IsShippingPackage>false</IsShippingPackage>
+    <ExcludeFromSourceOnlyBuild>true</ExcludeFromSourceOnlyBuild>
+    <PublishAot Condition="'$(DotNetBuildSourceOnly)' != 'true'">true</PublishAot>
+    <NativeLib>Shared</NativeLib>
+    <EnablePreviewFeatures>true</EnablePreviewFeatures>
+    <RuntimeIdentifiers Condition="'$(DotNetBuild)' != 'true' and '$(DotNetBuildSourceOnly)' != 'true'">$(BundledToolTargetRuntimeIdentifiers)</RuntimeIdentifiers>
```
wtgodbe, line 13: Can you add a comment here explaining why the RIDs are different per-scenario?

wtgodbe (approved):
aspnetcore part LGTM

src/aspnetcore/src/Tools/NativeAot/aspnetcoretools/aspnetcoretools.csproj
DamianEdwards: Will do a follow-up to dotnet/aspnetcore for this as I don't want to invalidate the green checks here!

DamianEdwards: https://github.com/dotnet/aspnetcore/issues/67229

lewing (commented):
Looks good me, you could do an official build if you wanted to be 100% sure

dsplaisted:
@DamianEdwards FYI I've added guidance to cover how to handle this to dotnet/skills: https://github.com/dotnet/skills/pull/770

DamianEdwards:
@wtgodbe can I get another approval after the merge to fix the conflict pls?

wtgodbe (approved):

PR #7167, opened 2026-06-10
Fix solaris build with gcc toolchain

am11:
It regressed in https://github.com/dotnet/dotnet/pull/6830. On illumos, the compiler _recognizes_ `retain` as a valid token but the actual codegen throws:

```
 [7/1783] Building CXX object Corehost.Static/hostmisc/CMakeFiles/hostmisc.dir/utils.cpp.o
  FAILED: Corehost.Static/hostmisc/CMakeFiles/hostmisc.dir/utils.cpp.o 
  /crossrootfs/x64/bin/x86_64-illumos-g++ --sysroot=/crossrootfs/x64 -DBUILDENV_DEBUG=1 -DDEBUG -DDISABLE_CONTRACTS -DHOST_64BIT -DHOST_AMD64 -DHOST_UNIX -DTARGET_64BIT -DTARGET_AMD64 -DTARGET_ILLUMOS -DTARGET_SUNOS -DTARGET_UNIX -DURTBLDENV_FRIENDLY=Debug -D_DBG -D_DEBUG -D_FILE_OFFSET_BITS=64 -D_POSIX_PTHREAD_SEMANTICS -D_REENTRANT -D_TIME_BITS=64 -D_XPG4_2 -D__EXTENSIONS__ -I/runtime/src/native -I/runtime/src/native/inc -I/runtime/src/native/corehost/apphost/static/.. -I/runtime/src/native/corehost/apphost/static/../.. -I/runtime/src/native/corehost/apphost/static/../../hostmisc -I/runtime/src/native/corehost/apphost/static/../../json -I/runtime/src/native/libs/System.IO.Compression.Native -I/runtime/src/native/libs/Common -I/runtime/artifacts/obj -I/runtime/artifacts/obj/coreclr/illumos.x64.Debug/Corehost.Static/hostmisc -I/runtime/src/native/corehost/hostmisc -I/runtime/artifacts/obj/coreclr/illumos.x64.Debug/inc/corehost -isystem /crossrootfs/x64/include -fstack-protector -g -std=gnu++17 -fPIC -O0 -Wall -g -fno-omit-frame-pointer -fno-strict-overflow -fno-strict-aliasing -fstack-protector-strong -Werror -Wno-unused-variable -Wno-unused-value -Wno-unused-function -Wno-tautological-compare -Wno-unknown-pragmas -Wimplicit-fallthrough -Wvla -Wno-invalid-offsetof -Wno-unused-but-set-variable -ffp-contract=off -fno-rtti -Wno-uninitialized -Wno-strict-aliasing -Wno-array-bounds -Wno-misleading-indentation -Wno-stringop-overflow -Wno-restrict -Wno-stringop-truncation -Wno-class-memaccess -faligned-new -fsigned-char -fvisibility=hidden -ffunction-sections -fdata-sections -MD -MT Corehost.Static/hostmisc/CMakeFiles/hostmisc.dir/utils.cpp.o -MF Corehost.Static/hostmisc/CMakeFiles/hostmisc.dir/utils.cpp.o.d -o Corehost.Static/hostmisc/CMakeFiles/hostmisc.dir/utils.cpp.o -c /runtime/src/native/corehost/hostmisc/utils.cpp
  In file included from /runtime/src/native/corehost/hostmisc/utils.cpp:10:
  /runtime/artifacts/obj/_version.c:1:49: error: 'retain' attribute ignored [-Werror=attributes]
      1 | static char sccsid[] __attribute__((used,retain)) = "@(#)Version 42.42.42.42424";
        |                                                 ^
  cc1plus: all warnings being treated as errors
  [8/1783] Building C object libs-native/System.IO.Compression.Native/CMakeFiles/System.IO.Compression.Native-Static.dir/pal_zlib.c.o
  [9/1783] Building C object _deps/fetchzlibng-build/CMakeFiles/zlib.dir/arch/generic/adler32_fold_c.c.o
  [10/1783] Building CXX object Corehost.Static/hostmisc/CMakeFiles/hostmisc.dir/pal.unix.cpp.o
```

The distinction that parser recognizes it and codegen doesn't was important because I used idiomatic `#if __has_attribute(retain)` but that branch was taken instead of `#else` on this system; therefore I ended up with clang||(gcc&&sunos) check.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.targets
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.targets
@@ -118,7 +118,14 @@
       -->
       <_NativeVersionFileContents>
 <![CDATA[
-static char sccsid[] __attribute__((used,retain)) = "@(#)Version $(FileVersion)$(_SourceBuildInfo)";
+#if defined(__clang__) || (defined(__GNUC__) && !defined(TARGET_SUNOS))
+#define RETAIN __attribute__((used, retain))
+#else
+#define RETAIN __attribute__((used))
+__asm__(".pushsection .init_array; .reloc ., R_X86_64_NONE, sccsid; .popsection");
+#endif
+
+static char sccsid[] RETAIN = "@(#)Version $(FileVersion)$(_SourceBuildInfo)";
  ]]>
       </_NativeVersionFileContents>
     </PropertyGroup>
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.targets
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.targets
@@ -118,11 +118,11 @@
       -->
       <_NativeVersionFileContents>
 <![CDATA[
-#if defined(__clang__) || (defined(__GNUC__) && !defined(TARGET_SUNOS))
-#define RETAIN __attribute__((used, retain))
-#else
+#if defined(__GNUC__) && !defined(__clang__) && defined(TARGET_SUNOS)
 #define RETAIN __attribute__((used))
 __asm__(".pushsection .init_array; .reloc ., R_X86_64_NONE, sccsid; .popsection");
+#else
+#define RETAIN __attribute__((used, retain))
 #endif
 
 static char sccsid[] RETAIN = "@(#)Version $(FileVersion)$(_SourceBuildInfo)";
--- a/src/deployment-tools/eng/native/build-commons.sh
+++ b/src/deployment-tools/eng/native/build-commons.sh
@@ -122,11 +122,11 @@ build_native()
         else
             # Generate the dummy version.c and runtime_version.h, but only if they didn't exist to make sure we don't trigger unnecessary rebuild
             __versionSourceContent=$(cat <<'EOF'
-#if defined(__clang__) || (defined(__GNUC__) && !defined(TARGET_SUNOS))
-#define RETAIN __attribute__((used, retain))
-#else
+#if defined(__GNUC__) && !defined(__clang__) && defined(TARGET_SUNOS)
 #define RETAIN __attribute__((used))
 __asm__(".pushsection .init_array; .reloc ., R_X86_64_NONE, sccsid; .popsection");
+#else
+#define RETAIN __attribute__((used, retain))
 #endif
 
 static char sccsid[] RETAIN = "@(#)No version information produced";
--- a/src/deployment-tools/eng/native/build-commons.sh
+++ b/src/deployment-tools/eng/native/build-commons.sh
@@ -121,11 +121,21 @@ build_native()
             fi
         else
             # Generate the dummy version.c and runtime_version.h, but only if they didn't exist to make sure we don't trigger unnecessary rebuild
-            __versionSourceLine="static char sccsid[] __attribute__((used,retain)) = \"@(#)No version information produced\";"
+            __versionSourceContent=$(cat <<'EOF'
+#if defined(__clang__) || (defined(__GNUC__) && !defined(TARGET_SUNOS))
+#define RETAIN __attribute__((used, retain))
+#else
+#define RETAIN __attribute__((used))
+__asm__(".pushsection .init_array; .reloc ., R_X86_64_NONE, sccsid; .popsection");
+#endif
+
+static char sccsid[] RETAIN = "@(#)No version information produced";
+EOF
+)
             if [[ -e "$__versionSourceFile" ]]; then
-                read existingVersionSourceLine < "$__versionSourceFile"
+                existingVersionSourceContent="$(cat "$__versionSourceFile")"
             fi
-            if [[ "$__versionSourceLine" != "$existingVersionSourceLine" ]]; then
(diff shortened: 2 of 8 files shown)


am11:
cc @janvorli, @akoeplinger

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.targets
```diff
@@ -118,7 +118,14 @@
       -->
       <_NativeVersionFileContents>
 <![CDATA[
-static char sccsid[] __attribute__((used,retain)) = "@(#)Version $(FileVersion)$(_SourceBuildInfo)";
+#if defined(__clang__) || (defined(__GNUC__) && !defined(TARGET_SUNOS))
```
akoeplinger: TARGET_SUNOS is probably only set by the runtime repo so might not be available in other repos (I guess you'll find out since it'll crash elsewhere then).

I'd personally also prefer having the workaround case first since it makes the condition easier to read:
```c
#if defined(__GNUC__) && defined(TARGET_SUNOS)
// workaround for compiler crash
#define RETAIN __attribute__((used))
__asm__(".pushsection .init_array; .reloc ., R_X86_64_NONE, sccsid; .popsection");
#else
#define RETAIN __attribute__((used, retain))
#endif
```

akoeplinger:
looking some more at the error this seems https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99587 and  https://github.com/seL4/seL4_libs/issues/97

should we just skip `retain` for gcc instead? doesn't seem specific to illumos to me?

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.targets
am11: We share eng/native infra with dotnet/diagnostics, so it sets there as well. To my knowledge, native stuff isn’t used elsewhere outside runtime and diagnostics repos. Also, clang has early bail, which we prefer for official builds.

I will update with your other point. Because GNUC is also set by clang, I will modify the condition a bit in a moment (need to test one theory with linker detection instead of compiler; it works with lld, the llvm linker but fails with Solaris linker, which is different than bfs/GNU linker as well; and we have a detection for that too in both repos).

am11:
We don't see any error on linux with gcc 16: e.g. https://dev.azure.com/dnceng-public/public/_build/results?buildId=1456513&view=logs&jobId=9d44ffa5-6dbb-5c44-43bb-984541de6129&j=9d44ffa5-6dbb-5c44-43bb-984541de6129&t=2c1dca5f-db26-547e-262f-8b403ec0d74c. Our validation pipeline would had failed.

akoeplinger:
so it might be fixed in some gcc version? we could probably check for that instead

am11:
I'm fixing the issue which is known atm. When we have a use-case for someone using gcc to build runtime and running into this issue, we can update the condition. Officially we use clang, and gcc is community supported toolchain (illumos is the only one "needing" it atm because llvm toolchain support isn't great there). It doesn't fail on linux with past six versions:

```sh
$ echo 'static char sccsid[] __attribute__((used,retain)) = "@(#)Version 42.42.42.42424"; int main() {}' | gcc-16 -xc -
$ echo 'static char sccsid[] __attribute__((used,retain)) = "@(#)Version 42.42.42.42424"; int main() {}' | gcc-15 -xc -
$ echo 'static char sccsid[] __attribute__((used,retain)) = "@(#)Version 42.42.42.42424"; int main() {}' | gcc-14 -xc -
$ echo 'static char sccsid[] __attribute__((used,retain)) = "@(#)Version 42.42.42.42424"; int main() {}' | gcc-13 -xc -
$ echo 'static char sccsid[] __attribute__((used,retain)) = "@(#)Version 42.42.42.42424"; int main() {}' | gcc-12 -xc -
$ echo 'static char sccsid[] __attribute__((used,retain)) = "@(#)Version 42.42.42.42424"; int main() {}' | gcc-11 -xc -
```
no warning. vs.:
```sh
$ /crossrootfs/x64/bin/x86_64-illumos-gcc -dumpversion
13.3.0

$ echo 'static char sccsid[] __attribute__((used,retain)) = "@(#)Version 42.42.42.42424"; int main() {}' | /crossrootfs/x64/bin/x86_64-illumos-gcc -xc - --sysroot=/crossrootfs/x64
<stdin>:1:1: warning: 'retain' attribute ignored [-Wattributes]
```

akoeplinger:
interesting, I wonder what is different in the x86_64-illumos-gcc

akoeplinger (approved):

am11:
It's basically the linker difference. Solaris/illumos brings its own linker and doesn't use GNU's bfd or LLVM's lld. I was able to build runtime with lld linker today and filed two issues for the workarounds I had to use: linker https://github.com/llvm/llvm-project/issues/202963 and assembler https://github.com/llvm/llvm-project/issues/202953. This was the first time we are able to use clang and lld like we use for other cross targets. The workarounds are self-contained in toolchain.cmake, but I still want to refine them before opening a PR in arcade. While doing that work, I also wanted to keep it working with gcc to maintain current state which is where I ran into this `retain` error. When building with clang, there is no problem as lld linker understands `retain`.

am11:
Good to merge? :)

janvorli (approved):

PR #6838, opened 2026-05-26
[EXPERIMENT] Stop forcing MSBuild nodeReuse=false on CI builds

JanProvaznik:
**EXPERIMENT — do not merge.** No reviewers pinged on purpose.

This branch removes every CI-side `nodeReuse=false` opt-out from the VMR's
build scripts so MSBuild's default (`nodeReuse=true`) applies on CI. The
intent is to exercise a "node reuse on everywhere" configuration on
dnceng-public and observe whether real source-build / unified-build pipelines
tolerate it, which custom tasks misbehave under reuse, and what the perf
impact looks like.

### Why this is an experiment, not a desirable change

- CI environments are short-lived; node reuse mostly buys persistent caches
  that benefit interactive/local builds, not single-shot CI runs.
- `nodeReuse=false` on CI has been the long-standing default for good
  reasons: build determinism, avoiding stale tasks across stages, and working
  around custom task assemblies that don't tolerate reuse.
- Re-enabling it broadly is likely to surface regressions in custom tasks,
  task host lifecycle, and locked-file scenarios on Windows.

### Scope (build infrastructure only)

Removed the `if (\) { ... \ = \False }` and
`if [[ \"\\" == true ]]; then node_reuse=false; fi` lines from:

- VMR root: `eng/build.ps1`, `eng/common/build.{ps1,sh}`, `eng/common/msbuild.{ps1,sh}`
- Arcade source: `src/arcade/eng/common/build.{ps1,sh}`, `src/arcade/eng/common/msbuild.{ps1,sh}`
- Every per-repo mirror of those Arcade scripts under `src/*/eng/common/`
- Repo-specific wrappers: `src/aspnetcore/eng/build.ps1`, `src/msbuild/eng/build.ps1`,
  `src/msbuild/eng/cibuild_bootstrapped_msbuild.{ps1,sh}`,
  `src/fsharp/eng/build.sh`, `src/nuget-client/eng/dotnet-build/build.{ps1,sh}`

Left intentionally untouched (product code or tests that disable reuse on purpose):

- `src/nuget-client/src/NuGet.Clients/NuGet.CommandLine/MsBuildUtility.cs`
- `src/{fsharp,roslyn}/eng/test-determinism.ps1`
- `src/nuget-client/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/DotnetIntegrationTestFixture.cs`
- `src/roslyn/src/Compilers/Core/MSBuildTaskTests/**`

### Diff summary

[...]

PR #7211, opened 2026-06-12
Add opt-in CI MSBuild node reuse via DotNetBuildNodeReuse property

JanProvaznik:
Makes MSBuild **node reuse opt-in** on CI via a property, instead of unconditionally removing the `nodeReuse=false` opt-outs (the original experiment in #6838). Follows the design of the existing `-mt` / `DotNetBuildMT` mechanism.

### What it does

- **`repo-projects/Directory.Build.props`** - when `/p:DotNetBuildNodeReuse=true` is passed (and a repo hasn't set `UseNodeReuse=false`), export `MSBUILD_NODEREUSE_ENABLED=1` to the inner repo build.
- **`src/arcade/eng/common` `build` / `msbuild` / `tools` scripts** - only force `nodeReuse=false` on CI when `MSBUILD_NODEREUSE_ENABLED != 1`, and gate the *"Node reuse must be disabled in CI build"* guard the same way.

Default CI behavior is **unchanged** (node reuse stays disabled) unless the property is passed:

``
./build.sh --ci ... /p:DotNetBuildNodeReuse=true
``

### Scope

Per review feedback on #6838, only the **outgoing Arcade source** is edited:

- `src/arcade/eng/common/**` is copied into the inner repo builds by the `UpdateEngCommonFiles` target (for every repo except the bootstrap-arcade ones), so the source-build product repos pick up the gating at build time.
- The root `/eng/common` copy is a **tooling file** (it must flow from `dotnet/arcade`; the VMR `ChangeValidation` step rejects manual edits to it), so it is intentionally **not** edited here. It will pick up the change once it flows from Arcade.
- No per-repo `src/*/eng/common` mirrors are touched (they are regenerated during the build).

> Note: because `src/arcade` is a mirrored mapping, these `eng/common` edits will back-flow to `dotnet/arcade` on merge.

### Validation

- The opt-in was exercised on the **source-build CI leg** (via `extraPropertiesStage2: /p:DotNetBuildNodeReuse=true`) in an earlier revision of this PR; the `SB_*` source-build stage-1 and stage-2 legs **passed** with node reuse enabled. That experiment commit was reverted to keep this PR focused/mergeable.
- `bash -n` passes on all modified `.sh` scripts; the PowerShell parser passes on all modified `.ps1` scripts.

### DIFF
--- a/src/arcade/eng/common/build.ps1
+++ b/src/arcade/eng/common/build.ps1
@@ -174,6 +174,7 @@ try {
       $binaryLog = $true
     }
     # Disable node reuse on CI unless explicitly opted in via MSBUILD_NODEREUSE_ENABLED.
+    # Internal testing only; this env var will be replaced with a switch (https://github.com/dotnet/arcade/issues/17013) and must not be depended on.
     if ($env:MSBUILD_NODEREUSE_ENABLED -ne "1") {
       $nodeReuse = $false
     }
--- a/src/arcade/eng/common/build.ps1
+++ b/src/arcade/eng/common/build.ps1
@@ -173,7 +173,10 @@ try {
     if (-not $excludeCIBinarylog) {
       $binaryLog = $true
     }
-    $nodeReuse = $false
+    # Disable node reuse on CI unless explicitly opted in via MSBUILD_NODEREUSE_ENABLED.
+    if ($env:MSBUILD_NODEREUSE_ENABLED -ne "1") {
+      $nodeReuse = $false
+    }
   }
 
   if (-not [string]::IsNullOrEmpty($binaryLogName)) {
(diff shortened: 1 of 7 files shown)


JanProvaznik:
This will be used in MSBuild VMR validation pipeline.

src/arcade/eng/common/build.ps1
```diff
@@ -173,7 +173,10 @@ try {
     if (-not $excludeCIBinarylog) {
       $binaryLog = $true
     }
-    $nodeReuse = $false
+    # Disable node reuse on CI unless explicitly opted in via MSBUILD_NODEREUSE_ENABLED.
```
ViktorHofer, line 176: I would mention the follow-up issue in these comments as well (in all 6 files) to make sure that callers understand that the env var will eventually be replaced with a switch and shouldn't be depended on.

ViktorHofer (approved):

PR #7218, opened 2026-06-12
Use direct NuGet.Frameworks reference on .NET Core

rainersigwald:
Port of dotnet/msbuild#14047 into the VMR.

### Changes

On .NET Core, `Microsoft.Build` takes a direct compile-time `PackageReference` to `NuGet.Frameworks` (`PrivateAssets="all"`, so no new public package dependency) and calls it through normal method calls instead of reflection — avoiding the per-call `MethodInfo`/`PropertyInfo` overhead in `NuGetFrameworkWrapper`. At runtime MSBuild still binds to whatever `NuGet.Frameworks.dll` the host SDK ships.

The `#if NETFRAMEWORK` path is unchanged in behavior: it still loads `NuGet.Frameworks` into a secondary AppDomain with synthesized binding redirects so the NGEN native image is used. Shared logic (`GetNonZeroVersionParts`, plus `FilterTargetFrameworks` via a struct adapter constrained on `ITfmAdapter<TParsed>`) lives in a new partial, and the two implementations are split into `NuGetFrameworkWrapper.Direct.cs` and `NuGetFrameworkWrapper.Reflection.cs`.

### Notes

- Applied via `darc vmr cherry-pick` of the squashed PR diff, so the `src/msbuild/` content matches the upstream PR exactly (6 files, +349/−223).
- This is a preview port; the change will land normally via codeflow once dotnet/msbuild#14047 merges. `source-manifest.json` is intentionally left untouched.
- Not built locally (full VMR build); relying on CI for validation.

### DIFF
--- a/src/msbuild/documentation/NETFramework-NGEN.md
+++ b/src/msbuild/documentation/NETFramework-NGEN.md
@@ -94,7 +94,8 @@ When this happens, the cost of JITting `Microsoft.Build.NuGetSdkResolver` will b
 ## NuGet.Frameworks
 
 When evaluating certain property functions, MSBuild requires functionality from `NuGet.Frameworks.dll`, which is not part of MSBuild proper.
-The assembly is loaded lazily from a path calculated based on the environment where MSBuild is running and the functionality is invoked
+
+On .NET Framework, the assembly is loaded lazily from a path calculated based on the environment where MSBuild is running and the functionality is invoked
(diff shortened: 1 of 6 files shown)


ViktorHofer:
Ok this automatically uses the N-1 version thanks to the Version.Details.xml entry in SB. Thanks for double checking with this validation PR.

PR #7278, opened 2026-06-17
Set file ownership of Debian package content

NikolaMilosavljevic:
Explicitly sets the owner and group attributes.


joeloff (approved):

akoeplinger:
Note for future readers: this already landed in 10.0.1xx, this just ports it to main.

PR #7300, opened 2026-06-19
Enable AOT-compiled SDK tools in source-build

### LINKED ISSUE #5567: Enable AOT-compiled SDK tools when the SDK is built via source-build
## Motivation

We want to ship AOT-compiled SDK tools as part of the .NET SDK. The first concrete example is `dotnet-user-secrets`, where AOT-enablement is being applied in dotnet/aspnetcore#66663.

For this to be viable, AOT publishing needs to succeed in every build of the SDK we ship, including the VMR's source-build legs and downstream distros' source-build runs. Today, it doesn't. Surface symptoms depending on configuration:

- **`PublishAot=true` with the default portable `RuntimeIdentifiers` list**: restore fails with `NU1102: Unable to find package Microsoft.NETCore.App.Runtime.linux-x64`. Source-build doesn't publish runtime packs under portable RID names.
- **Scope `RuntimeIdentifiers` to the build's non-portable RID via `$(TargetRuntimeIdentifier)`**: restore succeeds, then the SDK rejects the project with `NETSDK1203: Ahead-of-time compilation is not supported for the target runtime identifier 'centos.10-x64'`.

## What's blocking this from working

Every repo's build in the VMR is invoked against a previous SDK version. In a stage 1 source-build leg that SDK is the Microsoft-built portable SDK, which has none of the configuration the source-built SDK output ends up with:

- Its `Microsoft.NETCoreSdk.BundledVersions.props` does not include the build's non-portable RID in `ILCompilerRuntimeIdentifiers` or `RuntimePackRuntimeIdentifiers`.
- Its `packs/` folder does not contain the NativeAOT runtime pack for the build's RID.

So, for example, when `aspnetcore` builds `dotnet-user-secrets` with `PublishAot=true` for the build's non-portable RID, the previous SDK's `ProcessFrameworkReferences` task validates the target against its `KnownILCompilerPack.ILCompilerRuntimeIdentifiers` using `PortableRuntimeIdentifierGraph.json` (no non-portable entries), fails the lookup, and emits `NETSDK1203`. The source-built SDK that this build will eventually produce would accept the same target, but it doesn't exist yet at the point where `aspnetcore` is being built.

The stage 2 leg is the inverse: it uses stage 1's SDK output as its dependent SDK, so the AOT configuration and bundled AOT pack are present and AOT-publishing `dotnet-user-secrets` against it would succeed.

mthalman:
The SDK bundles three ASP.NET Core CLI tools (dotnet-dev-certs, dotnet-user-jwts, dotnet-user-secrets) as a Native AOT aggregate, but AOT publishing of these tools only worked in the Microsoft build. In source-build it failed in two ways: restore could not find a portable-RID runtime pack (`NU1102`), and once scoped to the build's non-portable RID the previous SDK rejected the project with `NETSDK1203` ("Ahead-of-time compilation is not supported for the target runtime identifier"). The root cause is that every repo in the VMR builds against a previous SDK; in a stage 1 source-build leg that SDK is the Microsoft portable SDK, whose bundled props do not list the build's non-portable RID (e.g. `centos.10-x64`) in the ILCompiler/runtime-pack RID sets and whose packs/ folder has no NativeAOT pack for that RID.

Teach the previous SDK about the source-built non-portable RID and unblock AOT publishing, then ship the resulting aggregate in source-build. Fixes dotnet/source-build#5567.

Also dependent on https://github.com/dotnet/dotnet/pull/7297

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
@@ -58,4 +58,20 @@
     <Copyright>$(CopyrightMicrosoft)</Copyright>
   </PropertyGroup>
 
+  <!--
+    In source-build, the SDK ships native binaries with embedded debug symbols (no separate
+    symbol files and no .gnu_debuglink); this is enforced for every native ELF in the SDK by the
+    source-build native-symbol validation. Native AOT defaults StripSymbols to true on non-Windows,
+    which strips its output into a separate file and adds a .gnu_debuglink section. Disable stripping
+    for any project that publishes with Native AOT so the produced binary retains embedded DWARF.
+    This is the AOT analog of the KeepNativeSymbols=true that source-build uses for native
+    (CMake-built) binaries. Scoped to PublishAot projects so non-AOT builds are unaffected, and only
+    applies in source-build (DotNetBuildSourceOnly); Microsoft builds still strip and ship separate
+    symbols. Imported via Sdk.targets (after the project sets PublishAot) so the condition is known.
+    See https://github.com/dotnet/source-build/issues/5567.
+  -->
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(PublishAot)' == 'true'">
+    <StripSymbols>false</StripSymbols>
+  </PropertyGroup>
+
 </Project>
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
@@ -59,19 +59,16 @@
   </PropertyGroup>
 
   <!--
-    In source-build, the SDK ships native binaries with embedded debug symbols (no separate
-    symbol files and no .gnu_debuglink); this is enforced for every native ELF in the SDK by the
-    source-build native-symbol validation. Native AOT defaults StripSymbols to true on non-Windows,
-    which strips its output into a separate file and adds a .gnu_debuglink section. Disable stripping
-    for any project that publishes with Native AOT so the produced binary retains embedded DWARF.
-    This is the AOT analog of the KeepNativeSymbols=true that source-build uses for native
-    (CMake-built) binaries. Scoped to PublishAot projects so non-AOT builds are unaffected, and only
-    applies in source-build (DotNetBuildSourceOnly); Microsoft builds still strip and ship separate
-    symbols. Imported via Sdk.targets (after the project sets PublishAot) so the condition is known.
-    See https://github.com/dotnet/source-build/issues/5567.
+    Source-build does not strip symbols from native binaries during the build; the binaries are
+    produced with embedded symbols and the distro packaging process is responsible for stripping
+    the symbols out of the shipping binaries and archiving those symbols separately. Native AOT
+    defaults StripSymbols to true on non-Windows, which would strip its output during the build;
+    align StripSymbols with KeepNativeSymbols so AOT-published binaries keep their symbols like
+    other native (CMake-built) binaries in source-build.
   -->
-  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(PublishAot)' == 'true'">
-    <StripSymbols>false</StripSymbols>
+  <PropertyGroup>
+    <KeepNativeSymbols Condition="'$(KeepNativeSymbols)' == '' and '$(DotNetBuildSourceOnly)' == 'true'">true</KeepNativeSymbols>
+    <StripSymbols Condition="'$(KeepNativeSymbols)' == 'true'">false</StripSymbols>
   </PropertyGroup>
 
 </Project>
--- a/src/runtime/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
+++ b/src/runtime/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
@@ -9,6 +9,15 @@
 
     <_originalTargetOS>$(RuntimeIdentifier.SubString(0, $(RuntimeIdentifier.LastIndexOf('-'))))</_originalTargetOS>
     <_originalTargetOS Condition="'$(_targetsNonPortableSdkRid)' == 'true'">$(NETCoreSdkPortableRuntimeIdentifier.SubString(0, $(NETCoreSdkPortableRuntimeIdentifier.LastIndexOf('-'))))</_originalTargetOS>
+    <!--
+      In a source-build leg the bootstrap SDK may be portable (so $(NETCoreSdkRuntimeIdentifier) is the
+      portable RID), yet we build for the non-portable target RID ($(TargetRid), e.g. centos.10-x64).
+      The SDK already resolves the host ILCompiler pack and the target NativeAOT runtime pack correctly
+      for that case, but the target OS derived from the RID (e.g. 'centos') is not a valid ILC target OS.
+      Map it to the portable base OS (e.g. 'linux'), mirroring the non-portable SDK RID handling above,
+      without altering the host/target pack names (those must match what the SDK resolved).
+    -->
+    <_originalTargetOS Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(RuntimeIdentifier)' == '$(TargetRid)' and '$(RuntimeIdentifier)' != '$(NETCoreSdkPortableRuntimeIdentifier)'">$(NETCoreSdkPortableRuntimeIdentifier.SubString(0, $(NETCoreSdkPortableRuntimeIdentifier.LastIndexOf('-'))))</_originalTargetOS>
     <_originalTargetOS Condition="$(_originalTargetOS.Contains('.'))">$(_originalTargetOS.SubString(0, $(_originalTargetOS.IndexOf('.'))))</_originalTargetOS>
     <_originalTargetOS Condition="$(_originalTargetOS.StartsWith('win'))">win</_originalTargetOS>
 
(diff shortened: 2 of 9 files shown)


mthalman:
Code reviewers - PTAL

src/runtime/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
```diff
@@ -9,6 +9,15 @@
...
     <_originalTargetOS Condition="'$(_targetsNonPortableSdkRid)' == 'true'">$(NETCoreSdkPortableRuntimeIdentifier.SubString(0, $(NETCoreSdkPortableRuntimeIdentifier.LastIndexOf('-'))))</_originalTargetOS>
+    <!--
+      In a source-build leg the bootstrap SDK may be portable (so $(NETCoreSdkRuntimeIdentifier) is the
+      portable RID), yet we build for the non-portable target RID ($(TargetRid), e.g. centos.10-x64).
+      The SDK already resolves the host ILCompiler pack and the target NativeAOT runtime pack correctly
+      for that case, but the target OS derived from the RID (e.g. 'centos') is not a valid ILC target OS.
+      Map it to the portable base OS (e.g. 'linux'), mirroring the non-portable SDK RID handling above,
+      without altering the host/target pack names (those must match what the SDK resolved).
+    -->
+    <_originalTargetOS Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(RuntimeIdentifier)' == '$(TargetRid)' and '$(RuntimeIdentifier)' != '$(NETCoreSdkPortableRuntimeIdentifier)'">$(NETCoreSdkPortableRuntimeIdentifier.SubString(0, $(NETCoreSdkPortableRuntimeIdentifier.LastIndexOf('-'))))</_originalTargetOS>
```
akoeplinger: this file is shipping in the ILCompiler nuget package and gets used during end-user builds so it feels weird to have VMR-specific properties here.

/cc @MichalStrehovsky

akoeplinger:
LGTM apart from one comment

src/runtime/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
mthalman: You're right. I intentionally wanted to avoid dependencies on VMR-properties to shipping MSBuild files. I missed this one.

mthalman: Fixed in https://github.com/dotnet/dotnet/pull/7300/commits/5898fb74b87027fb2f647682ebafcc6ffdf6453d

akoeplinger (approved):

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
```diff
@@ -58,4 +58,20 @@
...
+    which strips its output into a separate file and adds a .gnu_debuglink section. Disable stripping
+    for any project that publishes with Native AOT so the produced binary retains embedded DWARF.
+    This is the AOT analog of the KeepNativeSymbols=true that source-build uses for native
+    (CMake-built) binaries. Scoped to PublishAot projects so non-AOT builds are unaffected, and only
+    applies in source-build (DotNetBuildSourceOnly); Microsoft builds still strip and ship separate
+    symbols. Imported via Sdk.targets (after the project sets PublishAot) so the condition is known.
+    See https://github.com/dotnet/source-build/issues/5567.
+  -->
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(PublishAot)' == 'true'">
+    <StripSymbols>false</StripSymbols>
```
am11: Do we need debug symbol as in "if end user wanted to debug the binary with native debugger"? It has quite a significant impact on binary size and this is often considered non-production scenario. Note that the stripped binaries also print exceptions and native symbols are useless for 99.9% of use-cases.

IMO, for this (probably-never-needed-in-prod) use-case we shouldn't take this approach and either:

* publish debug symbols package separately (where all symstore native symbols are published) and ask user to pull the symbols package as needed.
* or ask the user to build that git tag and debug the app locally.

cc @jkotas, @MichalStrehovsky

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
```diff
@@ -58,4 +58,20 @@
     <Copyright>$(CopyrightMicrosoft)</Copyright>
   </PropertyGroup>
 
+  <!--
+    In source-build, the SDK ships native binaries with embedded debug symbols (no separate
```
jkotas: This is not accurate. The actual shipping native binaries when source-build is integrated in the distro build process are expected to have the symbols stripped. This stripping happens as part of the distro packaging process and archives the stripped symbols in the distro symbol store or packages.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
```diff
@@ -58,4 +58,20 @@
...
+    source-build native-symbol validation. Native AOT defaults StripSymbols to true on non-Windows,
+    which strips its output into a separate file and adds a .gnu_debuglink section. Disable stripping
+    for any project that publishes with Native AOT so the produced binary retains embedded DWARF.
+    This is the AOT analog of the KeepNativeSymbols=true that source-build uses for native
+    (CMake-built) binaries. Scoped to PublishAot projects so non-AOT builds are unaffected, and only
+    applies in source-build (DotNetBuildSourceOnly); Microsoft builds still strip and ship separate
+    symbols. Imported via Sdk.targets (after the project sets PublishAot) so the condition is known.
+    See https://github.com/dotnet/source-build/issues/5567.
+  -->
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true' and '$(PublishAot)' == 'true'">
```
jkotas: Is the `KeepNativeSymbols` property available here? I think it would be better to control this behavior by a single property so that it can be overriden, instead of the duplicating the current source build policy here.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
jkotas: > publish debug symbols package separately (where all symstore native symbols are published) and ask user to pull the symbols package as needed.

This is what happens today (see my comment above).

Default `DotNetBuildSourceOnly` behaviors are designed around Ubuntu and Red Hat requirements. Producing binaries without stripped symbols is one of them. I think it may make sense to allow overriding this behavior centrally with a `KeepNativeSymbols` property.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
```diff
@@ -58,4 +58,20 @@
...
   </PropertyGroup>
 
+  <!--
+    In source-build, the SDK ships native binaries with embedded debug symbols (no separate
+    symbol files and no .gnu_debuglink); this is enforced for every native ELF in the SDK by the
+    source-build native-symbol validation. Native AOT defaults StripSymbols to true on non-Windows,
+    which strips its output into a separate file and adds a .gnu_debuglink section. Disable stripping
+    for any project that publishes with Native AOT so the produced binary retains embedded DWARF.
+    This is the AOT analog of the KeepNativeSymbols=true that source-build uses for native
+    (CMake-built) binaries. Scoped to PublishAot projects so non-AOT builds are unaffected, and only
```
jkotas: > Scoped to PublishAot projects so non-AOT builds are unaffected,

This is unnecessary. This property does nothing for non-PublishAot projects. It can be set all the time with no harm.

AI likes to generate logic that is more complicated than necessary...

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
```diff
@@ -58,4 +58,20 @@
...
+    applies in source-build (DotNetBuildSourceOnly); Microsoft builds still strip and ship separate
+    symbols. Imported via Sdk.targets (after the project sets PublishAot) so the condition is known.
+    See https://github.com/dotnet/source-build/issues/5567.
```
jkotas:
```suggestion
```
The comment explains the problem more than well. The issue link is unnecessary.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
mthalman: Corrected -- the comment now says the distro packaging process strips the symbols out of the shipping binaries and archives them separately, instead of implying the SDK ships them with embedded symbols. Fixed in cda18344712.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
mthalman: Done -- `StripSymbols` now derives from `KeepNativeSymbols` instead of re-deriving the source-build policy from `DotNetBuildSourceOnly`. Since `KeepNativeSymbols` is otherwise only defaulted within the runtime build, I default it to `true` for source-build here so it's in scope for repos like aspnetcore that publish the AOT tools. Fixed in cda18344712.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
mthalman: Removed the `PublishAot` condition. Fixed in cda18344712.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
mthalman: Removed the issue link. Fixed in cda18344712.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/ProjectDefaults.targets
mthalman: This now keys off `KeepNativeSymbols`, which is overridable, so the keep-symbols behavior can be controlled centrally rather than being hard-coded here. Fixed in cda18344712.

jkotas (approved):

PR #7308, opened 2026-06-22
[release/10.0.4xx] Use product-repo SHA instead of VMR SHA in MSBuild banner and SDK --info

AlesProkop:
VMR (`release/10.0.4xx`) port of dotnet/msbuild#14073 and dotnet/sdk#54786, opened here per the suggestion to land it directly in the VMR release branch so product-construction and other teams can weigh in (other repos may want the same).

### Context
When MSBuild and the SDK are built in the VMR (dotnet/dotnet), the commit embedded via SourceLink (`SourceRevisionId`) is the **VMR** commit, not the dotnet/msbuild or dotnet/sdk commit. That commit is what the MSBuild banner, `ProjectCollection.DisplayVersion` (and `$(MSBuildSemanticVersion)` / telemetry), and `dotnet --info` report — so it cannot be traced back to the product-repo source.

The VMR already passes the original product-repo commit to the per-repo build as `RepoOriginalSourceRevisionId`. This change surfaces that commit as the primary SHA and keeps the VMR commit as secondary info.

### Changes Made
**MSBuild** (`src/msbuild`, port of dotnet/msbuild#14073):
- `Microsoft.Build.Framework.csproj`: emit `[assembly: AssemblyMetadata("RepoOriginalSourceRevisionId", ...)]` when `RepoOriginalSourceRevisionId` is set (VMR builds).
- `ProjectCollection.DisplayVersion`: prefer the MSBuild commit from `RepoOriginalSourceRevisionId` when present; otherwise fall back to existing behavior. Fixes the banner, `$(MSBuildSemanticVersion)`, and telemetry in one place.
- Startup banner (`XMake.cs`): when the VMR commit differs from the MSBuild commit, show it as secondary — `MSBuild version <ver>+<msbuildSha> (VMR <vmrSha>) for <fw>`. Adds the `MSBuildVersionMessageWithVmr` resource string (+ xlf).

**SDK** (`src/sdk`, port of dotnet/sdk#54786):
- `GenerateLayout.targets`: the commit written to `.version` / `.toolsetversion` prefers `RepoOriginalSourceRevisionId`, falling back to `SourceRevisionId`. In VMR builds the VMR commit is appended to `.version` as a trailing line.
- `DotnetVersionFile.cs`: parse the optional trailing line into a new `VmrCommitSha` property.
- `CommandLineInfo.cs`: `dotnet --info` prints a secondary `VMR Commit:` line when it differs from `Commit:`. *(The upstream PR edits `ParserOptionActions.cs`; this branch predates that refactor, so the change is ported into `CommandLineInfo.cs`, which holds the equivalent `--info` logic here.)*

### Notes
- `InformationalVersion` / NuGet package metadata intentionally keep the VMR commit (how the binary was physically built). Only the user-facing "product" version now leads with the product-repo commit.
- Standalone/dev builds are byte-for-byte unchanged (no `RepoOriginalSourceRevisionId` => identical fallback path).
- Tests from both source PRs are included.

### Screenshots
**Before**
`dotnet --info`
image

**After**
`dotnet --info`
image

**Simulated VMR build**
`MSBuild.exe -version`
image

**Product version in Details of DLL**
image

Original PRs: dotnet/msbuild#14073, dotnet/sdk#54786.


JanProvaznik:
@AlesProkop please attach screenshots how this looks before and after (you can get dotnet from artifacts of CI here)

JanProvaznik (approved):
this is helpful for msbuild investigations, other teams might find it useful too, hence we're posting this here for consideration

AlesProkop:
Updated the description to include screenshots of how it looks before/after.

JanProvaznik (approved):

akoeplinger:
cc @baronfel for the `dotnet --info` output changes

baronfel:
Might be nice to have a helper for getting the 'effective version' that other tools could use, like the aspnetcore binaries, etc.

But overall this kind of disambiguation makes sense to have IMO.

src/sdk/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
```diff
@@ -70,6 +70,21 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+         product source commit is recoverable from AssemblyInformationalVersion (e.g. the "Product version"
+         field shown in the Windows file properties dialog). The VMR commit is preserved as the primary SHA,
+         and standalone builds (where RepoOriginalSourceRevisionId is empty) are unchanged. -->
+    <PropertyGroup Condition="'$(RepoOriginalSourceRevisionId)' != '' and '$(RepoOriginalSourceRevisionId)' != '$(SourceRevisionId)'">
+      <!-- Follow SemVer 2.0 rules -->
+      <_InformationalVersionContainsPlus>false</_InformationalVersionContainsPlus>
+      <_InformationalVersionContainsPlus Condition="$(InformationalVersion.Contains('+'))">true</_InformationalVersionContainsPlus>
+
+      <InformationalVersion Condition="!$(_InformationalVersionContainsPlus)">$(InformationalVersion)+$(RepoOriginalSourceRevisionId)</InformationalVersion>
+      <InformationalVersion Condition="$(_InformationalVersionContainsPlus)">$(InformationalVersion).$(RepoOriginalSourceRevisionId)</InformationalVersion>
```
rainersigwald: This doesn't seem like the right place to put this logic--generally this file is user-facing, and concepts like the VMR don't apply. Can we instead modify `$(InformationalVersion)` or `$(SourceRevisionId)` in Arcade or MSBuild-repo logic?

src/sdk/src/Cli/dotnet/CommandLineInfo.cs
```diff
@@ -24,6 +24,10 @@ public static void PrintInfo()
         Reporter.Output.WriteLine($"{LocalizableStrings.DotNetSdkInfoLabel}");
         Reporter.Output.WriteLine($" Version:           {Product.Version}");
         Reporter.Output.WriteLine($" Commit:            {commitSha}");
+        if (!string.IsNullOrEmpty(versionFile.VmrCommitSha) && versionFile.VmrCommitSha != versionFile.CommitSha)
+        {
+            Reporter.Output.WriteLine($" VMR Commit:        {versionFile.VmrCommitSha}");
```
rainersigwald: Exposing the "VMR" abbreviation to users like this feels a bit hostile, but I'm not sure I have a better name to give out . . .

src/msbuild/src/MSBuild/XMake.cs
```diff
@@ -4343,8 +4343,59 @@ private static void DisplayVersionMessageIfNeeded(bool recursing, bool useTermin
...
+
+        /// <summary>
+        /// Returns the VMR (build) commit when it differs from the MSBuild commit shown by
+        /// <see cref="ProjectCollection.DisplayVersion"/>; otherwise null. In VMR builds the assembly
+        /// InformationalVersion carries the VMR commit while DisplayVersion carries the MSBuild commit.
+        /// </summary>
+        private static string GetVmrCommitForLogo()
+        {
+            string informationalVersion = typeof(ProjectCollection).Assembly
+                .GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
```
rainersigwald: Does this have to be duplicated here and in Microsoft.Build?

AlesProkop: There is an option on how to not have it as duplicate - Public API. But I think that in this case it is not worth it, as the duplication is quite small.

src/sdk/src/Cli/dotnet/CommandLineInfo.cs
AlesProkop: Maybe we could use "Source commit:"?

src/sdk/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets
AlesProkop: You are right. I added it into Arcade `Version.targets`

src/sdk/src/Cli/dotnet/CommandLineInfo.cs
JanProvaznik: dotnet/dotnet or VMR . Source commit is very generic

ViktorHofer (changes requested):
I don't agree with this change. The msbuild isolated repo commit SHA isn't representative. The shipping code is the one that lives in dotnet/dotnet. The msbuild repository is the development platform, not where the build lives, builds and ships. The code can diverge (we often have such cases, especially close to a release when changes are being made directly in the VMR).

Happy to discuss this further. cc @mmitche

mmitche:
> I don't agree with this change. The msbuild isolated repo commit SHA isn't representative. The shipping code is the one that lives in dotnet/dotnet. The msbuild repository is the development platform, not where the build lives, builds and ships. The code can diverge (we often have such cases, especially close to a release when changes are being made directly in the VMR).
> 
> Happy to discuss this further. cc @mmitche

I agree with @ViktorHofer but also agree that that it is useful to have repo source info in some fashion. A few things jump out to me as possibilities:
- The current commit info assumes the users knows where the commit lives. This was fine when users knew that SDK came from dotnet/sdk and msbuild from dotnet/msbuild but it's not as obvious now. Maybe we should be clarifying the source commit location.
- Exposing VMR and repo SHA as a top-level concept gives me the ick. It seems to expose too many underlying details of the development process. But I get it. It's useful for investigations.
- The current --info is mostly unstructured and not uniform as currently implemented (e.g. why msbuild but not NuGet or roslyn?).

Here's what I think I would do:

1. Clarify shas in info with the source repository (dotnet/dotnet).
2. Stick with VMR shas as the primary sha.
3. Change the info to list a more complete set of components, with name, version, SHA (vmr sha), and last forward flow sha as fields.

JanProvaznik:
> seems to expose too many underlying details of the development process

I would argue that's the point for agentic chasing down of regressions/changes

AlesProkop:
# Addressing discussion. New implementation.

Based on the discussion above, I changed the goal of this PR and made changes. Please feel free to review. The primary SHA now is the VMR commit and I implemented the table for each component.

`dotnet --info` renders like this
image


For the MSBuild side, we get this
image

Please, look at it and review, if possible @ViktorHofer, @mmitche. 
Matt, you mentioned that you would like version. The `source-manifest.json` from where we take the information about components does not include the version of the components, only the forward-flow SHA. So getting version would be difficult. Do you think, the SHA is enough or should I implement the version somehow as well?

ViktorHofer:
When @mmitche mentioned to display a list of components he likely didn't mean to list the name of the repositories. arcade, scenario-tests and source-build-assets are non-shipping repositories that don't offer any components to customers. Customers also shouldn't need to know what "vstest" (the repository name) means.

We definitely need some PM input and sdk input here. @marcpopMSFT & @baronfel

marcpopMSFT:
The version of SDK and MSBuild that we're shipping is the VMR commit. That build is the version we flow to other repos as well. The VMR build commit can be used to follow back to other repos.

Let's take a step back and think about what this commit is likely used for. It seems the primary use is customers sharing with us and us following back to the code. I think if we had a `dotnet info -v:diag`, this would make sense there but in a plain `dotnet --info` call, it's too much detail about our build system that shouldn't be front and center to the customer.

I think the current behavior is fine and we can just update our skills/guidance to make it easier for folks to follow the vmr commit back to the individual repo personally.

Did we want to bring back the `dotnet info` command?

AlesProkop:
As per comments above, this change is not a clear improvement. Closing the PR

PR #7309, opened 2026-06-22
Sync eng/common/cross,native across consumer repos

am11:
Changes keep getting lost so I'm manually syncing them with dotnet/arcade's current state. `runtime`, `diagnostics`, `sdk` and `aspnetcore` are using cross and native dirs; they should be kept in sync for native/nativeaot builds especially on the new platforms (whose configs we added in past few months)


am11:
cc @ViktorHofer, @akoeplinger

mmitche:
@am11 Where are you seeing the results of an out-of-sync eng/common?

- When building in the VMR, arcade's eng/common is copied over the eng common of all repos that depend on arcade (which is essentially every repo).
- On backflow, these repos (most? I think except diagnostics) will get a new arcade, and new eng/common. Then on forward flow the VMR gets that updated eng/common. However, as noted, that wouldn't affect the VMR build.

mmitche (changes requested):
We wouldn't typically alter eng/common directly in any repo as changes get overwritten by arcade updates. Need to understand the symptom.

am11:
@mmitche, the `eng/common/native` and `eng/common/cross` directories function as shared directories for native build files. This way, when we add a new platform or update a cross-compilation configuration, we do not have to duplicate those changes in the diagnostics repo. This was the original intent. Well in 2018, the original intent was to share them with core-setup repo, but that (as you know) was merged to runtime repo (as installer/corehost). Up until recently, the consumers of these dirs were runtime, diagnostics and dotnet-buildtools-prereqs-docker repos.

More recently, `PublishAot` is being adopted by other repositories (such as `sdk` and `aspnetcore`), we hoisted some more configs from runtime to arcade, to make cross-builds works across repos with our prereqs container infra.

> Where are you seeing the results of an out-of-sync eng/common?

Right now; in runtime, diagnostics, sdk and aspnetcore.. basically everywhere these two dirs are used, code is currently out of sync for one reason or the other. The source of truth ATM is arcade's eng/common/{cross,native} where I copied it from.

> When building in the VMR, arcade's eng/common is copied over the eng common of all repos that depend on arcade (which is essentially every repo).

Few days ago, I was told that it's src/arcade/eng/common copied to other repos' eng/common during the sync; but if you are generally understanding/agreeing with the motivation here, then I can update VMR's eng/common as well. Secondly, somehow `native` and `cross` are being completely omitted by auto sync in some repos: https://github.com/dotnet/sdk/pull/54762#issuecomment-4753671250.

mmitche:
The VMR's eng/common should not be updated. That version needs to match the bootstrap arcade that is in use. Otherwise, it looks like the primary issue is that those two dirs are not getting updated (from src/arcade/eng/common/*) to the consumer repos on backflow. Let's figure out that issue. Can you open an issue in arcade-services if you haven't? /cc @dkurepa.

Changing the eng/common in this individual repos wont' have any effect in the VMR build, and if the sync process is broken in some fashion, may just result in those files getting overwritten on the next backflow. Generally, we try to avoid any kind manual edit to eng/common.

am11:
Opened https://github.com/dotnet/arcade-services/issues/6436.

mmitche:
> Opened [dotnet/arcade-services#6436](https://github.com/dotnet/arcade-services/issues/6436).

Thanks!

PR #7311, opened 2026-06-22
[no review, do not merge, for validation] Make restore process state resettable + run RestoreTask in the main process (retire the #13315 transient-TaskHost workaround)

JanProvaznik:
## Summary

Make `dotnet restore` safe to run in a **reused** MSBuild process (multi-threaded MSBuild / MSBuild Server), by two complementary changes that are only fully correct **together**:

1. **`src/nuget-client`** — a generic keyed reset registry, `NuGet.Common.NuGetProcessState`, that resets restore's process-wide static state at the start and end of every restore (port of [JanProvaznik/NuGet.Client#10](https://github.com/JanProvaznik/NuGet.Client/pull/10)).
2. **`src/msbuild`** — route `NuGet.Build.Tasks.RestoreTask` to the **main process** in multi-threaded / long-lived-host mode, **retiring** the transient-TaskHost workaround for [dotnet/msbuild#13315](https://github.com/dotnet/msbuild/issues/13315).

## Background: the problem and the existing workaround

`dotnet restore` runs in a process that exits after the build, so process exit reclaims everything restore touches: environment/config-derived caches, plugin child processes, the HTTP throttle, proxy and credential services. In a long-lived host the main process survives across builds, so that state **leaks**: stale caches, leaked plugin processes, and credential/proxy services pinned with the first build's settings (a correctness and security concern).

MSBuild currently works around this in `TaskRouter`/`AssemblyTaskFactory`: `RestoreTask` is forced into a **transient TaskHost** that terminates after each invocation, so process exit still reclaims its static state. That comment calls it *"a temporary workaround until the task authors fix their static state issues"* ([dotnet/msbuild#13315](https://github.com/dotnet/msbuild/issues/13315)). **This PR is that fix** — so the workaround is removed.

## Analysis of the static-state work (and why this fixes all of it)

A whole-program, **cross-assembly interprocedural** reachability analysis was run from the restore task entry points: **13 assemblies, ~7,267 reachable methods**. It catalogued every reachable static field/property and every "lingering process-state sink" (child processes, env edits/reads, timers, static event subscriptions). Three NuGet designs were measured against that inventory:

- **[#8](https://github.com/JanProvaznik/NuGet.Client/pull/8) (`NuGetTraits`)** — centralized environment reads + an end-of-build cleanup. Reset 8 items; 8 gaps.
- **[#9](https://github.com/JanProvaznik/NuGet.Client/pull/9)** — completed the env/path/throttle residual on top of [#8](https://github.com/JanProvaznik/NuGet.Client/pull/8). Reset 14; 4 intentional gaps.
- **[#10](https://github.com/JanProvaznik/NuGet.Client/pull/10) (ported here)** — a **minimal** generic keyed reset registry; no centralization. Each cache/resource stays internal and **self-registers** an internal reset; restore invokes them by key. [#10](https://github.com/JanProvaznik/NuGet.Client/pull/10) subsumes [#8](https://github.com/JanProvaznik/NuGet.Client/pull/8)+[#9](https://github.com/JanProvaznik/NuGet.Client/pull/9).

**Reset points** (`NuGetProcessState.ResetKey`):
- `StartRestore` — re-read environment/config-derived caches (they freeze on first use and would otherwise be stale in a reused process).
- `EndRestore` — tear down live OS resources (plugin processes) that process-exit used to reclaim.

**Coverage** (validated against the report) — all twelve resets are present in this PR: `StartRestore` resets `NuGetEnvironment`, `ConcurrencyUtilities`, `ExceptionLogger`, `NuGetFeatureFlags` (NuGet.Common/Shared), `ProxyCache` (NuGet.Configuration), `DefaultCredentialServiceUtility` credential-service (NuGet.Credentials), `PreviewFeatureSettings`, `HttpSourceResourceProvider` throttle, `NuGetTestMode`, `PackageIdValidator` (NuGet.Protocol), `DependencyGraphSpec` (NuGet.ProjectModel), and the `SourceRepositoryDependencyProvider` throttle (NuGet.Commands); `EndRestore` disposes plugins via `PluginManager`. The only unreset statics are intentional and documented: `X509ChainBuildPolicyFactory.Policy` (experimental), `PluginLogger.DefaultInstance` (debug-only), two `*.Testable.Default` test seams; the `ProtocolDiagnostics` static events are raise-sites only (subscribers are external hosts), so not a leak under `dotnet restore`.

**Why the MSBuild change makes this sufficient.** The reset registry is a per-process coordination point: it only works if restore runs in **one** coherent, long-lived process where the state and its reset are co-located. The previous workaround instead kept restore's state in a **dying** transient TaskHost — which made the reset registry pointless (a process that dies each invocation never reuses state). Routing `RestoreTask` to the **main** process, and relying on `StartRestore`/`EndRestore` to scrub its state around each restore, is what lets the persistent host reuse one process safely:

> [#10](https://github.com/JanProvaznik/NuGet.Client/pull/10) makes restore's process state **resettable**; routing `RestoreTask` to the main process (and removing the transient-host workaround) makes that reset **load-bearing** — the persistent main process is returned to a clean slate around every restore, exactly as a fresh per-build process used to be.

That is why the pair resolves the full static-state inventory, and why the two changes must ship together.

## Changes

**`src/msbuild`**
- `TaskRouter.NeedsTaskHostInMultiThreadedMode` now returns `false` (main process) for the type whose full name is `NuGet.Build.Tasks.RestoreTask`, matched by name (not assembly identity), consistent with the existing attribute detection.
- The `RequiresTransientTaskHost` special case and its `forceTransientTaskHost` handling in `AssemblyTaskFactory` are removed (the [dotnet/msbuild#13315](https://github.com/dotnet/msbuild/issues/13315) workaround is retired).
- `TaskRouter` integration tests inverted: RestoreTask now runs **in-process** in MT and Server modes and in the **same** process across builds (its static state is reset per restore rather than reclaimed by process death).

**`src/nuget-client`**
- `NuGetProcessState` registry (the only new public API) + per-owner internal self-registrations; once-per-build guard (`BuildTasksUtility.ResetProcessStateOncePerBuild`, scoped via MSBuild's build-lifetime registered-task-object registry) wired into `GetRestoreProjectStyleTask` (the first restore task on every node), `RestoreTask` (backstop + `EndRestore` in `finally`), and the static-graph `Console` entry.

## Testing notes

- Included: NuGet unit tests (registry once-per-build / parallel-invocation guard; proxy / HTTP-throttle / credential-service resets) and inverted MSBuild `TaskRouter` integration tests (RestoreTask in-process, same-process-across-builds, MT + Server).

### End-to-end validation (✅ passed) — two-PAT credential-leak scenario

Reproduced the exact scenario from [dotnet/msbuild#13660](https://github.com/dotnet/msbuild/pull/13660) (the temporary workaround this PR retires) against a **coordinated VMR build of this PR**.

**Setup**
- SDK: `dotnet-unified-build` of `refs/pull/7311/merge` (dnceng-public build `1477414`), `dotnet-sdk-11.0.100-ci-win-x64`. Verified to carry both halves: MSBuild `RunsInMainProcessDespiteNoAttribute` present and `RequiresTransientTaskHost` **gone**; NuGet `NuGetProcessState` / `ResetProcessStateOncePerBuild` present.
- Feed/package: `Microsoft.VisualStudio.OpenTelemetry` from the private DevDiv `vssdk` Azure Artifacts feed (auth required), via the Azure Artifacts Credential Provider plugin (so `PluginManager` + the credential service are exercised).
- Host: MSBuild Server (`DOTNET_CLI_USE_MSBUILD_SERVER=1` + `MSBUILDUSESERVER=1`) so one process services both restores back-to-back. Two short-lived PATs (A, B) minted/revoked via the ADO REST API; a fresh `RestorePackagesPath` per restore forces a real re-download + re-auth (`RestoreForce=true`, `RestoreNoHttpCache=true`).

**Procedure & outcome**

| Step | Result |
|---|---|
| Restore #1 with PAT-A | **OK** (7.4 s, real auth) |
| Revoke PAT-A, confirm rejected on the flat2 endpoint | **confirmed (HTTP 401)** |
| Restore #2 with PAT-B, **same** server process | **OK** (11.8 s, real re-download/re-auth) |
| Same MSBuild server across #1/#2 | **YES** — `MSBuild.exe /nodemode:8`, PID stable (corroborated by `MSBUILDDEBUGCOMM` comm-trace log for that PID) |
| RestoreTask routing | ran **in-process in the server** (binlog: 1 `RestoreTask`, not under a TaskHost) — confirms the leak-prone path is genuinely exercised |

**Verdict: no leak.** On `main` this scenario returns `error NU1301: 401 (Unauthorized)` on restore #2 (the revoked PAT-A leaks from the cached plugin in the long-lived process). With this PR, RestoreTask runs in the persistent server process (workaround retired) yet restore #2 still succeeds with PAT-B — the reset registry (`StartRestore` credential reset + `EndRestore` plugin teardown) refreshes credentials in the reused process. Same correctness as the #13660 workaround, without killing the process each restore.

> Not yet covered: a `/mt` multithreaded run of the same scenario, and the non-RestoreTask restore "satellite" tasks (e.g. `GetRestoreSettingsTask`) that run in a reused sidecar TaskHost and read environment-derived caches — their refresh rides on `GetRestoreProjectStyleTask`'s once-per-build reset in that sidecar and is not separately validated here.

> The NuGet half is hand-applied to `src/nuget-client` for review/coordination; it is tracked in [JanProvaznik/NuGet.Client#10](https://github.com/JanProvaznik/NuGet.Client/pull/10) and its canonical path into the VMR is the NuGet source mirror once that change merges upstream, which would supersede these edits on the next sync.

### DIFF
--- a/src/nuget-client/src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
+++ b/src/nuget-client/src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
@@ -145,6 +145,11 @@ namespace NuGet.Build.Tasks
 
             NuGet.Common.Migrations.MigrationRunner.Run();
 
+            // Re-read environment-derived caches at the start of this restore so a reused process observes the current
+            // environment. This normally happens in the first restore task on each node (GetRestoreProjectStyleTask);
+            // the call here is a guarded backstop that no-ops if that task already reset this node for this build.
+            BuildTasksUtility.ResetProcessStateOncePerBuild(BuildEngine);
+
             try
             {
                 return ExecuteAsync(log).Result;
@@ -159,6 +164,12 @@ namespace NuGet.Build.Tasks
             {
                 ExceptionUtilities.LogException(e, log);
                 return false;
+            }
+            finally
+            {
+                // End of restore on the entry node: tear down plugin processes that the per-build process exit used to
+                // reclaim. RestoreTask runs once per restore, so this needs no parallel-invocation guard.
+                NuGet.Common.NuGetProcessState.Reset(NuGet.Common.NuGetProcessState.ResetKey.EndRestore);
             }
         }
 
--- a/src/msbuild/src/Build.UnitTests/BackEnd/TaskRouter_IntegrationTests.cs
+++ b/src/msbuild/src/Build.UnitTests/BackEnd/TaskRouter_IntegrationTests.cs
@@ -385,20 +385,23 @@ namespace Microsoft.Build.Engine.UnitTests.BackEnd
         }
 
         [Fact]
-        public void RequiresTransientTaskHost_ReturnsTrueForRestoreTask()
+        public void NeedsTaskHostInMultiThreadedMode_RestoreTask_RunsInMainProcess()
         {
-            TaskRouter.RequiresTransientTaskHost(typeof(NuGet.Build.Tasks.RestoreTask)).ShouldBeTrue();
+            // RestoreTask resets its own static state, so it runs in the main process (no TaskHost) despite having
+            // no MSBuildMultiThreadableTaskAttribute.
+            TaskRouter.NeedsTaskHostInMultiThreadedMode(typeof(NuGet.Build.Tasks.RestoreTask)).ShouldBeFalse();
         }
 
         [Fact]
-        public void RequiresTransientTaskHost_ReturnsFalseForRegularTask()
+        public void NeedsTaskHostInMultiThreadedMode_RegularNonEnlightenedTask_NeedsTaskHost()
         {
-            TaskRouter.RequiresTransientTaskHost(typeof(NonEnlightenedTestTask)).ShouldBeFalse();
-            TaskRouter.RequiresTransientTaskHost(typeof(AttributeTestTask)).ShouldBeFalse();
+            // A regular task without the attribute is still isolated in a TaskHost sidecar.
+            TaskRouter.NeedsTaskHostInMultiThreadedMode(typeof(NonEnlightenedTestTask)).ShouldBeTrue();
+            TaskRouter.NeedsTaskHostInMultiThreadedMode(typeof(AttributeTestTask)).ShouldBeFalse();
         }
 
         [Fact]
-        public void RequiresTransientTaskHost_RoutedToTaskHost_InMultiThreadedMode()
+        public void RestoreTask_RunsInMainProcess_InMultiThreadedMode()
         {
             string projectContent = $@"
 <Project>
@@ -432,12 +435,12 @@ namespace Microsoft.Build.Engine.UnitTests.BackEnd
             BuildResult result = buildManager.Build(buildParameters, buildRequestData);
 
             result.OverallResult.ShouldBe(BuildResultCode.Success);
-            TaskRouterTestHelper.AssertTaskUsedTaskHost(logger, "RestoreTask");
+            TaskRouterTestHelper.AssertTaskRanInProcess(logger, "RestoreTask");
             logger.FullLog.ShouldContain("RestoreTask executed");
         }
 
         [Fact]
-        public void RequiresTransientTaskHost_RoutedToTaskHost_InServerMode()
+        public void RestoreTask_RunsInMainProcess_InServerMode()
         {
             string projectContent = $@"
 <Project>
@@ -475,12 +478,12 @@ namespace Microsoft.Build.Engine.UnitTests.BackEnd
             BuildResult result = buildManager.Build(buildParameters, buildRequestData);
 
             result.OverallResult.ShouldBe(BuildResultCode.Success);
-            TaskRouterTestHelper.AssertTaskUsedTaskHost(logger, "RestoreTask");
+            TaskRouterTestHelper.AssertTaskRanInProcess(logger, "RestoreTask");
             logger.FullLog.ShouldContain("RestoreTask executed");
         }
 
         [Fact]
-        public void RequiresTransientTaskHost_GetsFreshProcess_OnEachInvocation_InMultiThreadedMode()
+        public void RestoreTask_RunsInSameProcess_AcrossBuilds_InMultiThreadedMode()
         {
             string projectContent = $@"
 <Project>
@@ -491,7 +494,7 @@ namespace Microsoft.Build.Engine.UnitTests.BackEnd
     </Target>
 </Project>";
 
-            string projectFile = Path.Combine(_testProjectsDir, "RestoreTaskFreshProcess.proj");
+            string projectFile = Path.Combine(_testProjectsDir, "RestoreTaskSameProcess.proj");
             File.WriteAllText(projectFile, projectContent);
 
             var logger = new MockLogger(_output);
@@ -500,10 +503,6 @@ namespace Microsoft.Build.Engine.UnitTests.BackEnd
                 MultiThreaded = true,
                 Loggers = [logger],
                 DisableInProcNode = false,
-
-                // Reuse must stay ON so we test the workaround, not natural process death.
-                // With reuse off the test cannot distinguish "transient TaskHost (workaround)"
-                // from "sidecar TaskHost that happened to die between builds".
                 EnableNodeReuse = true,
             };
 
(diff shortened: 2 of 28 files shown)


src/nuget-client/src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
zivkan, line 151: It concerns me that this PR is being opened on the dotnet VMR, where NuGet's tests don't run. I see the PR is still in draft mode at this time, but I don't think it should be merged until the same changes are make in a NuGet.Client repo PR so that tests are run to see if it catches any regressions.

zivkan, line 151: It looks to me like most of the changes in this PR are related to things that NuGet reads from environment variablies.

Is there a design document that explains how MSBuild server handles environment variables when run from multiple command prompts, each of which have different environment variables?  I think a link to that design document would be very useful to have in the comments, so people working on NuGet can quickly get a reference to how MSBuild works in this regard.

zivkan, line 151: What happens if the customer has multiple terminals open (maybe multiple AI agents running), causing multiple solutions to be restored at the same time?   If multiple RestoreTask runs in the same process for different solutions at the same time, the current approach feels very dangerous to me.

For example, there's a timing issue between when the credential provider usage is reset back to "disabled", and then RestoreTask enables it again.  If another restore is running at the same time that needs the credential provider, it could have a bunch of failed requests due to unnecessarily resetting the cred provider statis/global variables.

nkolev92: We've been chatting offline too.

Link to some doc: https://github.com/dotnet/msbuild/blob/main/documentation/specs/multithreading/multithreaded-msbuild.md. 

There are probably cleaner ways to do most of this that match NuGet's coding patterns.

Me & @zivkan locally and we may just have to propagate EnvironmentVariableReader we have everywhere and then removing static caching.

zivkan: I'd go further and say that this PR probably does not work correctly. This PR will reset a bunch of static/global variables so that NuGet will read them again via `Environment.GetEnvironmentVariable`. But unless MSBuild explicitly uses `Environment.SetEnvironmentVariable` to change the value for the whole process (in which case, like my other comment asks, what happens when there are multiple builds happening at the same time), then NuGet will just re-read the same old value and resetting the variablies didn't change any outcome. If MSBuild saves per-build environment variables in some state object, NuGet needs to be modified to know how to read that state object, and stop using the .NET `Environment.GetEnvironmentVariable` API.

It's going to be very impactful to NuGet's public APIs to have to pass a NuGet `IEnvironmentVariableReader` everywhere, but I think it might be the only way if we want to be strictly correct.

But a more pragmatic approach might be if we even care about certain environment variables having different values in different shells. For example, what's the probability that someone changes the NuGet.Protocol enhanced retry settings?  Or their HOME directory?  What if we document how to restart the MSBuild server in these scenarios, and only make product changes for the highest risk environment variables?

I override the NUGET_PACKAGES environment variable regularly for testing, but PackageReference has an MSBuild property that I could use instead.  Can we just say that the NUGET_PACKAGES environment variable is unsupported for PackageReference? I guess no, that'd be a breaking change.

But I'd like us to think about whether we really need a "perfect" implementation, or what's "good enough".

JanProvaznik:
sorry nuget folks for your energy here, it's a draft so I assumed no one would look yet, I created this PR to get a coherent dotnet from the CI to do local testing with a coordinated change in msbuild. After having a first pass in our team I will open a PR to nuget.client

Your questions and pushback are good and reminded me I need to write a coherent design as an issue. https://github.com/NuGet/Home/issues/14958

I owe you an explanation of the assumptions and how we're changing them and why we need this PR before I ask for review.

src/nuget-client/src/NuGet.Core/NuGet.Build.Tasks/RestoreTask.cs
JanProvaznik: > MSBuild explicitly uses Environment.SetEnvironmentVariable to change the value for the whole process 

yes, we do that

> But I'd like us to think about whether we really need a "perfect" implementation, or what's "good enough".

the minimalist implementation is just to dispose the plugin manager which is a clear issue, other state issues are mostly theoretical but we're aiming for very high level of compatibility in the msbuild server and mutlithreaded modes so they can be set as defaults without breaks.

JanProvaznik: it's not intended to be reviewed here, hence a draft, sorry for not clarifying in title not to review

JanProvaznik: Multiple solutions from different terminals would be still restored in multiple processes. Having multiple independent build invocations in one process concurrently is not my aim. As far as I understand from shipped targets and looking at binlogs there is only one RestoreTask running as part of any single `dotnet build` or `dotnet restore` (though there are a few "satellite Nuget Tasks" that run for each project, presently in main and also worker processes which are reused by default already - so I don't think they will pose a reuse problem, they have some concurrency issues which we'll be adressing later)

MSBuild server is a persistent main process that should survive singular invocation which presently for dotnet build/restore dies. In later invocation then client sends as part of a request it's environment, that gets applied to server for the next invocation hence I claim resetting at Restore task time is sufficient.

JanProvaznik:
outdated, was merged in NuGet

PR #7334, opened 2026-06-23
Use default ProviderKey for Desktop bundle

joeloff:
# Overview

The Desktop Runtime bundle generates an explicit provider key instead of using the default which matches the bundle registration code. This can result in blocking the removal of a previous install. As shown below, the provider key and dependent (bundle registration code) are not the same.

image

# Customer Impact

When the deferred removal key is set to `nextSession`, the old version of the runtime is not removed because there are multiple dependents. The screenshot below shows that the 10.0.2 and 10.0.9 installs are registered against the same provider.

image

This also results in overwriting the provider data when attempting to perform an out-of-order installs. This scenario is supported to allow ISVs to redist runtimes with their software. If a user already has a newer runtime installed, we do not want the ISV install to fail.

# Regression

Yes, this is a regression from .NET 9.0

# Testing

Manually verified. Below is a screenshot from a dev build showing that the provider key and dependent now match

image

### DIFF
--- a/src/windowsdesktop/src/windowsdesktop/src/bundle/Wix.targets
+++ b/src/windowsdesktop/src/windowsdesktop/src/bundle/Wix.targets
@@ -126,25 +126,15 @@
   </Target>
 
   <Target Name="GenerateUpgradeCode" AfterTargets="SetInstallerInfo">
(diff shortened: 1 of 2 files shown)


joeloff:
Approved by tactics over email.

lewing (approved):

PR #7357, opened 2026-06-24
Add mono platforms to exclude list

### LINKED ISSUE #7323: s390x/ppc64le have CMake Error at nativeaot/Runtime/Full/CMakeLists.txt
The cross-builds we have for s390x and ppc64le are failing:
```
    CMake Error at nativeaot/Runtime/Full/CMakeLists.txt:187 (add_library):
      Cannot find source file:
    
        /home/tester/workdir/dotnet/src/runtime/src/coreclr/runtime/s390x/AllocFast.S
    
    
    CMake Error at nativeaot/Runtime/Full/CMakeLists.txt:187 (add_library):
      No SOURCES given to target: Runtime.AsmHelpers
```

Probably this regressed in https://github.com/dotnet/dotnet/pull/7004.

https://github.com/dotnet/dotnet/blob/d8addc1562ad98c7cab19ed654e9caabd1112d87/src/arcade/eng/common/native/NativeAotSupported.props#L13-L17

vs

https://github.com/dotnet/dotnet/blob/e9cde69d4ee92243ccd47d9dd0b2ffd5f9cf4c3c/src/runtime/eng/Subsets.props#L51

cc @am11 @akoeplinger @ViktorHofer @omajid

am11:
Fix https://github.com/dotnet/dotnet/issues/7323.

### DIFF
--- a/src/arcade/eng/common/native/NativeAotSupported.props
+++ b/src/arcade/eng/common/native/NativeAotSupported.props
@@ -13,6 +13,8 @@
     <!-- Reject unsupported architectures via RID suffix match -->
     <_NativeAotSupportedArch Condition="
       '$(TargetArchitecture)' != 'wasm' and
+      '$(TargetArchitecture)' != 's390x' and
+      '$(TargetArchitecture)' != 'ppc64le' and
       ('$(TargetArchitecture)' != 'x86' or '$(TargetOS)' == 'windows')
     ">true</_NativeAotSupportedArch>
 
--- a/src/runtime/src/coreclr/CMakeLists.txt
+++ b/src/runtime/src/coreclr/CMakeLists.txt
@@ -156,7 +156,7 @@ endif ()
 if(NOT CLR_CROSS_COMPONENTS_BUILD)
   # NativeAOT is buildable for all CoreCLR-supported configurations except a known reject list.
   # Keep this in sync with the NativeAotSupported reject list in eng/common/native/NativeAotSupported.props.
-  if(NOT (CLR_CMAKE_HOST_ARCH_WASM OR (CLR_CMAKE_HOST_ARCH_I386 AND NOT CLR_CMAKE_HOST_WIN32)))
+  if(NOT (CLR_CMAKE_HOST_ARCH_WASM OR CLR_CMAKE_HOST_ARCH_S390X OR CLR_CMAKE_HOST_ARCH_POWERPC64 OR (CLR_CMAKE_HOST_ARCH_I386 AND NOT CLR_CMAKE_HOST_WIN32)))
     add_subdirectory(nativeaot)
   endif()
 endif(NOT CLR_CROSS_COMPONENTS_BUILD)
--- a/src/aspnetcore/eng/common/native/NativeAotSupported.props
+++ b/src/aspnetcore/eng/common/native/NativeAotSupported.props
@@ -13,6 +13,8 @@
     <!-- Reject unsupported architectures via RID suffix match -->
     <_NativeAotSupportedArch Condition="
       '$(TargetArchitecture)' != 'wasm' and
+      '$(TargetArchitecture)' != 's390x' and
+      '$(TargetArchitecture)' != 'ppc64le' and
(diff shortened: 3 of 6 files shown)


src/runtime/src/coreclr/CMakeLists.txt
```diff
@@ -156,7 +156,7 @@ endif ()
 if(NOT CLR_CROSS_COMPONENTS_BUILD)
   # NativeAOT is buildable for all CoreCLR-supported configurations except a known reject list.
   # Keep this in sync with the NativeAotSupported reject list in eng/common/native/NativeAotSupported.props.
-  if(NOT (CLR_CMAKE_HOST_ARCH_WASM OR (CLR_CMAKE_HOST_ARCH_I386 AND NOT CLR_CMAKE_HOST_WIN32)))
+  if(NOT (CLR_CMAKE_HOST_ARCH_WASM OR CLR_CMAKE_HOST_ARCH_S390X OR CLR_CMAKE_HOST_ARCH_POWERPC64 OR (CLR_CMAKE_HOST_ARCH_I386 AND NOT CLR_CMAKE_HOST_WIN32)))
```
tmds, line 159: Does the coreclr file need to change? s390x and ppc64le use Mono.

src/arcade/eng/common/native/NativeAotSupported.props
```diff
@@ -13,6 +13,8 @@
     <!-- Reject unsupported architectures via RID suffix match -->
     <_NativeAotSupportedArch Condition="
       '$(TargetArchitecture)' != 'wasm' and
+      '$(TargetArchitecture)' != 's390x' and
```
tmds, line 16: Can/should we use Mono here as a condition?

Sometimes I build x64 with Mono, how should NativeAotSupported behave for that?

am11: We would need to implement RuntimeFlavor=coreclr/mono in all consuming repos with same name. We actually don't need that here. If the target platform is supported by NativeAOT, and something is requiring NativeAOT, it will use ilc regardless of runtime flavor.

src/runtime/src/coreclr/CMakeLists.txt
am11: Yes, your error message is reaching this point, and the code comment suggests we should keep them in sync.

src/arcade/eng/common/native/NativeAotSupported.props
tmds: Could we use `DotNetBuildUseMonoRuntime`?

am11: We can't use that in other parts which we need to keep this in sync with. There shouldn't be any error with current state. The tools which are using PublishAot should disable it at callsite for mono (which they are already doing; cDAC etc.).

src/runtime/src/coreclr/CMakeLists.txt
MichalStrehovsky: Does the patch I proposed fix the subset of the build that currently works? Like I wrote in https://github.com/dotnet/dotnet/pull/7004#pullrequestreview-4511743472, I expected there could be some fallout, but it should be minimal. Whatever is currently in this list here is misleading people doing bringup into thinking they made the entire `clr` subset build when in fact they only made `clr.iltools+clr.pal+clr.runtime+...` build (because as a "convenience" we auto-excluded clr.aot behind their backs). I don't think it's a good strategy.

The comment about keeping things in sync should be deleted. The fact that some platforms can build on Mono is unrelated to CoreCLR build. The checks now serve different purposes. I don't see reason to keep them in sync (or we can split the groups in NativeAotSupported.props into two: one to keep in sync, and another to delete when we delete Mono).

am11: I don't know. This was a divergence in my "put this code in shared dir" PR which I'm fixing. The refactoring should be done separately.

MichalStrehovsky (approved):

tmds:
@am11, thanks for the quick fix!

PR #7378, opened 2026-06-25
[release/11.0.1xx-preview6] Fix vstest Linux pack failures (NU5129 + CLI net11.0) for #7356

azat-msft:
## Summary

This PR carries the same vstest codeflow update as the automated PR #7356, **plus** a fix for the VMR build failures that PR is hitting. It is opened as a standalone (non-codeflow) branch so the fix is not force-overwritten by the next maestro flow.

Supersedes / fixes the build of: #7356

## Why #7356 fails

The failing leg `VMR Vertical Build Validation Ubuntu2404_Ubuntu_BuildTests_x64` (a non‑source‑only `--productBuild` **Linux** leg) fails during `dotnet pack` of the vstest packages with two distinct root causes, both introduced by the vstest nuspec → SDK‑pack migration:

1. **NU5129** on `Microsoft.NET.Test.Sdk` and `Microsoft.CodeCoverage` — `PackagePath` used trailing **backslashes** (e.g. `build\netstandard2.0\`). On Linux NuGet does not treat a trailing `\` as a directory separator, so the entry becomes a double‑slash path (`build/netstandard2.0//<PackageId>.props`). NuGet's package‑convention check then can't find `build/<tfm>/<PackageId>.props` and raises NU5129 (promoted to error).

2. **`Could not find a part of the path '.../Microsoft.TestPlatform.CLI/Release/net11.0'`** — the migrated `Microsoft.TestPlatform.CLI` csproj's "source‑build / non‑Windows" content `ItemGroup` pulls files from `$(OutputPath)$(NetCurrent)` (net11.0). A `net11.0` output only exists in **source‑only** builds, where `src/vstest/Directory.Build.targets` overrides `TargetFrameworks` to `$(NetCurrent)`. In the **non‑source‑only Linux product build** the project builds `$(NetSDKTargetFramework)` (net10.0), so `.../Release/net11.0` never exists and pack fails. (Windows is unaffected because it uses the hard‑coded `net10.0` content group; the source‑only CentOS legs are unaffected because they actually build net11.0.)

## Changes

- **`Microsoft.CodeCoverage.csproj`**, **`Microsoft.NET.Test.Sdk.csproj`**, **`Microsoft.TestPlatform.TestHost.csproj`** — use forward‑slash `PackagePath` values so paths resolve cleanly on both Windows and Linux (fixes NU5129).
- **`Microsoft.TestPlatform.CLI.csproj`** — add a mode‑aware property `_CliContentTfm` (= `$(NetCurrent)` when `DotNetBuildSourceOnly == true`, else `$(NetSDKTargetFramework)`) and use it for the cross‑platform content TFM, so pack references the output folder that is actually produced in each build mode.

## Validation

Validated locally without a full VMR build:

- **NU5129**: standalone `dotnet pack` with the real PackageId/filename — backslash `PackagePath` reproduces NU5129; forward slash produces clean single‑slash entries (`build/net462/Microsoft.NET.Test.Sdk.props`) and no NU5129.
- **CLI**: standalone MSBuild evaluation of `_CliContentTfm` — product build resolves to `net10.0` (the folder built in the failing leg); source‑only build resolves to `net11.0` (the folder built in source‑build legs).

> [!NOTE]
> `src/vstest/**` mirrors `microsoft/vstest`. The canonical fix should also land in microsoft/vstest so it survives future codeflow updates; this PR is the VMR‑side unblock.


nohwnd (approved):
LGTM, we just need to push this past the manifest edit.

azat-msft:
@dotnet-policy-service agree company="Microsoft"

PR #7461, opened 2026-06-29
Update WiX toolset

joeloff:
# Description

Update the WiX toolset. The new build includes a fix to address a regression related to deferred uninstalls.

Fixes https://github.com/dotnet/aspnetcore/issues/67128

This was approved by tactics over email. The actual fix is [here](https://dev.azure.com/dnceng/internal/_git/dotnet-wix/pullrequest/62408)

# Notes

Because the fix changes the bootstrapper application logic, the fix will only be applicable after two consecutive versions that include the fix have been shipped. Any previous installation of .NET 10 with the regression will still result in the behavior described in the linked issue because the older version of .NET will detect the upgrade related bundle of the newer version and still write the RunOnce key. We cannot fix installs that have shipped and already installed on users' machines.

### DIFF
--- a/src/arcade/Directory.Packages.props
+++ b/src/arcade/Directory.Packages.props
@@ -11,7 +11,7 @@
     <!-- Keep Microsoft.Data.Services.Client package version in sync with Microsoft.Data.OData -->
     <MicrosoftDataServicesClientVersion>5.8.4</MicrosoftDataServicesClientVersion>
     <MicrosoftSignedWixVersion>3.14.1-9323.2545153</MicrosoftSignedWixVersion>
-    <MicrosoftWixToolsetSdkVersion>5.0.2-dotnet.2811440</MicrosoftWixToolsetSdkVersion>
+    <MicrosoftWixToolsetSdkVersion>5.0.2-dotnet.3009129</MicrosoftWixToolsetSdkVersion>
     <!-- Overwrite XUnitVersion/XUnitAnalyzersVersion/XUnitRunnerConsoleVersion/XUnitRunnerVisualStudioVersion that comes from the Arcade SDK to be in sync as Arcade doesn't use a live Arcade SDK.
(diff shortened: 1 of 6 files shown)


joeloff:
I've already verified a private build of WiX and a rebuild of the desktop runtimes last week. Now that we have an official build I'm going to verify that before doing the ports for 3xx and 4xx branches. There will be a separate fix for .NET 11 since it requires changes to the v6 copy of the toolset.

[EDIT]

Verified. Log excerpt below is from the RunOnce to remove a private 10.0.10-dev build created by installing 10.0.11-dev private build.

New toolset version is confirmed: 5.0.2+4ccb02... The related 10.0.11-dev build is detected as a downgrade and the request BA action becomes None. The newer version remains.

```
[1A48:1A4C][2026-06-29T09:54:38]i001: Burn x86 v5.0.2+4ccb02a8ed3ebda81ea8f3b51436fc8c3b8993be, Windows v10.0 x64 (Build 17763: Service Pack 0), path: C:\ProgramData\Package Cache\{C6ADE9C9-F28A-4E40-928B-DE10FBEA9082}\windowsdesktop-runtime-10.0.10-dev-win-x64.exe
[1A48:1A4C][2026-06-29T09:54:38]i009: Command Line: '/uninstall /quiet'
[1A48:1A4C][2026-06-29T09:54:39]i000: Setting string variable 'RemoveUpgradeRelatedBundle' to value 'nextSession'
[1A48:1A4C][2026-06-29T09:54:39]i207: Planned related bundle: {1E144B2F-B6DF-409A-9530-E54E66A50974}, detect type: Upgrade, default plan type: Downgrade, ba plan type: Downgrade, default requested: None, ba requested: None, execute: None, rollback: None, default requested restore: None, ba requested restore: None, restore: None, dependency: None
```

akoeplinger (approved):

PR #7465, opened 2026-06-29
Remove TSA bug filing from pipelines

mmitche:
Removes TSA bug filing configuration from the repo's pipeline YAML files. This eliminates the `tsa` nodes and the `codeql.tsaEnabled` / `tsaEnabled` settings; since the default for TSA is disabled, no TSA bugs will be filed.

Affected files:
- eng/pipelines/official.yml
- eng/pipelines/source-build-outer-loop.yml
- eng/pipelines/source-build-sdk-diff-tests.yml
- eng/pipelines/vmr-compare.yml
- eng/pipelines/vmr-license-scan.yml


ViktorHofer (approved):

akoeplinger:
I assume we can also remove https://github.com/dotnet/dotnet/blob/main/.config/tsaoptions.json ?

should we do the same for the individual repos?

mmitche:
> I assume we can also remove https://github.com/dotnet/dotnet/blob/main/.config/tsaoptions.json ?
> 
> should we do the same for the individual repos?

Individual repos still need to be onboarded to WIM before they can have their TSA filing removed. I'll remove TSA options.

ViktorHofer (approved):

akoeplinger:
> Individual repos still need to be onboarded to WIM before they can have their TSA filing removed

I meant individual repos which don't have an official build anymore, like runtime. Those shouldn't need to onboard to WIM separately since we just go through the VMR right?

mmitche:
> > Individual repos still need to be onboarded to WIM before they can have their TSA filing removed
> 
> I meant individual repos which don't have an official build anymore, like runtime. Those shouldn't need to onboard to WIM separately since we just go through the VMR right?

They may need to onboard for servicing 8/9 branches, if we want to have bug filing for those releases. Otherwise, no

akoeplinger:
Ok. What about this PR, backport to 10.0?

akoeplinger:
we're seeing warnings about this in 10.0 as well, backporting

PR #7490, opened 2026-06-30
Revert "Disable NETSDK1242 mobile-mono check in the Mono source-build leg"

kotlarmilos:
## Description

This reverts dotnet/dotnet#7359, which passed the SDK property in the Mono source-build leg so the multi-targeted libraries wouldn't throw an mobile on Mono check. That was a temporary workaround, and the runtime now sets `UseMonoRuntime=false` for those mobile target frameworks so they build as CoreCLR and never hit the check in the first place. With the runtime fix in place the property is no longer needed.

### DIFF
--- a/repo-projects/runtime.proj
+++ b/repo-projects/runtime.proj
@@ -22,15 +22,6 @@
     -->
     <BuildArgs Condition="'$(DotNetBuildUseMonoRuntime)' == 'true'">$(BuildArgs) $(FlagParameterPrefix)usemonoruntime</BuildArgs>
 
-    <!--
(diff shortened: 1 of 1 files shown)


ViktorHofer (approved):

mthalman (approved):

akoeplinger (approved):

kotlarmilos:
When https://github.com/dotnet/runtime/pull/130036 flows here, we can merge this PR

kotlarmilos:
Vertical build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1493108

PR #7600, opened 2026-07-08
Add SCCSID to AOT published shipping binaries

### LINKED ISSUE #67637: aspnetcoretools.so: use source control version string format.
.so files in the SDK use this pattern for the version string:
```
$ strings shared/Microsoft.NETCore.App/11.0.0-preview.7.26356.111/libSystem.Net.Security.Native.so | grep '@(#)'
@(#)Version 11.0.26.35711 @Commit: ccf0d067ca5d747484d7e330ca6aa8c14ede9cb3
```
`aspnetcoretools.so` doesn't include such a string:
```
$ strings /home/tmds/projects/tools-dont-execute/dotnet/sdk/11.0.100-preview.7.26356.111/DotnetTools/aspnetcoretools/11.0.0-preview.7.26356.111/tools/any/fedora.44-x64/libaspnetcoretools.so | grep '@(#)'
```

It would be nice if it would.

Note: the hash is appearing in some places in the binary already:
```
$ strings sdk/11.0.100-preview.7.26356.111/DotnetTools/aspnetcoretools/11.0.0-preview.7.26356.111/tools/any/fedora.44-x64/libaspnetcoretools.so | grep ccf0d067ca5d747484d7e330ca6aa8c14ede9cb3
11.0.0-preview.7.26356.111+ccf0d067ca5d747484d7e330ca6aa8c14ede9cb3
CommitHashPccf0d067ca5d747484d7e330ca6aa8c14ede9cb3
https://github.com/dotnet/dotnet/tree/ccf0d067ca5d747484d7e330ca6aa8c14ede9cb3,OriginalRepoCommitHashP2b7a3d5c1757296f038696e208662370edac5368
.NET: ilc 11.0.0-preview.7.26356.111+ccf0d067ca5d747484d7e330ca6aa8c14ede9cb3
```

am11:
```sh
$ strings libaspnetcoretools.so | grep '@(#)'
@(#)Version 11.0.26.35999 @Commit: 666180633e44fd2a7af63577d024bc630075f93b

$ strings crossgen2 | grep '@(#)'
@(#)Version 11.0.26.35999 @Commit: 666180633e44fd2a7af63577d024bc630075f93b

$ strings ilc | grep '@(#)'
@(#)Version 11.0.26.35999 @Commit: 666180633e44fd2a7af63577d024bc630075f93b
```

Fixes https://github.com/dotnet/aspnetcore/issues/67637.


src/arcade/eng/common/native/NativeAotSccsId.targets
```diff
@@ -0,0 +1,13 @@
+<Project>
+  <!--
+    Inject SCCSID in AOT apps which we include in libcoreclr and other runtime native libs.
+    Keep in sync with arcade's Microsoft.DotNet.Build.Tasks.Packaging/src/build/Packaging.targets
```
am11: @ViktorHofer, @akoeplinger, just wanted to know your take on the approach here; is this standalone approach better or would you rather have this target in `Packaging.targets` directly (near the similar infra)? e.g. we could condition this target on `UseArcadeSccsidTarget` set by relevant consumers.

am11:
@MichalStrehovsky, PTAL.

MichalStrehovsky:
> @MichalStrehovsky, PTAL.

Sorry, I'm not the best person to review this. I literally wrote "I don't have an opinion on this. I don't know why we include _version.c in the first place.". As far as I'm concerned, it's useless information and I would not sign off on a change that doesn't do anything useful because we still need to maintain it.

It will add this string to native AOT outputs, so from that perspective I'm happy to sign off, provided we need this.

am11:
>  I don't know why we include _version.c in the first place.

It's a unix convention https://en.wikipedia.org/wiki/Source_Code_Control_System. Windows binaries have PE metadata populated via source link, so _version.c is the unix-y answer to that.

> As far as I'm concerned, it's useless information and I would not sign off on a change that doesn't do anything useful because we still need to maintain it.

Off the top of my head, three current use-cases:
* we use it for diagnostics and to answer like "which commit"
* corehost prints that value with DOTNET_HOST_TRACING
* libSOS uses it to find matching runtime by commit.

One may not be a fan of it but it certainly is not useless. We are adding SCCSID in "native binaries shipped in .NET". This is to make it consistent "nativeaot published binaries shipped by .NET". I think it should be fine.

cc @jkotas

MichalStrehovsky:
Okay, I'm fine with it. Found the original motivation for the use in our repos at https://github.com/dotnet/runtime/issues/5107.

Do you have an opinion on whether we should match the version.c inclusion rules? (I assume we don't include version.c on Windows at minimum.)

akoeplinger:
I agree we shouldn't emit this on Windows, it has no meaning there.

I'm a bit torn on the approach, I don't like that we need to remember to include the target everywhere we have a NativeAOT'd app.

I saw that at least on Windows the NativeAOT'd .exe still contains the PE Copyright / ProductVersion fields from somewhere, I assume it propagates them from the original .NET assembly? maybe we could do the same for sccsid on Unix?

am11:
I can exclude it but we had this discussion earlier https://github.com/dotnet/runtime/issues/51554#issuecomment-827184895, the idea was we don't want to override PE resource with our version but rather emit it in Windows. The only reason why we don't emit it in apphost is because how arcade infra is written and nobody worked on it.

I will exclude it from here and we can align it once other win32 runtime binaries receive the fix.

jkotas:
> I'm a bit torn on the approach, I don't like that we need to remember to include the target everywhere we have a NativeAOT'd app.

+1

Can this be done by Arcade targets that generate the @Version resource for the C/C++ binaries?

am11:
@jkotas, you mean https://github.com/dotnet/dotnet/pull/7600#discussion_r3546147398? We are including files from `eng/common/native` in all AOT projects. We can unify those in arcade SDK as well. For now there are only four places using AOT in shipping matrix, so we can unify it later as well.

jkotas:
I think it would be better to do it correctly right of the gate so that we do not have deal with this again. We have been adopting for tools accros the stack and I expect that it is going to continue.

Can the equivalent of this change be achieved by adding the logic to https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.targets ?

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.targets
```diff
@@ -40,23 +40,39 @@
...
+          DependsOnTargets="$(_GenerateVersionInfoDependsOn)">
+    <PropertyGroup>
+      <!-- To support builds without a source control provider available, allow this property to be unset. -->
+      <_SourceBuildInfo Condition="'$(SourceRevisionId)' != ''"> %40Commit: $(SourceRevisionId)</_SourceBuildInfo>
+      <_SccsIdString>@(#)Version $(FileVersion)$(_SourceBuildInfo)</_SccsIdString>
+    </PropertyGroup>
+  </Target>
+
+  <!--
+    Inject SCCSID in AOT apps which we include in libcoreclr and other runtime native libs.
```
jkotas:
```suggestion
    Inject SCCSID in AOT apps.
```

am11:
Moved to arcade SDK and verified that it still works: https://github.com/am11/dotnet-riscv/releases/tag/11.0.100-preview.7.26364.199

```sh
$ find ~/.dotnet-new -name libaspnetcoretools.so | xargs strings | grep '@(#)'
@(#)Version 11.0.26.36599 @Commit: df02874e4e7bdd772c58bf6342a3756f03b68888

$ find ~/.dotnet-new -name crossgen2 | xargs strings | grep '@(#)'
@(#)Version 11.0.26.36599 @Commit: df02874e4e7bdd772c58bf6342a3756f03b68888

$ find ~/.dotnet-new -name ilc | xargs strings | grep '@(#)'
@(#)Version 11.0.26.36599 @Commit: df02874e4e7bdd772c58bf6342a3756f03b68888

$ find ~/.dotnet-new -name libcoreclr.so | xargs strings | grep '@(#)'
@(#)Version 11.0.26.36599 @Commit: df02874e4e7bdd772c58bf6342a3756f03b68888
@(#)Version 11.0.26.36599 @Commit: df02874e4e7bdd772c58bf6342a3756f03b68888
@(#)Version 11.0.26.36599 @Commit: df02874e4e7bdd772c58bf6342a3756f03b68888
@(#)Version 11.0.26.36599 @Commit: df02874e4e7bdd772c58bf6342a3756f03b68888
````

jkotas (approved):
Thanks!

MichalStrehovsky (approved):
This turned out pretty nice! Thanks!

PR #7622, opened 2026-07-09
Change RepoDotNetFinalVersionKind to 'unstable'

ViktorHofer:
To counter msbuild which already stabilized its vs18.9 branch that inserts into 4xx. Fixes consumption in sdk: https://github.com/dotnet/sdk/pull/55202

Needs fix: https://github.com/dotnet/dotnet/pull/7623


mmitche (changes requested):
I don't think this is correct. MSBuild should be able to stabilize without issue

PR #7643, opened 2026-07-09
Update allowed-sb-binaries.txt to remove MicroBuild

ellahathaway:
https://github.com/dotnet/arcade/pull/16307 moved MicroBuild installation into a temp directory, so we should be able to remove `MicroBuild/` from allowed binaries list.


ellahathaway:
Need to run a test signed build to verify this change before it can be merged

mthalman (approved):

PR #7719, opened 2026-07-14
Move Windows dnx.cmd ownership from the SDK dev MSI to the sharedhost

mthalman:
This is the Windows follow-up to https://github.com/dotnet/sdk/issues/55116.

Today on Windows, `dnx.cmd` is shipped only by the SDK `dev` MSI; the runtime sharedhost MSI does not ship it (the Unix `dnx` shim already lives in the host, but Windows was never brought into line). This change moves ownership of `dnx.cmd` to the sharedhost package and stops the SDK component from installing it, so the two install roots stay in sync and only one MSI ever writes `dnx.cmd`.

This also includes a fix in `dnx.cmd` for a preexisting issue to ensure that `dnx`'s exit code is propagated instead of resetting it to 0.


dsplaisted:
Definitely need feedback from @joeloff on the impact of this and whether it will break anything.  I can imagine due to the order things run in that this could leave you without a dnx.cmd, for example.

I see a new dnx.cmd file being added, should there be a corresponding deletion of the old one?

mthalman:
> I see a new dnx.cmd file being added, should there be a corresponding deletion of the old one?

No, it's still bundled by the SDK for the zip file.

dsplaisted:
> No, it's still bundled by the SDK for the zip file.

Do we need to keep two copies in the source tree then?  Can we use one copy as the source for both cases?

mthalman:
> Do we need to keep two copies in the source tree then? Can we use one copy as the source for both cases?

I don't know how that would be done. Do you have suggestions?

dsplaisted:
> I don't know how that would be done. Do you have suggestions?

Oh, I guess the issue is that they're in different repos.  Maybe the SDK build could extract it from a runtime package, but that's a lot more complicated.

akoeplinger:
@mthalman that's how it works for Unix though? the dnx script is extracted from the host nupkg

(I'm fine with the duplication if we want to get it in for p7 and clean up later)

mthalman:
> @mthalman that's how it works for Unix though? the dnx script is extracted from the host nupkg

No, for Unix the sdk repo also has its own copy of the [`dnx` script](https://github.com/dotnet/dotnet/blob/main/src/sdk/src/Layout/redist/dnx) that is referenced when producing the SDK tarball.

src/runtime/src/installer/pkg/sfx/installers/host.wxs
```diff
@@ -25,6 +25,12 @@
           <CopyFile Id="copyFileCoreHostExe" DestinationDirectory="PROGRAMFILES_DOTNET" />
         </File>
       </Component>
+
+      <Component Id="cmpDnxCmd" Directory="DOTNETHOME" Guid="{BD6B9661-33D5-4BAE-A581-A0BCD0369B2A}">
```
joeloff: Generate a stable GUID, it should be the same as before given that the keypath was the same.

A hard coded component GUID will result in problems when you have SxS installs or dealing with upgrades, etc.

Copying the file should also be avoided.

joeloff:
Today, for an x64 install of multiple SDKs I have this:

image

There are three MSIs associated with DNX and it's authored as a shared component (`{2E5090DD-84E9-5983-A4D7-1222CDAE23B9}`) You need to ensure that moving it to the runtime has the same component GUID. That should happen automatically if you generate stable GUID.

Below is from having both x86 and x64 SDKs installed. Two separate components with unique IDs because their keypaths are different.

image

joeloff (changes requested):
Please fix the component authoring and verify the stable GUIDs from a build with the change against existing SDK installations.

src/sdk/src/Layout/redist/dnx.cmd
```diff
@@ -14,4 +14,4 @@ set "SDK_PATH=%~dp0sdk\%SDK_VERSION%\dotnet.dll"
 
 "%DOTNET%" exec "%SDK_PATH%" dnx %*
 
-endlocal
+endlocal & exit /b %ERRORLEVEL%
```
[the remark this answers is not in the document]
mthalman: Fixed in e1ea531

src/runtime/src/installer/pkg/sfx/installers/host.wxs
mthalman: Switched `cmpDnxCmd` to an auto-generated GUID. Fixed in 3d991e5.

On avoiding the copy: I modeled `cmpDnxCmd` on `cmpCoreHost`, which uses `CopyFile` to mirror `dotnet.exe` into `PROGRAMFILES_DOTNET` so `dotnet` on `PATH` still resolves when the shared host is installed to a custom location. My intent was to keep `dnx.cmd` consistent with `dotnet.exe`. Are you suggesting we drop the CopyFile specifically for `dnx.cmd`?

mthalman: Whoops, sorry. I realize this should not be auto-generated. I'll fix that. Question on CopyFile still stands.

akoeplinger:
> No, for Unix the sdk repo also has its own copy of the [dnx script](https://github.com/dotnet/dotnet/blob/main/src/sdk/src/Layout/redist/dnx) that is referenced when producing the SDK tarball.

ok, sorry, looks like I confused this with something else.

mthalman:
## Windows dnx.cmd ownership handoff — validation summary

All scenarios validated using Preview 6 SDK as the "old" version and the bits from this PR as the "new" version. The host and SDK `dnx.cmd` component GUIDs match per install path, so Windows Installer ref-counts the file correctly across products:

| Arch | Install path | SDK component GUID | Host cmpDnxCmd GUID | Match |
|------|--------------|--------------------|---------------------|-------|
| x64 | C:\Program Files\dotnet | {2E5090DD-84E9-5983-A4D7-1222CDAE23B9} | {2E5090DD-84E9-5983-A4D7-1222CDAE23B9} | Yes |
| arm64 | C:\Program Files\dotnet | {2E5090DD-84E9-5983-A4D7-1222CDAE23B9} | {2E5090DD-84E9-5983-A4D7-1222CDAE23B9} | Yes |
| x86 | C:\Program Files (x86)\dotnet | {A2FDCE00-6A7A-5F7B-A14D-0F992E92DDBB} | {A2FDCE00-6A7A-5F7B-A14D-0F992E92DDBB} | Yes |

### New install (host owns dnx.cmd)
1. Clean machine — dnx.cmd absent.
2. Install host — dnx.cmd present.
3. Uninstall host — dnx.cmd removed.

### Existing-SDK coexistence handoff
1. Install old SDK — dnx.cmd present.
2. Install new host — both co-own the file.
3. Uninstall old SDK — dnx.cmd survives (host still owns it).
4. Uninstall new host — dnx.cmd removed.

### Host upgrade
1. Install old host + old SDK — dnx.cmd present.
2. Uupgrade the host — dnx.cmd survives, now co-owned by new host.
3. Remove old SDK — dnx.cmd survives (host owns it).
4. Install new SDK (no dnx.cmd) — file undisturbed.
5. Uninstall host — dnx.cmd removed.

Result: **PASS** in all scenarios. `dnx.cmd` is retained as long as any product owns it and cleaned up only when the last owner is removed.

joeloff (approved):

PR #7192, opened 2026-06-11
Fix vstest rel/18.7 SCI 10.0.0 sync 

nohwnd:
`System.Collections.Immutable` in vstest needs to be **10.0.0**, not `9.0.11`.

`rel/18.7` in microsoft/vstest is on SCI 10.0.0 and its binding redirects in all three product app.configs (`vstest.console`, `testhost.x86`, `datacollector`) point at `10.0.0.0`. The VMR copy currently says `9.0.11` in `Version.Details.xml`, so when `testhost.x86.exe` starts under an SDK built from this VMR it can't load `System.Collections.Immutable 10.0.0.0` and dies with `FileLoadException`. That's DevDiv bug 3015885 — `dotnet test --collect:"Code Coverage"` aborting under SDK `10.0.400-preview.0.26310.101`.

### How we got here

The previous codeflow PR #7018 was supposed to bring this bump over from `rel/18.7` and silently dropped it. `src/vstest/src/vstest.console/app.config` had been reformatted from a single-line blob to multi-line XML on the VMR side at some past point, so darc couldn't auto-merge. The PR was finished with `darc vmr resolve-conflict --unsafe`, which picked the VMR side for `vstest.console/app.config` and the rel/18.7 side for the other two. Net effect: `testhost.x86` and `datacollector` got the new `10.0.0.0` redirects, but `vstest.console`'s redirect stayed at `9.0.0.0` and `Version.Details.xml` stayed at `9.0.11`.

Source-build stayed green because nothing checked redirect↔package consistency — vstest PR #16092 had disabled `_VerifyNuGetPackages` under `DotNetBuild=true` the day before. And the VMR doesn't run vstest's acceptance tests, so the runtime failure didn't surface until the resulting SDK ran an end-to-end scenario.

### Changes

- `eng/Version.Details.xml` — SCI `9.0.11` → `10.0.0` (Sha matches `rel/18.8` and `main`).
- `eng/Version.Details.props` — declare `SystemCollectionsImmutablePackageVersion=10.0.0` and the `SystemCollectionsImmutableVersion` alias so the vstest csprojs resolve to `10.0.0`. (rel/18.7 declares this version with a `Versions.props` literal; the literal gets stripped in the VMR, which is why we declare it the Maestro way here instead.)
- `src/vstest/src/vstest.console/app.config` — restored byte-identical to rel/18.7. `testhost.x86/app.config` and `datacollector/app.config` were already byte-identical.

[...]

PR #7755, opened 2026-07-15
[DO NOT MERGE] Break vstest SCI pin to 9.0.11 to trip the coverage scenario test

nohwnd:
Not for merge. This is a deliberate regression to check that the new `dotnet test` coverage scenarios (dotnet/scenario-tests#348) actually catch the System.Collections.Immutable binding redirect mismatch, so we know the guard works before we rely on it.

It is the reverse of #7454 (and #7192 on 4xx): vstest goes back to shipping System.Collections.Immutable 9.0.11 while the three vstest app.configs (vstest.console, datacollector, testhost.x86) still redirect every load to 10.0.0.0. That is the FileLoadException mismatch we fixed.

I expect the Windows-only `VerifyMSTestVSTestNetFrameworkWithCoverage` scenario to go red, because the net48 VSTest leg runs through vstest.console.exe where the app.config redirect applies. The MTP legs and the .NET (non-Framework) legs do not go through that redirect, so they should stay green.

Will close once we have seen the scenario fail.

Related: #7454, #7192, dotnet/scenario-tests#348

### DIFF
--- a/src/vstest/eng/Version.Details.props
+++ b/src/vstest/eng/Version.Details.props
@@ -13,7 +13,7 @@ This file should be imported by eng/Versions.props
     <MicrosoftDiagnosticsNETCoreClientPackageVersion>0.2.0-preview.26360.111</MicrosoftDiagnosticsNETCoreClientPackageVersion>
     <!-- dotnet-runtime dependencies -->
     <MicrosoftExtensionsDependencyModelPackageVersion>6.0.2</MicrosoftExtensionsDependencyModelPackageVersion>
-    <SystemCollectionsImmutablePackageVersion>10.0.0</SystemCollectionsImmutablePackageVersion>
+    <SystemCollectionsImmutablePackageVersion>9.0.11</SystemCollectionsImmutablePackageVersion>
     <!-- dotnet-symreader-converter dependencies -->
     <MicrosoftDiaSymReaderConverterPackageVersion>1.1.0-beta2-19575-01</MicrosoftDiaSymReaderConverterPackageVersion>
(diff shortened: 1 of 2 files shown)


nohwnd:
Closing. This was a deliberate poison PR to check that the new `dotnet test` coverage scenario tests catch a binding-redirect regression, and they do.

I reversed #7454 so the VMR ships System.Collections.Immutable 9.0.11 while the vstest app.configs still redirect it to 10.0.0.0. The interesting part is that the test failed on a sub-dependency, not on SCI directly:

- SCI 10.0.0 depends on System.Memory 4.6.3 (assembly 4.0.5.0)
- SCI 9.0.11 depends on System.Memory 4.5.5 (assembly 4.0.1.2)

The vstest.console app.config redirects System.Memory `0.0.0.0-4.0.5.0` to `4.0.5.0`. With SCI downgraded, the deployed System.Memory is 4.0.1.2 but the redirect forces the reference up to 4.0.5.0, so the net48 loader throws `Could not load file or assembly 'System.Memory, Version=4.0.5.0' ... manifest definition does not match` (HRESULT 0x80131040). Same class of bug as #7192/#7454, one level down the dependency tree.

Failure isolation matches the design: only `VMR Vertical Build Windows_x64` failed, in `VerifyMSTestVSTestNetFrameworkWithCoverage` (net48 + VSTest). The net48 MTP leg and every non-Framework leg passed, because MTP does not use the vstest.console binding-redirect path.

Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1510219

Net: the coverage scenario test does what we wanted. It catches an SCI-driven binding-redirect break, including when it surfaces on a transitive dependency.

PR #7759, opened 2026-07-15
Use managed FileCatalog task for .js catalog generation

akoeplinger:
Replace the Windows-only makecat.exe/generate-catalog.ps1 approach with the cross-platform managed GenerateFileCatalog task from Arcade.

- emsdk: generate emscripten-js.cat via GenerateFileCatalog on all platforms; drop generate-catalog.ps1, swap Templating dependency for FileCatalog.
- runtime: generate runtime-js.cat for wasm runtime packs, hooked after _GenerateFrameworkList and packaged under data/; add FileCatalog dependency and sign the .cat explicitly by name.


src/emsdk/eng/emsdk.proj
```diff
@@ -1,4 +1,9 @@
 <Project Sdk="Microsoft.Build.Traversal">
+  <!-- FileCatalog is restored via eng/Tools.props but, unlike a normal SDK project, the
+       Traversal build does not auto-import package build assets, so import the UsingTask
+       for GenerateFileCatalog explicitly. -->
+  <Import Project="$(NuGetPackageRoot)\microsoft.dotnet.build.tasks.filecatalog\$(MicrosoftDotNetBuildTasksFileCatalogVersion)\build\Microsoft.DotNet.Build.Tasks.FileCatalog.props" />
```
ViktorHofer: This is an anti-pattern that we flag in our skills. It would break whenever the package layout changes. Why can you do the normal thing and just define a PackageReference so that the file gets imported automatically? This is a NET SDK style project (Traversal imports the Microsoft.NET.Sdk) so it's fully restore-able.

ViktorHofer: I see that it's defined in Build.props as a ProjectToBuild item so it should already get restored today.

akoeplinger: I tried it and it didn't work, see previous commit and CI failure before that. I'm not worried about layout changes since these will show up on arcade bump PRs anyway.

ViktorHofer (approved):

src/emsdk/eng/emsdk.proj
akoeplinger: fixed

PR #7766, opened 2026-07-15
[release/10.0.1xx] Fix nuget audit failures for msbuild dependencies

mmitche:
Resolves the NuGet audit (NU1902/NU1903) failures in the `dotnet-unified-build` (build 3022982).

**Changes (all under `src/msbuild`):**
- Bump `System.Security.Cryptography.Xml` 9.0.0 → 9.0.15 (fixes GHSA-37gx-xxp4-5rgx, GHSA-w3x6-4m5h-cxqf; high).
- Pin transitive `OpenTelemetry.Api` to 1.15.3 (fixes GHSA-g94r-2vxg-569j / CVE-2026-40894; moderate). The `Microsoft.VisualStudio.OpenTelemetry.*` packages still reference the vulnerable 1.9.0 even at their latest version (1.0.2), so a direct pin is required.

The `SQLitePCLRaw` advisory (GHSA-2m69-gcr7-jv3q) is intentionally **not** addressed here — its only fix is a major 2.x → 3.x jump and will be handled separately if necessary.

Validation official build (dnceng/internal `dotnet-unified-build`): https://dev.azure.com/dnceng/internal/_build/results?buildId=3023440

### DIFF
--- a/src/msbuild/eng/Packages.props
+++ b/src/msbuild/eng/Packages.props
@@ -42,6 +42,7 @@
     <PackageVersion Include="xunit.console" Version="$(XUnitVersion)" />
     <PackageVersion Include="Microsoft.VisualStudio.OpenTelemetry.Collector" Version="$(MicrosoftVisualStudioOpenTelemetryVersion)" />
     <PackageVersion Include="Microsoft.VisualStudio.OpenTelemetry.ClientExtensions" Version="$(MicrosoftVisualStudioOpenTelemetryVersion)" />
+    <PackageVersion Include="OpenTelemetry.Api" Version="$(OpenTelemetryApiVersion)" />
 
     <!-- Microsoft.VisualStudio.SolutionPersistence is maintained in eng/dependabot/Packages.props -->
   </ItemGroup>
--- a/src/msbuild/eng/Version.Details.props
+++ b/src/msbuild/eng/Version.Details.props
@@ -18,7 +18,7 @@ This file should be imported by eng/Versions.props
     <SystemResourcesExtensionsPackageVersion>9.0.0</SystemResourcesExtensionsPackageVersion>
     <SystemSecurityCryptographyPkcsPackageVersion>9.0.6</SystemSecurityCryptographyPkcsPackageVersion>
     <SystemSecurityCryptographyProtectedDataPackageVersion>9.0.6</SystemSecurityCryptographyProtectedDataPackageVersion>
-    <SystemSecurityCryptographyXmlPackageVersion>9.0.0</SystemSecurityCryptographyXmlPackageVersion>
+    <SystemSecurityCryptographyXmlPackageVersion>9.0.15</SystemSecurityCryptographyXmlPackageVersion>
     <SystemTextEncodingCodePagesPackageVersion>9.0.0</SystemTextEncodingCodePagesPackageVersion>
     <SystemTextJsonPackageVersion>9.0.0</SystemTextJsonPackageVersion>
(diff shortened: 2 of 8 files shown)


mmitche:
@rainersigwald and msbuild folks....take a look and make sure none of those are getting bumped over their VS maximums.

rainersigwald (approved):
18.0 is not in support for VS so it doesn't matter :)

mmitche:
The  System.Memory  assembly bump to  4.0.5.0  landed in 1.14.0, but the advisory (GHSA-g94r-2vxg-569j) isn't fixed until 1.15.3. So there's no overlap — every non-vulnerable version (≥ 1.15.3) hard-binds  System.Memory 4.0.5.0  and will collide with MSBuild's frozen 4.5.5 on net472. The last version that keeps the compatible  4.0.1.2  is 1.13.1, which is still vulnerable.

mmitche:
@rainersigwald In order to resolve the OTel issue we'll have to update the System.Memory version to 4.6.3..any concerns there?

rainersigwald:
Nope, the "frozen" stuff is VS-only so change whatever in this branch.

mmitche:
Pushed the System.Memory fix (commit 37e1fe4a85a).

**What changed** (per @rainersigwald's OK to change the frozen versions on this branch):
- Bumped the frozen net472 compile references so `Microsoft.Build.Framework` compiles against System.Memory **4.6.3** (assembly 4.0.5.0) instead of 4.5.5 (4.0.1.2), matching what OpenTelemetry.Api 1.15.3 hard-binds. This resolves the MSB3277 conflict.
- Bumped the companion frozen packages to exactly the versions System.Memory 4.6.3 declares as dependencies: System.Buffers 4.6.1, System.Numerics.Vectors 4.6.1, System.Runtime.CompilerServices.Unsafe 6.1.2 (and switched their frozen lib paths from net461/net46 to net462, which is the only TFM shipped by the 4.6.x packages).
- Aligned the distributed (non-source-build) versions and the `app.config` / `app.amd64.config` binding redirects + codeBase entries to the same assembly versions (4.0.5.0 / 4.1.6.0 / 6.0.3.0) so runtime resolution stays consistent.

I confirmed via the failing build's log that System.Memory was the **only** MSB3277 conflict; DiagnosticSource (which 1.15.3 also pulls to 10.0.0) resolves fine.

Queued official validation build (internal) `3024287` and the public PR build is running on this push.

mmitche:
@rainersigwald ... https://github.com/dotnet/dotnet/pull/7766/changes/e52e6c9c44bf3f0bd4ca09dfffba6949e9d09de8

Now an upgrade to 10.0.0...makes me want to look at that OTel suppression a bit more.

dkurepa (approved):

dkurepa:
@ViktorHofer do you think we should merge this as is, or try to go with the suppression for the Open Telemetry?

dkurepa:
the alternative doesn't seem to be green. Can we merge this PR to get 1xx unblocked?

ViktorHofer (changes requested):
Left a comment in https://github.com/dotnet/dotnet/pull/7821#issuecomment-5021998347. I don't think that this PR or the alternative one are the right paths. I left a proposal in that other comment with in my opinion the right solution for the current state that we are in.

For future it would be really to good to avoid these kind of issues by offering a real targeting pack solution for msbuild. These days with package pruning, 90% of nuget audit warnings/errors are due to msbuild and should be considered false positives.

PR #7767, opened 2026-07-15
Update VersionPrefix to 10.0.303

### LINKED ISSUE #7751: Microsoft.SourceLink.Common version does not align with .NET SDK version
The versions of Microsoft.SourceLink.Common and friends (Microsoft.SourceLink.GitHub, Microsoft.Build.Tasks.Git, etc.) should align with the latest .NET SDK version, but it is behind: 10.0.301 instead of 10.0.302.

mmitche:


ViktorHofer (approved):

akoeplinger:
looks like @vseanreesermsft's automation just does +1 on the version number, I think we should add some validation that  the resulting versions align since we should always have the same

Issue #7758, opened 2026-07-15
[Operational Issue]: NuGet Audit failures in 10.0.1xx & 10.0.3xx


ViktorHofer:
### Error

https://dev.azure.com/dnceng/internal/_build/results?buildId=3023197&view=logs&j=9050e078-31bf-5111-d8ec-8b6fa95caf9c&t=250cd8ff-81de-528c-ae44-9d7a96939fe2&l=366

### Root Cause

unknown

### Blocking

- [ ] This issue is blocking

### Affected Releases

- [ ] .NET 10
- [ ] .NET 10 (SDK band build)
- [ ] .NET 11

dkurepa:
opened https://github.com/dotnet/msbuild/issues/14398 to deal with this on the msbuild side

PR #7781, opened 2026-07-16
Bump flagged packages

dkurepa:
https://github.com/dotnet/dotnet/issues/7758


ViktorHofer (changes requested):
This looks wrong, we shouldn't reference such old packages: Regex and System.Net.Http. They should get pruned automatically already. There's also a complication in msbuild that hinders us from updating our runtime dependencies freely. Those changes need to be coordinated with VS.

I think it would be best to file an issue in dotnet/msbuild and we will take a look and do the work there.

ViktorHofer:
cc @AlesProkop & @rainersigwald

JanProvaznik:
10.0.3xx => VS18.6 which is out of support so the VS complication does not apply

PR #7788, opened 2026-07-16
Forward NuGet auth into installer test containers

### LINKED ISSUE #7787: [Operational Issue]: ValidateInstallers scenario tests fail with NU1301 / 401 Unauthorized on internal feeds (auth not passed into Docker container)
### Error

The `ValidateInstallers_Linux_arm64` leg (step **Validate installer packages**) fails during the SDK scenario tests, which run **inside a Docker container** (`Microsoft.DotNet.Installer.Tests` → `DistroTest` → `docker run`). NuGet restore inside the container cannot authenticate to the internal Azure DevOps feeds and fails with `NU1301` / `401 (Unauthorized)`:

```
SdkTemplateTestsComplex_Console_CSharp.csproj : error NU1301: Unable to load the service index for source
  https://pkgs.dev.azure.com/dnceng/internal/_packaging/darc-int-dotnet-dotnet-f7d90799/nuget/v3/index.json.
SdkTemplateTestsComplex_Console_CSharp.csproj : error NU1301:   Response status code does not indicate success: 401 (Unauthorized).
```

The same `401 (Unauthorized)` occurs for every internal feed referenced by the scenario tests:
- `darc-int-dotnet-dotnet-f7d90799`
- `dotnet10-internal`
- `dotnet10-internal-transport`
- `dotnet-eng-internal`
- `dotnet-tools-internal`

- **Pipeline:** dotnet-unified-build (definition 1330)
- **Branch:** `release/10.0.4xx`
- **Build:** 20260715.18 (buildId 3023455)

### Links

- Failing build (leg `ValidateInstallers_Linux_arm64`, task *Validate installer packages*): https://dev.azure.com/dnceng/internal/_build/results?buildId=3023455&view=logs&j=a6a3a584-660a-5ea0-ec00-3008ec772a89&t=adc9f0dc-a0b8-5f3a-de57-ecb6a3ddbae7
- Similar fix (same 401 internal-feed-auth class, forwarding the NuGet credential-provider environment across the `sudo` boundary for offline validation test restore): https://github.com/dotnet/dotnet/pull/7696

### Root Cause

Following the PAT-less internal feed auth migration (#7686, #7688), the Azure Artifacts NuGet credential provider authenticates using a provisioned session token located via environment (e.g. `HOME` / credential-provider session token env). The SDK installer scenario tests run the restore **inside a Docker container** (`LinuxInstallerTests.DistroTest` builds an image and invokes `DockerHelper.Run` → `docker run`). The container is started without forwarding the credential-provider environment (the `docker run` only mounts the log directory via `-v`), so restore inside the container falls back to unauthenticated and returns `401 (Unauthorized)` against the internal feeds.

This is the Docker-container analog of #7696, which fixed the same failure for the offline validation legs by preserving the credential-provider environment across `sudo` (`--preserve-env=HOME`). The Docker path needs the equivalent env-var passthrough into the container (e.g. `docker run -e ...`).

### Blocking

- [ ] This issue is blocking

### Affected Releases

- [ ] .NET 10
- [x] .NET 10 (SDK band build)
- [ ] .NET 11

mmitche:
Fixes #7787

## Root cause
`ValidateInstallers_Linux_*` runs the SDK installer scenario tests inside Docker. After the PAT-less internal-feed-auth migration, the host job is authenticated by `NuGetAuthenticate@1`, but the container did not receive the Azure Artifacts credential provider environment, plugin, or session-token cache. Restores from internal feeds inside the container therefore fell back to unauthenticated requests and failed with NU1301 / 401.

## Fix
When NuGet auth environment is present on the host, forward the relevant Azure Artifacts credential provider variables into `docker run` by name (without logging secret values), and mount the host credential provider plugin plus session-token cache into the container. Public builds with no auth environment keep the existing behavior.

This is the Docker-container analog of #7696, which preserved `HOME` across `sudo` so the credential provider could locate its cached credentials.

## Target branch
This PR targets `release/10.0.4xx`, matching the original failing official build (3023455). After validation, equivalent fixes may be needed in other affected 10.0 release branches.

## Validation
- `./build.cmd -ci -projects C:\d\test\Microsoft.DotNet.Installer.Tests\Microsoft.DotNet.Installer.Tests.csproj /p:TestRpmPackages=true /p:TestDebPackages=true`
- Official validation queued against internal branch `release/10.0.4xx-installer-tests-docker-feed-auth`: https://dev.azure.com/dnceng/internal/_build/results?buildId=3024133

### DIFF
--- a/test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
+++ b/test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
@@ -90,6 +90,18 @@ public partial class LinuxInstallerTests : IDisposable
     private const string DotnetApphostPackPrefix = "dotnet-apphost-pack-";
     private const string DotnetSdkPrefix = "dotnet-sdk-";
     private const string DowngradeFxVersionsScript = "downgrade-fx-versions.sh";
+    private const string ContainerNuGetPluginsDir = "/root/.nuget/plugins";
+    private const string ContainerCredentialProviderCacheDir = "/root/.local/share/MicrosoftCredentialProvider";
+    private static readonly string[] NuGetAuthEnvironmentVariables =
+    [
+        "ARTIFACTS_CREDENTIALPROVIDER_EXTERNAL_FEED_ENDPOINTS",
+        "ARTIFACTS_CREDENTIALPROVIDER_FEED_ENDPOINTS",
+        "ARTIFACTS_CREDENTIALPROVIDER_SESSIONTOKENCACHE_ENABLED",
+        "NUGET_CREDENTIALPROVIDER_SESSIONTOKENCACHE_ENABLED",
+        "VSS_NUGET_ACCESSTOKEN",
+        "VSS_NUGET_EXTERNAL_FEED_ENDPOINTS",
+        "VSS_NUGET_URI_PREFIXES"
+    ];
 
     public static bool IncludeRpmTests => Config.TestRpmPackages;
     public static bool IncludeDebTests => Config.TestDebPackages;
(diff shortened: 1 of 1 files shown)


mmitche:
Build is green with this.

ViktorHofer:
Obvious question, we forward $HOME in other places. Why don't we do that here? Because it's more correct?

mmitche:
> Obvious question, we forward $HOME in other places. Why don't we do that here? Because it's more correct?

I was wondering the same thing...I don't know why HOME would be required for the credential provider to work. Maybe if HOME was mapped into the container?

mmitche:
@ViktorHofer The reason that the other fix does not apply here:
- In the other case, NuGetAUthenicate has already happened within the docker environment (because SB uses the native container support). However, sudo resets that environment. This means that all the variables that NuGetAuthenticate sets are still around.
- In the installer case, the docker run command for the specific installer tests is executed manually. NuGetAuthenticate happens outside the container, and so the env vars need to be mapped in.

ViktorHofer (approved):

PR #7793, opened 2026-07-16
Reenable DeterministicTimestamp in build

### LINKED ISSUE #7331: Re-enable deterministic timestamping after resolving NuGet issue
https://github.com/NuGet/NuGet.Client/pull/7020 introduced an issue where file locks are incorrectly held, leading to failures in signing when copying files. See https://github.com/NuGet/NuGet.Client/pull/7020#issuecomment-4781601563 for a potential implementation. After resolution and a rebootstrap, Re-enable deterministic timestamps.

akoeplinger:
Fixes https://github.com/dotnet/dotnet/issues/7331


mthalman (approved):

ViktorHofer:
Do you want to remove the opt-out switch from the ported arcade-validation code as well which now lives in arcade?

akoeplinger:
Ah good point, yes. I didn't remember because arcade-validation main still has the old sources in it, we should archive it.

ViktorHofer (approved):

PR #7821, opened 2026-07-17
Alternative: drop net472 from Microsoft.Build.Framework to avoid vulnerable OpenTelemetry.*

mmitche:
## Alternative approach: drop net472 from `Microsoft.Build.Framework`

This is an **alternative** to #7766 for resolving the NuGet audit failures (`NU1902`/`NU1903`). The novel part is how the OpenTelemetry advisory (**GHSA-g94r-2vxg-569j / CVE-2026-40894**) is handled: instead of upgrading `OpenTelemetry.Api`, drop `net472` targeting.

### OpenTelemetry advisory - the alternative approach
`OpenTelemetry.Api` is pulled in transitively by `Microsoft.VisualStudio.OpenTelemetry.Collector` / `Microsoft.VisualStudio.OpenTelemetry.ClientExtensions`, which live in the `.NETFramework`-only `ItemGroup` in `Microsoft.Build.Framework.csproj`. By targeting only `.NET Core + netstandard2.0` (dropping `net472`), the vulnerable dependency is never restored - **no `OpenTelemetry.Api` upgrade**.

This sidesteps the cascade the OTel upgrade triggers in #7766: net472 `MSB3277` assembly-version conflicts (`System.Memory` -> `System.Diagnostics.DiagnosticSource`), each needing frozen-maintenance-package bumps + binding-redirect updates.

### Remaining audit fixes (folded in, approach-independent)
Same bumps as in #7766 so this PR resolves **all** reported audit failures on its own:
- `System.Security.Cryptography.Xml` 9.0.0 -> 9.0.15 (GHSA-37gx-xxp4-5rgx, GHSA-w3x6-4m5h-cxqf)
- `System.Formats.Asn1` 9.0.6 -> 9.0.15
- `System.Security.Cryptography.Pkcs` 9.0.6 -> 9.0.15

### Trade-off / caveat
The `net472` flavor of `Microsoft.Build.Framework` is consumed by the rest of MSBuild's `net472` (Visual Studio) build. Dropping it here will break the net472 build of the rest of MSBuild, so this is offered **for evaluation** of the approach vs. #7766, not a drop-in change.

cc @rainersigwald

### DIFF
--- a/src/msbuild/eng/Version.Details.props
+++ b/src/msbuild/eng/Version.Details.props
@@ -11,14 +11,14 @@ This file should be imported by eng/Versions.props
     <SystemConfigurationConfigurationManagerPackageVersion>9.0.0</SystemConfigurationConfigurationManagerPackageVersion>
     <SystemDiagnosticsDiagnosticSourcePackageVersion>9.0.0</SystemDiagnosticsDiagnosticSourcePackageVersion>
     <SystemDiagnosticsEventLogPackageVersion>9.0.0</SystemDiagnosticsEventLogPackageVersion>
-    <SystemFormatsAsn1PackageVersion>9.0.6</SystemFormatsAsn1PackageVersion>
+    <SystemFormatsAsn1PackageVersion>9.0.15</SystemFormatsAsn1PackageVersion>
     <SystemFormatsNrbfPackageVersion>9.0.0</SystemFormatsNrbfPackageVersion>
     <SystemReflectionMetadataPackageVersion>9.0.0</SystemReflectionMetadataPackageVersion>
     <SystemReflectionMetadataLoadContextPackageVersion>9.0.0</SystemReflectionMetadataLoadContextPackageVersion>
     <SystemResourcesExtensionsPackageVersion>9.0.0</SystemResourcesExtensionsPackageVersion>
-    <SystemSecurityCryptographyPkcsPackageVersion>9.0.6</SystemSecurityCryptographyPkcsPackageVersion>
+    <SystemSecurityCryptographyPkcsPackageVersion>9.0.15</SystemSecurityCryptographyPkcsPackageVersion>
     <SystemSecurityCryptographyProtectedDataPackageVersion>9.0.6</SystemSecurityCryptographyProtectedDataPackageVersion>
-    <SystemSecurityCryptographyXmlPackageVersion>9.0.0</SystemSecurityCryptographyXmlPackageVersion>
+    <SystemSecurityCryptographyXmlPackageVersion>9.0.15</SystemSecurityCryptographyXmlPackageVersion>
     <SystemTextEncodingCodePagesPackageVersion>9.0.0</SystemTextEncodingCodePagesPackageVersion>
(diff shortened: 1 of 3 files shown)


baronfel:
I was under the impression that today VS is the primary consumer of our OTel-based metrics, via the VS-only OTel Collectors we have? If that is the case then not including OTel deps seems like it would defeat the purpose. cc @JanProvaznik

mmitche:
> I was under the impression that today VS is the primary consumer of our OTel-based metrics, via the VS-only OTel Collectors we have? If that is the case then not including OTel deps seems like it would defeat the purpose. cc @JanProvaznik

VS 18.0 isn't in support any longer, and we don't include the net framework version of msbuild in the SDK (I think)

JanProvaznik:
I don't like this because if we needed to service 18.0 nugets (still 2+ years in support IIRC) the net472 should remain inside even though that VS is out of support. But I am not sure what's our exact support contract with nugets.

JanProvaznik:
and @baronfel VS is no longer the consumer because we removed otel in ~18.4

rainersigwald:
> I don't like this because if we needed to service 18.0 nugets (still 2+ years in support IIRC) the net472 should remain inside even though that VS is out of support. But I am not sure what's our exact support contract with nugets.

We build & ship the NuGet packages from our repo so we can always revive the TF and reship if we need to for a security event.

ViktorHofer:
I also don't like this approach for the stated reason but also for a another one. These change automatically flow forward to newer in support branches: vs18.0 -> vs18.6. We would need to revert these which is fine but unusual.

I would instead silence that NuGet Audit report via the `NuGetAuditSuppress` item in msbuild. Any other repos that are impacted should like switch to only consume the compile time asset either via PackageDownload+Reference (as runtime and roslyn do), by picking the asset directly from within the SDK or silence the warning also.

mmitche:
> I also don't like this approach for the stated reason but also for a another one. These change automatically flow forward to newer in support branches: vs18.0 -> vs18.6. We would need to revert these which is fine but unusual.
> 
> I would instead silence that NuGet Audit report via the `NuGetAuditSuppress` item in msbuild. Any other repos that are impacted should like switch to only consume the compile time asset either via PackageDownload+Reference (as runtime and roslyn do), by picking the asset directly from within the SDK or silence the warning also.

Is that the correct fix for OTel? I understand with the S.S.C.Xml since msbuild is only referencing surface area..but is that same thing with OTel?

ViktorHofer:
Right, that is only correct for compile-time only dependencies.

ViktorHofer:
I think stop building the TFM is a big hammer. I don't want to see that code flowing forward anywhere, accidentally. Suppressing the audit report sounds like the better option if there aren't any downstream consumers of it.

mmitche:
Hmm. If we update OTel, it pulls SystemDiagnosticsDiagnosticSourcePackageVersion forward from 9.0 to 10.0.

ViktorHofer:
OTel is also just a compile time dependency, I just checked. Suppress the warning.

PR #7824, opened 2026-07-17
Fix dnx ownership in native Linux packages

mthalman:
## Summary

Released .NET 10 SDK native packages own the public dnx entries, which conflicts when package resolution upgrades the shared dotnet-host package to .NET 11. This makes the shared host authoritative for those entries while preserving side-by-side SDK 10 and SDK 11 use.

- Extend Arcade RPM authoring with ghost files, file triggers, and post-uninstall script support.
- Ship a private authoritative dispatcher in dotnet-host and repair the public dispatcher and symlink from package lifecycle scripts.
- Use bounded DEB Replaces metadata plus path triggers, and RPM ghost ownership plus a transaction file trigger.
- Clean generated public entries only when the final host package is removed.
- Exercise the generated host package against the unpinned, repository-selected dotnet-sdk-10.0 package on Azure Linux 3 and Debian 12.

## Design notes
The RPM package ghost-owns /usr/bin/dnx and /usr/share/dotnet/dnx so it can coexist with released SDK 10 packages without payload conflicts. A transaction file trigger repairs the entries after SDK installation or servicing.

The DEB package uses Replaces: dotnet-sdk-10.0 (<< 10.1.0) to transfer ownership only from affected SDK 10 packages. Its post-install script and path triggers provide the same idempotent repair behavior.

The lifecycle tests accept both the legacy SDK layout and a future SDK package that already uses the repaired layout. The current Azure Linux package exercises the legacy broken direct invocation, while the current Debian package verifies forward-compatible interoperability.

## Testing

- Arcade installer task tests: 13 passed.
- VMR installer test project builds with zero warnings and errors.
- RPM and DEB lifecycle tests passed through the xUnit Docker harness on Azure Linux 3 and Debian 12.
- RPM ghost-to-real-file transition and package ordering scenarios passed on Fedora 42 and Azure Linux 3.
- DEB install ordering, replacement bounds, reinstall, removal, and cleanup scenarios passed on Debian 12.

> [!NOTE]


test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -90,6 +90,9 @@ public partial class LinuxInstallerTests : IDisposable
     private const string DotnetApphostPackPrefix = "dotnet-apphost-pack-";
     private const string DotnetSdkPrefix = "dotnet-sdk-";
     private const string DowngradeFxVersionsScript = "downgrade-fx-versions.sh";
+    private const string DnxPackageLifecycleScript = "dnx-package-lifecycle.sh";
+    private const int RpmFileGhost = 1 << 6;
+    private const int RpmTransFileTriggerNameTag = 5079;
```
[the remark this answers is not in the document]
mthalman: This cannot use `RpmHeaderTag.TransFileTriggerName` yet because the VMR test project compiles against the bootstrap Arcade package (`11.0.0-beta.26359.118`), whose enum predates this tag. Using it fails with CS0117, so the numeric RPM tag is required here until the bootstrap package advances.

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -643,6 +729,51 @@ private void ValidatePackageMetadata(string image, PackageType packageType)
     {
         List<string> list = GetPackageList(image, packageType);
         ValidatePackageDependencies(list, packageType);
+        ValidateDnxPackageMetadata(packageType);
+    }
+
+    private void ValidateDnxPackageMetadata(PackageType packageType)
+    {
+        string hostPackage = GetContentPackage(DotnetHostPrefix, packageType);
+        string hostPackagePath = Path.Combine(_contextDir, hostPackage);
```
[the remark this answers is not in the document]
mthalman: Fixed in bebdb00.

mthalman:
Reviewers: this will need to get in for Preview 7 so prioritize this appropriately.

src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/installer.build.targets
```diff
@@ -125,7 +125,7 @@
 
     <MSBuild Projects="$(MSBuildProjectFullPath)"
              Targets="PublishToDisk"
-             Properties="OutputPath=$([MSBuild]::EnsureTrailingSlash('$(_LayoutPackageRoot)'))"
+             Properties="OutputPath=$([MSBuild]::EnsureTrailingSlash('$(_LayoutPackageRoot)'));IsPublishingToLinuxPackage=true"
```
jkoritzinsky: Can we avoid adding another property here? We try to avoid having additional properties set so multiple parallel builds pointing to the same directory don't cause file access issues (and get MSBuild to deduplicate the `PublishToDisk` calls).

Either that or we need to ensure that the Shared Framework SDK strips this property for the `GetFilesToPackage` inner invocation in its `PublishToDisk` implementation.

src/runtime/src/installer/pkg/sfx/installers/dotnet-host.proj
```diff
@@ -51,6 +51,17 @@
...
+        Private, non-conflicting copy of the robust dnx dispatcher used as the repair source by
+        the Linux package post-install and file-trigger scriptlets (https://github.com/dotnet/sdk/issues/55303).
+        Only laid out for the native Linux packages (via IsPublishingToLinuxPackage, set only by
+        the deb/rpm layout target); macOS/archive/MSI behavior is unchanged. Note that
+        GenerateDeb/GenerateRpm are stripped from this nested PublishToDisk invocation, so they
+        cannot be used here.
+      -->
+      <FilesToPublish Include="$(MSBuildThisFileDirectory)dnx"
+            Destination="$(OutputPath)dnx.dispatcher"
+            Condition="'$(TargetsUnix)' == 'true' and '$(IsPublishingToLinuxPackage)' == 'true'" />
```
jkoritzinsky: Instead of using `IsPublishingToLinuxPackage`, can we introduce an "installer only" metadata as a more general concept?

jkoritzinsky (commented):
Some concerns on the MSBuild targets logic. Everything else looks good.

src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/installer.build.targets
mthalman: Fixed in 23b8ff3

src/runtime/src/installer/pkg/sfx/installers/dotnet-host.proj
mthalman: Fixed in 23b8ff3

jkoritzinsky (approved):

PR #7873, opened 2026-07-21
Make release/10.0.1xx buildable locally on Windows

ViktorHofer:
- Resolve issues when a matching SDK version is globally installed
- Define an sdk::version entry in all global.json so that the host picks the correct SDK version (the one determined by the VMR root).
- Use of TaskHostFactory in runtime to avoid file lock issues between the stage1 and stage2 build.

With that, everything except the vcxproj toolchain (wpf + aspnetcore) builds fine. Those two require VS2022 installed.


src/runtime/eng/testing/tests.targets
```diff
@@ -99,7 +99,7 @@
     <RemoveDir Condition="'$(ContinuousIntegrationBuild)' == 'true' and '$(OS)' != 'Windows_NT'" Directories="$(PublishDir)" ContinueOnError="WarnAndContinue" />
   </Target>
 
-  <UsingTask TaskName="GenerateRunScript" AssemblyFile="$(InstallerTasksAssemblyPath)"/>
+  <UsingTask TaskName="GenerateRunScript" AssemblyFile="$(InstallerTasksAssemblyPath)" TaskFactory="TaskHostFactory"/>
```
akoeplinger, line 102: I assume these changes are unrelated?

akoeplinger:
FYI nuget.client recently removed the `sdk` entry from their global.json: https://github.com/NuGet/NuGet.Client/pull/7182

src/runtime/eng/testing/tests.targets
ViktorHofer: Intentional changes. Without them runtime couldn't build due to file lock errors.

ViktorHofer:
> FYI nuget.client recently removed the sdk entry from their global.json: https://github.com/NuGet/NuGet.Client/pull/7182

Doesn't matter for 1xx as they still have it there. I will need to fight the battle for 10.0.3xx onwards with them though. I prefer adding the entry to each repo over forcing the VMR to always add it as it's in general the right thing to do. But I might need to special case the VMR yet again for nuget-client. EricStj and I let them know multiple times that the sdk::version entry is essential. Otherwise you would run into these exact kind of issues.

akoeplinger (approved):

PR #7875, opened 2026-07-21
Fix building release/10.0.300 and fix NuGet Audit

ViktorHofer:
Summary: I decided to backport the change that enabled NuGet Audit as part of the dev/PR validation instead of in official builds (by adding the audit source to the NuGet.config files). With that, we can be confident that the fix is working.

I'm also backporting a change from EricStj from main to be able to pass -warnNotAsErrors <codes> in so that downgrading errors to warnings for NuGet Audit codes works as expected. I'm using that new switch in fsharp, nuget-client and razor as I wasn't able to mitigate the flagged vulnerability reports in there. They are all coming from msbuild packages and are therefore false positives. Those repos make it really hard to update the vulnerable transitive dependency (S.S.Cryptography.Xml) as they either don't use CPM or don't have transitive pinning enabled.

I fixed the flagged vulnerabilities in msbuild and sourcelink by updating dependencies.


mmitche (approved):

ViktorHofer:
PR is ready. Feel free to merge when mostly green (needs force permissions as the "Validate user changes" job is reporting red - ignore).

rbhanda (approved):

PR #7928, opened 2026-07-23
Extend reproducibility test to source-built artifacts

omajid:
Add baselines, exclusions and so on for those as well.


test/Microsoft.DotNet.Reproducibility.Tests/assets/SourceBuiltArtifactsReproducibilityBaseline.diff
```diff
@@ -0,0 +1 @@
+
```
omajid, line 1: The baseline is obviously not empty at the moment. What's a good way to populate these files?

mthalman:
@omajid - I made some fixes after testing out the changes.

mthalman (approved):

omajid:
Thanks!

ellahathaway (approved):

PR #7940, opened 2026-07-24
Allow submodule changes when reseting submodules

dkurepa:
https://github.com/dotnet/arcade-services/issues/6443

### DIFF
--- a/eng/tools/ChangeValidation/SubmoduleValidation.cs
+++ b/eng/tools/ChangeValidation/SubmoduleValidation.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using Microsoft.Extensions.FileSystemGlobbing;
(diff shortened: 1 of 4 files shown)


eng/tools/ChangeValidation/SubmoduleValidation.cs
```diff
@@ -37,39 +38,102 @@ public SubmoduleValidation(
...
+                !string.Equals(baseEntry.RemoteUri, headEntry.RemoteUri, StringComparison.Ordinal))
+            {
+                changedEntryPaths.Add(path);
+            }
         }
 
-        var directory = new InMemoryDirectoryInfo("/", prInfo.ChangedFiles);
+        // Determine, per submodule, which of the PR's changed files belong to that submodule.
+        var changedFilesDirectory = new InMemoryDirectoryInfo("/", prInfo.ChangedFiles);
+        var submodulesWithFileChanges = new Dictionary<string, List<string>>(StringComparer.Ordinal);
```
premun, line 84: It would be good to have tests for this (do we have some?).

Cases like submodule changes in other submodules than the changed one etc.

eng/tools/ChangeValidation/ToolingFilesValidation.cs
```diff
@@ -14,7 +14,6 @@ internal class ToolingFilesValidation : IValidationStep
     private static readonly string[] ToolingFilePatterns = new[]
     {
         "^eng/Version\\.Details\\.xml$",
-        "^src/source-manifest\\.json$",
```
premun, line 17: We should still not allow changes to source-manifest outside of the submodule section. Will that happen?

eng/tools/ChangeValidation/SubmoduleValidation.cs
dkurepa: I was trying that locally, not sure if these tools here have tests

eng/tools/ChangeValidation/ToolingFilesValidation.cs
dkurepa: it will get caught here https://github.com/dkurepa/dotnet/blob/cb849c734b07695ca5e39d3de46db707bc9b129d/eng/tools/ChangeValidation/SubmoduleValidation.cs#L104
that someone is trying to change the source-manifest without changing any of the submodules

dkurepa: I realize it's not structured the best, source-manifest validation is happening in submodule validation

dkurepa: maybe it could be improved didn't really dive that deep into the structure of the validation tool

eng/tools/ChangeValidation/SubmoduleValidation.cs
dkurepa: from what I see in the repo, there don't appear to be tests for these tools

premun (approved):

dkurepa:
@dotnet/source-build could I get an approval please?

mthalman (approved):

premun:
@dkurepa this needs to get backported

dkurepa:
> @dkurepa this needs to get backported

yup, was doing it locally, but the bot should work too, gonna try it

PR #7976, opened 2026-07-27
Refresh .config/guardian/.gdnbaselines file

akoeplinger:
New paths after https://github.com/dotnet/dotnet/pull/7908 got merged

PR #7993, opened 2026-07-27
Fix Windows x86 Guardian baselines

mmitche:
Replaces the stale win-x64 BA2007 baseline records with the exact Windows_x86 records generated by unified build 3032528 for DirectWriteForwarder.dll and System.Printing.dll.

### DIFF
--- a/.config/guardian/.gdnbaselines
+++ b/.config/guardian/.gdnbaselines
@@ -300,6 +300,38 @@
       "ruleId": "BA2008",
       "createdDate": "2026-07-27 01:55:14Z"
     },
+    "52381a07121ad8c6d8874d31fcc9ec0c878cdc13a221c23dfb68d6e003c42ffe": {
+      "signature": "52381a07121ad8c6d8874d31fcc9ec0c878cdc13a221c23dfb68d6e003c42ffe",
+      "alternativeSignatures": [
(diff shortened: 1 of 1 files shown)


mmitche:
@akoeplinger Might need both x86 and x64 baselines. Checking.

akoeplinger:
yeah I think you'll need both. I took mine directly from the uploaded .gdnbaselines file from the Windows x64 run

ViktorHofer (commented):
Do you also need different baselines for win-arm64?

dkurepa:
good point, Arm has been failing earlier recently but binskim did fail in https://dev.azure.com/dnceng/internal/_build/results?buildId=3032807&view=logs&j=e45a3293-4024-5a66-d70f-cda470bd6fb5&t=65ae88fa-2931-5e34-26a7-b8ed76bf76f5. will add

dkurepa:
Ah I think Alex already did this in https://github.com/dotnet/dotnet/pull/7976/changes, so no need for arm64 here

ViktorHofer (approved):

dkurepa:
this prob isn't tested in the PR pipeline I imagine? Can we fast merge?

akoeplinger:
it's possible we'll need something for arm64 too but let's see

PR #8056, opened 2026-07-30
Only request reviews on PR open/synchronize

akoeplinger:
This ensures that dotnet-policy-service isn't constantly requesting reviews:
image

### DIFF
--- a/.github/policies/assign_ownership.yml
+++ b/.github/policies/assign_ownership.yml
@@ -9,6 +9,11 @@ configuration:
       if:
       - payloadType: Pull_Request
       - or:
+        - isAction:
+            action: Opened
+        - isAction:
+            action: Synchronize
+      - or:
         - filesMatchPattern:
             pattern: 'src/arcade/.*'
             matchAny: true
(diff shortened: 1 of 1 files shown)


jkoritzinsky (approved):

.github/policies/assign_ownership.yml
```diff
@@ -8,6 +8,11 @@ configuration:
     - description: 
       if:
       - payloadType: Pull_Request
+      - or:
```
mmitche, line 11: This might not do exactly what we want? It looks like it will request for pull requests if opened/synchronized OR the files match a pattern.

akoeplinger: this should be fine, there's an implicit AND between the direct siblings of the if (see also similar cases in aspnetcore: https://github.com/dotnet/aspnetcore/blob/747d2cdb584079a0c7309115979f13c331fb7df7/.github/policies/resourceManagement.yml#L210-L226)

so this reads as:
`if Pull_Request AND (Opened OR Synchronize) AND (filesMatchPattern OR titleContains)`

PR #8060, opened 2026-07-30
Migrate VMR test projects to xunit.v3 and Microsoft.Testing.Platform

ViktorHofer:
Migrates the test projects under `test/` from xUnit.net v2 on VSTest to xUnit.net v3 on Microsoft.Testing.Platform (MTP).

## Build configuration

Arcade's test infrastructure is disabled (`DisableArcadeTestFramework`). MTP contributes its own `Test` → `InvokeTestingPlatform` target that builds the project and executes the test application, so Arcade's parallel `Test`/`RunTests` targets and the xunit package references it injects aren't needed. `eng/Tests.targets` is deleted; it overrode the `Test` target to depend on `VSTest`, which would otherwise shadow MTP's.

That leaves one execution path shared by `build.cmd -test`, `dotnet test` and `dotnet msbuild -t:Test`. `global.json` declares the MTP runner so `dotnet test` uses it too.

The per-project VSTest properties (`VSTestLogger`, `VSTestCLIRunSettings`, `VSTestResultsDirectory`, `VSTestUseMSBuildOutput`) collapse into a single `TestingPlatformCommandLineArguments` property in `test/Directory.Build.props`.

Packages reduce to a single `xunit.v3` reference: it pulls in `xunit.v3.mtp-v2` (core, assert, analyzers), which brings MTP and its MSBuild integration transitively. `Microsoft.Testing.Platform.MSBuild` is pinned forward because `xunit.v3` asks for an older version that predates `--ignore-exit-code`.

Two behaviours are worth calling out:

- **Reports are now requested explicitly.** MTP emits none by default, and every pipeline that publishes results from this repo globs for trx. They go to `artifacts/TestResults` as `<Project>.trx` / `.xml` / `.html` (the xunit writers, so no extra `Microsoft.Testing.Extensions.*` package).
- **All-skipped runs no longer fail.** Many of these tests are conditional on the build configuration (source-only, official build, availability of an SDK archive), and MTP reports "zero tests ran" as exit code 8. Real test failures still fail the build — verified.

## Code changes

The custom `[ConditionalFact]` / `[ConditionalTheory]` attributes are replaced by xunit v3's built-in support:

```csharp
[ConditionalFact(typeof(SdkContentTests), nameof(IncludeSdkContentTests))]
// becomes
[Fact(Skip = "Condition not met: IncludeSdkContentTests", SkipUnless = nameof(IncludeSdkContentTests))]
```

The v2-only discoverer infrastructure they relied on (`ConditionalFactDiscoverer`, `ConditionalTheoryDiscoverer`, `ConditionalTestDiscoverer`, `SkippedTestCase`, …) is deleted — 7 files, ~370 lines.

`using Xunit.Abstractions;` is gone throughout; `ITestOutputHelper` moved into the `Xunit` namespace and gained `Write`/`Output` members, which `DebugTestOutputHelper` now implements.

## Runtime selection

xunit.v3 requires an app host, and MTP launches that executable directly rather than through the `dotnet` muxer. An app host ignores `PATH` — it resolves a runtime through `DOTNET_ROOT` and otherwise falls back to the machine-wide install. Arcade's runner target used to set this; with Arcade's test infrastructure disabled, nothing did, so tests silently ran on the machine-wide runtime instead of the repo-local SDK, and would fail outright on a machine with no .NET installed:

```
no DOTNET_ROOT   -> .NET 11.0.0-preview.7.26377.110   # machine-wide
with DOTNET_ROOT -> .NET 11.0.0-preview.6.26359.118   # repo-local .dotnet/
```

microsoft/testfx#10413 fixes this upstream: `InvokeTestingPlatformTask` now configures `DOTNET_ROOT` for app host launches itself, and the `InvokeTestingPlatform` target gained a `TestingPlatformEnvironmentVariable` item for passing additional variables through. No workaround is needed in this repo.

That fix isn't on the regular feeds yet, so `MicrosoftTestingPlatformVersion` is overridden to a build from the testfx [`test-tools` preview channel](https://aka.ms/mstest/preview), which the root `NuGet.config` now references. **Both should be reverted once a release containing the fix flows to the regular feeds.**

## Pipelines

`sdk-diff-tests.yml` and `vmr-license-scan.yml` invoke `dotnet test` directly rather than going through Arcade, so their VSTest arguments are translated:

| before | after |
|---|---|
| `--filter "Category=SdkContent"` | `--filter-trait "Category=SdkContent"` |
| `--filter "FullyQualifiedName=..."` | `--filter-method "..."` |
| `--logger:'console;verbosity=detailed'` | `--output Detailed` |
| `--logger:'trx;LogFileName=...'` | (now a default in `test/Directory.Build.props`) |

MTP has no unified `--filter`; the platform itself only provides `--filter-uid`, and the `--filter-class/method/namespace/trait` options come from xunit.

## Validation

All five projects build with zero warnings. `-t:Test` and `dotnet test` both pass, emit trx/xml/html into `artifacts/TestResults`, and leave nothing behind next to the test binaries. Verified that an all-skipped project exits 0 while a genuine failure still exits non-zero, and that tests are hosted on the repo-local SDK with no `DOTNET_ROOT` set in the environment.


src/global.json
```diff
@@ -0,0 +1 @@
+{}
```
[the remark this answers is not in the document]
ViktorHofer: This is intentional. No one should invoke a command from the src directory.

global.json
```diff
@@ -1,10 +1,18 @@
 {
+  "sdk":{
+    "version": "11.0.100-preview.6.26359.118",
```
Evangelink, line 3: **Blocking:** this SDK pin breaks the offline current-source-built SDK leg.

That leg extracts and selects `11.0.100-ci`, but its first repository `dotnet` invocation discovers this `global.json`, requests `11.0.100-preview.6.26359.118`, reports that no compatible SDK is available, and exits with code 155 before restore or tests begin. This is the current failure on head `df8eb33`: [raw log](https://dev.azure.com/dnceng-public/public/_apis/build/builds/1538777/logs/106).

`allowPrerelease` plus `latestFeature` does not make the supplied `11.0.100-ci` bootstrap SDK a valid roll-forward candidate for this pinned preview version. Please avoid pinning the root SDK here, or otherwise preserve compatibility with the source-built bootstrap SDK.

global.json
```diff
@@ -1,10 +1,18 @@
...
+    "allowPrerelease": true
+  },
   "tools": {
     "dotnet": "11.0.100-preview.6.26359.118"
   },
   "msbuild-sdks": {
     "Microsoft.Build.NoTargets": "3.7.0",
     "Microsoft.Build.Traversal": "3.4.0",
     "Microsoft.DotNet.Arcade.Sdk": "11.0.0-beta.26359.118"
+  },
```
Evangelink, line 14: **Blocking:** the repository's change-validation policy unconditionally rejects modifications to the root `global.json`.

`eng/tools/ChangeValidation/ToolingFilesValidation.cs` includes `^global\.json$` in `ToolingFilePatterns`; when matched, validation logs an error and returns `false` with no exception path. The latest build therefore fails with: “The file `global.json` is a tooling file reserved for automated processes. Modifying this file is not permitted.” ([failed job](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1538777&view=logs&jobId=912e8062-b0ba-5930-681b-db020daf9df9))

Removing only the new `sdk` block would fix the offline bootstrap failure, but this `test.runner` addition would still keep required validation red. Please use an authorized tooling update or a permitted MTP integration path that does not modify the protected root file.

test/Microsoft.DotNet.Installer.Tests/LinuxInstallerTests.cs
```diff
@@ -141,7 +138,7 @@ public void Dispose()
         }
     }
 
-    [ConditionalTheory(typeof(LinuxInstallerTests), nameof(IncludeRpmTests))]
+    [Theory(Skip = "Condition not met: IncludeRpmTests", SkipUnless = nameof(IncludeRpmTests))]
```
akoeplinger: nit, since we're establishing a pattern here, this should use nameof in the skip string too:

```suggestion
    [Theory(Skip = "Condition not met: " + nameof(IncludeRpmTests), SkipUnless = nameof(IncludeRpmTests))]
```

akoeplinger: we can also try to just not set `Skip`, I think xunit will record the skip reason regardless.

akoeplinger: and alternatively, it would read more naturally if we used `SkipWhen = nameof(ExcludeRpmTests)`

ViktorHofer: @copilot+claude-opus-5 this sounds like a great idea

ViktorHofer: @copilot+claude-opus-5 looks like we don't need the `IncludeRpmTests` property anymore. Let's just have the negative one that is used by the SkipWhen conditions. Also apply this feedback to all touched .cs files.

ViktorHofer:
> we can also try to just not set Skip, I think xunit will record the skip reason regardless.

@akoeplinger turns out the Skip message is actually required otherwise xunit throws at run time.

ViktorHofer:
OK this PR is now officially ready and will turn green :) I expect minor time savings but more importantly a modern UX due to xunit v3 + MTP.

akoeplinger (approved):

PR #8091, opened 2026-08-01
Use toolchain macros for sccsid attr

am11:
JIT build has a mode where it sets `IGNORE_DEFAULT_TARGET_ARCH` and `IGNORE_DEFAULT_TARGET_OS` to prevent defining `TARGET_` macros and building those components on illumos was failing. Therefore, PR switches to toolchain macros.


am11:
cc @jkotas, @akoeplinger

jkotas (approved):

PR #8151, opened 2026-08-05
Add -mt and -nodereuse switches to build scripts

ViktorHofer:
Implements https://github.com/dotnet/arcade/issues/17242.

MSBuild's experimental multi-threaded mode was previously gated behind an internal `MSBUILD_MT_ENABLED` environment variable. This promotes it to a first-class build script switch.

**The switch defaults to off for now**, on CI and locally, so this is a no-op for existing builds. Enabling it by default for local builds is deliberately deferred until the mode has proven out; that will be a one-line change per `tools.ps1`/`tools.sh`.

### Resolution semantics

Identical across the PowerShell and shell variants:

1. An explicitly specified `-mt` / `--mt` value always wins, on CI and locally.
2. Otherwise it's off.

`tools.ps1` / `tools.sh` own the default resolution and emit `-mt`; the build scripts just expose the switch.

`-mt` takes a value, consistent with the existing `-nodeReuse` and `-warnAsError` parameters:

```console
.\build.cmd -mt 1     # local build with multi-threading
./build.sh --mt 1     # same on Unix
```

### Changes

| Area | Change |
| --- | --- |
| `eng/common` and `src/arcade/eng/common` | `tools.ps1`/`tools.sh` resolve `$msbuildMultiThreaded` / `msbuild_multi_threaded` and emit `-mt`; `build.ps1`/`build.sh` expose `-mt <value>` (long form `-msbuildMultiThreaded`) |
| `build.sh` and `eng/build.ps1` | The VMR's own entry points don't go through `eng/common/build.*`, so they expose the switch directly. An explicit value is also flowed down as `/p:DotNetBuildMT=<value>`, so a single `-mt 0` covers the orchestrator *and* every repo build. |
| `fsharp`, `roslyn`, `aspnetcore`, `nuget-client` | These repos don't use `eng/common/build.*`, so their own build scripts gained `-mt` and `-nodeReuse`. `nuget-client`'s `eng/dotnet-build/build.sh` invokes `msbuild` directly and therefore resolves the defaults itself. |
| `repo-projects/Directory.Build.props` | Replaced the `MSBUILD_MT_ENABLED=1` environment variable with the new switch. `DotNetBuildMT` now passes `-mt <value>`, so it can be used to opt *out* as well as in. |

All other repos pick the change up automatically — the `UpdateEngCommonFiles` target copies the VMR's `eng/common` over every `src/<repo>/eng/common` before that repo builds.

### Repos with custom build scripts

`fsharp`, `roslyn`, `aspnetcore` and `nuget-client` don't call `eng/common/build.*` directly, so both switches had to be added to their own scripts. `fsharp` and `roslyn` reject unknown arguments outright, and `aspnetcore`, `fsharp` and `nuget-client` hard-disabled node reuse, so an explicitly passed value was silently dropped. Each script now declares the parameter and only applies its own default when the caller didn't specify one.

`msbuild` and `runtime` do forward unrecognized arguments to `eng/common/build.*`, but only `runtime` did so in a way that preserves argument types. `msbuild`'s wrapper re-invoked Arcade with `pwsh -File`, which passes every argument as a literal string, and a PowerShell `[bool]` parameter can never bind from a string:

```
Cannot process argument transformation on parameter 'msbuildMultiThreaded'.
Cannot convert value "System.String" to type "System.Boolean".
```

It now uses `pwsh -Command`, which re-parses the arguments as PowerShell so `-mt 1` arrives as an integer — the same thing `runtime`'s wrapper achieves with `Invoke-Expression`, and what `msbuild`'s own `build.cmd` already does when it calls this script. The build still runs out-of-proc so that stage 2 doesn't inherit stage 1's state variables.

> Two behaviours `-File` provided for free had to be restored: arguments that aren't simple tokens are single-quoted so they survive the re-parse (`-warnnotaserror NU1901;NU1902;NU1903`, paths containing spaces), while bare tokens such as `1` stay unquoted so they keep their type; and because `-Command` doesn't propagate a script's `exit N`, the generated command ends with `exit $LASTEXITCODE`, pre-seeded to `1` so a failure that prevents the script from running at all isn't reported as success.

Hard-disabling node reuse is worth fixing at the source, since it also defeats MSBuild Server. Filed dotnet/aspnetcore#68251 and dotnet/fsharp#20217; the changes here are a stop-gap.

### `-nodeReuse` gets the same treatment

Also implements https://github.com/dotnet/arcade/issues/17013.

Node reuse was gated the same way, behind an internal `MSBUILD_NODEREUSE_ENABLED` environment variable, with the added twist that `tools.ps1` / `tools.sh` hard-failed any CI build that had node reuse enabled. `-nodeReuse` already existed as a parameter, it just couldn't be turned *on* for a CI build.

It now works the same way as `-mt` in that an explicitly passed value wins, on CI and locally. Unlike `-mt` its default is unchanged: `!ci`, i.e. on for local builds and off on CI. The `MSBUILD_NODEREUSE_ENABLED` environment variable and the CI hard error are gone, and `eng/build.ps1` / `build.sh` flow an explicit value down as `/p:DotNetBuildNodeReuse`, which `repo-projects/Directory.Build.props` turns into `-nodeReuse 1|0` for every repo build.

`msbuild.ps1` / `msbuild.sh` gained the `-mt` switch as well, so the two smaller entry points behave like `build.ps1` / `build.sh`.

> Removing the CI hard error is a deliberate relaxation: node reuse stays off on CI by default, but it can now be opted into.

### Drive-by: `1`/`0` is now accepted by the shell scripts

PowerShell's `[bool]` parameters only bind `$true`/`$false`/`1`/`0` — passing the *string* `true` or `false` throws a `ParameterBindingArgumentTransformationException`. The shell scripts, on the other hand, only understood `true`/`false`. That meant a caller driving both (like `repo-projects/Directory.Build.props`) had to pick a different value per OS.

`tools.sh` now normalizes `1`/`0` to `true`/`false` for the boolean build arguments — `-nodeReuse`, `-warnAsError` and `-msbuildMultiThreaded` — so `1`/`0` is the one form that works everywhere. Normalization happens on the variable rather than at the comparison, so values forwarded straight to MSBuild (`/nr:$node_reuse`, `/p:TreatWarningsAsErrors=$warn_as_error`) stay valid. Empty still means "not specified".

This is why the per-OS `ArcadeTrueBoolBuildArg` / `ArcadeFalseBoolBuildArg` properties could be removed.

> Note that this slightly changes existing behaviour: `--nodereuse 1` previously meant *off*.

### Validation

- All modified `.ps1` files parse cleanly; all modified `.sh` files pass `bash -n`.
- Sandboxed harnesses with a stubbed build tool captured the real command line produced by `tools.sh`, `eng/common/build.sh` and `eng/common/build.ps1`. `-mt` is suppressed by default both locally and on CI, and emitted for `-mt 1` in either case.
- Verified the full `-mt` matrix against the real `eng/common/build.ps1` parameter block and the real `tools.ps1` / `tools.sh` resolution lines: unset locally, unset on CI, `-mt 1`, `-mt 0`, `-ci -mt 1`, `-ci -mt 0` and the long-form `-msbuildMultiThreaded` all resolve identically on both platforms.
- `dotnet msbuild -getProperty:CommonArgs` on a repo project confirms no `-mt` by default, `-mt 1` with `DotNetBuildMT=true`, and `-mt 0` with `DotNetBuildMT=false`.
- `-mt` resolves to a `Boolean` parameter with the `mt` alias on all six PowerShell entry points, and `--mt 0` is accepted by the `roslyn`/`fsharp` shell scripts (which reject unknown arguments).
- Verified `1`/`0`, `true`/`false` and the unset case all resolve identically through `tools.sh`, and that `--mt 1` propagates as `/p:DotNetBuildMT=true`.
- Verified `dotnet msbuild -mt` is accepted by the MSBuild version in this repo (18.9).
- The same harnesses confirm the `-nodeReuse` matrix on both platforms: off by default on CI, on by default locally, `-ci -nodeReuse 1` yields `/nr:true`, and `-nodeReuse 1|0` propagates as `/p:DotNetBuildNodeReuse`. `dotnet msbuild -getProperty:CommonArgs` confirms `-nodeReuse 1|0` is only emitted when `DotNetBuildNodeReuse` is set.
- `MSBUILD_MT_ENABLED` and `MSBUILD_NODEREUSE_ENABLED` no longer appear anywhere in the repository.
- `msbuild`'s wrapper was replayed against the real Arcade parameter block with the exact command line that failed in CI: `-mt 1 -nodeReuse 1` now bind as booleans, `-mt 0` yields `False`, every `/p:` and `/clp:` argument survives (including one containing spaces) along with `NU1901;NU1902;NU1903`, exit codes propagate through both process hops, and a genuine parameter binding failure exits non-zero.

### Note

`-mt` requires MSBuild 18+. Repos built with `-msbuildEngine vs` (for example `aspnetcore` on Windows) fail with `MSB1001` on a VS 17.x install if `-mt` is passed. This isn't hit while the switch defaults to off, but it needs resolving before the default is flipped.

> [!NOTE]

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -274,6 +274,8 @@ while [[ $# > 0 ]]; do
       ;;
     -msbuildmultithreaded|-mt)
       msbuild_multi_threaded=$2
+      # Flow the explicit choice down to the individual repo builds as well.
+      properties+=( "/p:DotNetBuildMT=$2" )
       shift
       ;;
     -preparemachine)
--- a/build.sh
+++ b/build.sh
@@ -114,7 +114,7 @@ build_check=false
 ci=''
 exclude_ci_binary_log=''
 node_reuse=''
-# Empty means "not specified"; tools.sh defaults it to on for local builds and off on CI.
+# Empty means "not specified"; tools.sh leaves it off unless it's explicitly requested.
 msbuild_multi_threaded=''
 prepare_machine=''
 warn_as_error=''
--- a/build.sh
+++ b/build.sh
@@ -54,6 +54,7 @@ usage()
   echo "  --clean-while-building          Cleans each repo after building (reduces disk space usage, short: -cwb)"
   echo "  --excludeCIBinarylog            Don't output binary log (short: -nobl)"
   echo "  --nodeReuse <value>             Sets nodereuse msbuild parameter ('true' or 'false')"
+  echo "  --msbuildMultiThreaded <value>  Sets MSBuild's multi-threaded mode, i.e. the -mt switch ('true' or 'false') (short: --mt)"
   echo "  --prepareMachine                Prepare machine for CI run, clean up processes after build"
   echo "  --projects <value>              Project or solution file to build"
   echo "  --use-mono-runtime              Output uses the mono runtime"
@@ -113,6 +114,8 @@ build_check=false
 ci=''
 exclude_ci_binary_log=''
 node_reuse=''
+# Empty means "not specified"; tools.sh defaults it to on for local builds and off on CI.
+msbuild_multi_threaded=''
 prepare_machine=''
 warn_as_error=''
 targetOS=''
@@ -267,6 +270,10 @@ while [[ $# > 0 ]]; do
       ;;
     -nodereuse)
       node_reuse=$2
+      shift
+      ;;
+    -msbuildmultithreaded|-mt)
+      msbuild_multi_threaded=$2
       shift
       ;;
     -preparemachine)
--- a/build.sh
+++ b/build.sh
@@ -300,15 +300,19 @@ if [[ "$ci" == true ]]; then
   fi
 fi
 
-# tools.sh fills in the default below, so remember whether -mt was passed explicitly.
+# tools.sh fills in the defaults below, so remember whether these were passed explicitly.
 msbuild_multi_threaded_explicit=$msbuild_multi_threaded
+node_reuse_explicit=$node_reuse
 
 source "$scriptroot/eng/common/tools.sh"
 source "$scriptroot/eng/source-build-toolset-init.sh"
 
-# Flow an explicit -mt choice down to the individual repo builds as well.
+# Flow the explicit choices down to the individual repo builds as well.
 if [[ -n "$msbuild_multi_threaded_explicit" ]]; then
   properties+=( "/p:DotNetBuildMT=$msbuild_multi_threaded" )
+fi
+if [[ -n "$node_reuse_explicit" ]]; then
+  properties+=( "/p:DotNetBuildNodeReuse=$node_reuse" )
 fi
 
 # Default properties
--- a/build.sh
+++ b/build.sh
@@ -274,8 +274,6 @@ while [[ $# > 0 ]]; do
       ;;
     -msbuildmultithreaded|-mt)
       msbuild_multi_threaded=$2
-      # Flow the explicit choice down to the individual repo builds as well.
-      properties+=( "/p:DotNetBuildMT=$2" )
       shift
       ;;
     -preparemachine)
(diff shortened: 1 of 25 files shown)


ViktorHofer:
Validation build with -mt and -nodeReuse turned on in CI: https://dev.azure.com/dnceng-public/public/_build/results?buildId=1542440&view=results

JanProvaznik (approved):

Issue #8131, opened 2026-08-04
[Operational Issue]: Source-built source archive signing repeatedly exhausts retries


mthalman:
### Error

Official VMR builds across `release/10.0.1xx`, `release/10.0.3xx`, and `release/10.0.4xx` repeatedly fail during post-build real signing.

The source-only build itself succeeds. The failure occurs in `Sign Source-Built Artifacts` after approximately 3.5–3.75 hours:

```text
Sign.proj(76,5): error: Failed to execute MSBuild on
Round0-Sign.proj with exit code '1'.

1 files had signing errors:
dotnet-source-<version>.tar.gz
```

The approximately 487 MB source archive receives `FailCanRetry` from the signing backend with no error details. MicroBuild resubmits it five times, then reports:

```text
Max resubmission attempts 5 reached
Status: FailCanRetry
Component: Sign
Error: null
```

All other roughly 3,200 inputs complete signing successfully. Equivalent source archives signed successfully earlier on August 3, 2026. The current failure pattern began between 15:55 and 20:46 UTC that day and has reproduced across multiple builds and SDK bands.

### Links

- [release/10.0.4xx](https://dev.azure.com/dnceng/7ea9116e-9fac-403d-b258-b31fcf1bb293/_build/results?buildId=3038627)
- [internal/release/10.0.1xx](https://dev.azure.com/dnceng/internal/_build/results?buildId=3038118&view=results)

### Root Cause

The signing backend consistently returns `FailCanRetry` without diagnostic details for the large `dotnet-source` archive. The underlying service-side cause is unknown.

### Blocking

- [x] This issue is blocking

### Affected Releases

- [x] .NET 10
- [ ] .NET 11

mmitche:
New occurrences since the tracked build:

- https://dev.azure.com/dnceng/internal/_build/results?buildId=3038118 — internal/release/10.0.1xx
- https://dev.azure.com/dnceng/internal/_build/results?buildId=3038213 — release/10.0.1xx
- https://dev.azure.com/dnceng/internal/_build/results?buildId=3038623 — release/10.0.1xx
- https://dev.azure.com/dnceng/internal/_build/results?buildId=3038624 — internal/release/10.0.1xx
- https://dev.azure.com/dnceng/internal/_build/results?buildId=3038626 — release/10.0.3xx
- https://dev.azure.com/dnceng/internal/_build/results?buildId=3038627 — release/10.0.4xx
[...]

PR #8214, opened 2026-08-10
Cloak SignatureCycle2 binary from the VMR

mthalman:
The checked-in SignatureCycle2 test executable should not be synchronized into the VMR. It blocks signing due to the cycle contained in it. See https://github.com/dotnet/dotnet/issues/8131


mmitche (approved):

src/roslyn/src/Compilers/Test/Resources/Core/MetadataTests/Invalid/Signatures/SignatureCycle2.exe
333fred, line 1: Why is the file being deleted in this PR, and not just cloaked?

333fred (dismissed):

src/roslyn/src/Compilers/Test/Resources/Core/MetadataTests/Invalid/Signatures/SignatureCycle2.exe
333fred: And more specifically, what is going to prevent this from being backported to roslyn, breaking our tests, and also breaking the ability to run tests in the VMR?

mthalman: A file being cloaked necessitates it being deleted. Otherwise it still exists in the repo and continues to cause the issue.

The deletion won't be backflowed because it's marked as a cloaked file.

333fred: That does mean that you cannot run the test associated with this file in the VMR though.

mthalman: That's correct. But do you see that `src/Compilers/Test/Resources/Core/MetadataTests/Invalid/Obfuscated*.dll` is also already cloaked so this nothing new and those tests are already not operable in the VMR. Automated repo tests do not run in the VMR, FWIW.

Anyway, this is needed to unblock servicing. This can likely be reverted once the ESRP issue is resolved.

This is a great reason why we should avoid binaries in the repo too.

jjonescz (approved):

PR #8227, opened 2026-08-11
Cloak SignatureCycle2 binary from the VMR

mthalman:
Backport of #8214 

The obfuscated `SignatureCycle2.exe` test resource blocks source tarball signing when synchronized into the VMR.

Exclude the binary from Roslyn synchronization and remove the existing VMR copy. The resource project now includes it only when present, preserving normal upstream Roslyn builds while allowing VMR builds to omit it.

Validation:
- Built `Microsoft.CodeAnalysis.Compiler.Test.Resources.csproj` in Release
- Built `eng/tools/ChangeValidation/ChangeValidation.csproj` in Release
- Confirmed the source mapping parses and the excluded binary is absent from evaluated embedded resources


akoeplinger (approved):

src/source-mappings.json
```diff
@@ -56,7 +56,8 @@
       "name": "roslyn",
       "defaultRemote": "https://github.com/dotnet/roslyn",
       "exclude": [
-        "src/Compilers/Test/Resources/Core/MetadataTests/Invalid/Obfuscated*.dll"
+        "src/Compilers/Test/Resources/Core/MetadataTests/Invalid/Obfuscated*.dll",
+        "src/Compilers/Test/Resources/Core/MetadataTests/Invalid/Signatures/SignatureCycle2.exe"
```
akoeplinger, line 60: looks like the Obfuscated* ones were removed from roslyn in https://github.com/dotnet/roslyn/pull/82520

akoeplinger (approved):

PR #8281, opened 2026-08-13
Build IdentityModel against Cryptography source-build reference package

dsplaisted:
Fixes the standalone Linux source-build-assets failure flowing through dotnet/source-build-assets#1767.

IdentityModel 8.19.2 requires the Microsoft.Bcl.Cryptography 10.0.10 API surface. Instead of adding a previously-source-built package feed, this follows the established source-build-reference-package design from `main`:

- adds generated reference packages for Microsoft.Bcl.Cryptography 10.0.10 and its System.Formats.Asn1 dependency
- builds IdentityModel against the local Microsoft.Bcl.Cryptography 10.0.10 reference package
- removes the Microsoft.Bcl.Cryptography product dependency and the feed-manipulation target
- allows the Cryptography net10 reference assembly to use the bootstrap SDK targeting pack because this release branch only carries source-built targeting packs through net9

This keeps standalone source-build-assets restore offline and does not require `PreviouslySourceBuiltNupkgCacheDir`.

Validation:
- the source-build reference-package stage builds successfully
- both 10.0.10 reference packages are emitted with the expected assets and dependency groups
- IdentityModel restore resolves Microsoft.Bcl.Cryptography 10.0.10 from the local source-build-assets package output for all target frameworks
- the external IdentityModel build proceeds through restore and into compilation; the remaining local WSL failure is an unrelated IDE0055 formatting error in the cloned upstream source

### DIFF
--- a/src/source-build-assets/eng/Version.Details.props
+++ b/src/source-build-assets/eng/Version.Details.props
@@ -6,7 +6,6 @@ This file should be imported by eng/Versions.props
 <Project>
   <PropertyGroup>
     <!-- dotnet-dotnet dependencies -->
-    <MicrosoftBclCryptographyPackageVersion>10.0.12</MicrosoftBclCryptographyPackageVersion>
     <MicrosoftBuildPackageVersion>18.0.11</MicrosoftBuildPackageVersion>
     <MicrosoftBuildUtilitiesCorePackageVersion>18.0.11</MicrosoftBuildUtilitiesCorePackageVersion>
     <MicrosoftDotNetArcadeSdkPackageVersion>10.0.0-beta.26403.103</MicrosoftDotNetArcadeSdkPackageVersion>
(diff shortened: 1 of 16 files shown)


mmitche (changes requested):
This looks wrong to me. Why is the PSB source required at all now that identitymodel has been released?

mmitche:
I'd also fix this (probaby be removing the target, since I don't think it should be necessary any longer?) in the backflow PR.

mthalman:
> This looks wrong to me. Why is the PSB source required at all now that identitymodel has been released?

The SBA repo needs to build IdentityModel which has a dependency on Microsoft.Bcl.Cryptography that can only be resolved by PSB, since SBA builds before runtime. And yes, a fix should be applied directly to https://github.com/dotnet/source-build-assets/pull/1767. I'll work on that.

dsplaisted:
You're right—the previously-source-built feed was the wrong abstraction here. I've replaced it with the source-build-reference-package approach used on main: IdentityModel now compiles against a local Microsoft.Bcl.Cryptography 10.0.10 reference package (plus its System.Formats.Asn1 reference dependency), and the product dependency/feed target are removed. The reference-package build and IdentityModel restore both succeed locally; the downstream standalone Linux leg still needs to verify the flowed change end to end.

dsplaisted:
> You're right—the previously-source-built feed was the wrong abstraction here. I've replaced it with the source-build-reference-package approach used on main: IdentityModel now compiles against a local Microsoft.Bcl.Cryptography 10.0.10 reference package (plus its System.Formats.Asn1 reference dependency), and the product dependency/feed target are removed. The reference-package build and IdentityModel restore both succeed locally; the downstream standalone Linux leg still needs to verify the flowed change end to end.

OK, this was copilot posting for me without my permission...

PR #8331, opened 2026-08-16
[NuGet] Defer plugin/static-state teardown to end of build (mirror of NuGet/NuGet.Client#7630)

### LINKED ISSUE #15044: RestoreTask tears down credential provider plugins before the restore's own async work has drained, breaking builds that use authenticated feeds (NRE at NuGet.targets(201,5), then 401s)
> *Body updated with stronger evidence from a Windows binary log: restores demonstrably share MSBuild worker processes (up to 6 restores per process), and the credential-service reset has a second, "sticky" failure mode described below. The `SemaphoreFullException` variant originally listed here turned out to be a separate defect from the same commit and is tracked in NuGet/Home#15045.*

## Summary

[NuGet/NuGet.Client@6728fd4](https://github.com/NuGet/NuGet.Client/commit/6728fd42a6835a036fc5bd437a3c33f20a40345f) ("Make RestoreTask resettable across builds (NuGet/Home#14958 item 1)", PR NuGet/NuGet.Client#7507) makes `RestoreTask` terminate the process-wide `PluginManager` — and with it any running credential provider plugin — at the end of **every** restore, and nulls the process-wide credential service at the start of every restore.

Build systems that issue more than one restore per MSBuild process rely on the credential provider plugin started by an earlier restore still being alive for the next one; that reuse is the whole point of the plugin idle/keep-alive timers. After this change each restore kills the plugin for the restores that follow it in the same process.

The user-visible result is an error at `NuGet.targets(201,5)` — most often a bare **`NullReferenceException`** — immediately followed by `401 (Unauthorized)` against authenticated feeds, which then **never recovers for the rest of that process**. The reported error masks the real problem, which is that credential acquisition was killed.

This blocked .NET product builds (`dotnet/dotnet`) as soon as the SDK carrying this change was picked up.

## Symptoms

All errors are reported at `NuGet.targets(201,5)`:

```
NuGet.targets(201,5): error : Object reference not set to an instance of an object.
NuGet.targets(201,5): error : Problem starting the plugin '.../CredentialProvider.Microsoft.dll'. A task was canceled.
NuGet.targets(201,5): error : Problem starting the plugin '.../CredentialProvider.Microsoft.dll'. Cannot access a closed pipe.
NuGet.targets(201,5): error : Cannot write to a closed TextWriter. Object name: 'StreamWriter'.
```

The causal ordering is visible in the logs — the credential request dies **first**, then authentication fails:

```
Retrying 'FindPackagesByIdAsync' for source 'https://pkgs.dev.azure.com/.../microsoft.netcore.app/index.json'.
A task was canceled.                                                    <-- credential request killed
NuGet.targets(201,5): error : Object reference not set to an instance of an object.
Retrying 'FindPackagesByIdAsync' for source '...'
Response status code does not indicate success: 401 (Unauthorized).     <-- consequence
```

## Evidence: restores share worker processes, and there are many of them

From a Windows binary log of one `dotnet/dotnet` build of `src/sdk` (SDK `11.0.100-preview.7.26381.103`). MSBuild reports the real worker node for each project in its `… is building "<project>" (<ctx>) on node N` messages:

| MSBuild worker node | `RestoreTask` invocations in that process |
|---|---|
| 1 | 2 — `Tools.proj`, then `sdk.slnx` (the Arcade pair) |
| 2 | 1 |
| 4 | 5 |
| 6 | 6 |
| **8** | **2 — `dotnet-hello.csproj`, then `dotnet-portable.csproj` (the failure)** |

**16 `RestoreTask` invocations in a single build, spread over 5 processes — so up to 6 plugin teardown/restart cycles in one process.** This is not an exotic configuration: `Microsoft.DotNet.Arcade.Sdk/toolset/Build.proj` issues two restores from the same target (`Tools.proj`, then the solution), and repos add more — here the SDK restores each test asset through a separate `MSBuild` task invocation.

> Note for anyone reproducing this analysis: the `NNN>` prefix in a replayed `-v:diag` log is a **project context id, not a node id** (a 231-project build produces ids 0–230). The real node only appears in the `on node N` messages.

The failing sequence, entirely within **node 8**:

```
12:47:49.888  RestoreTask  dotnet-hello.csproj      -> finally: RaiseEndMSBuildRestoreTasks
                                                    -> PluginManager.ResetSharedInstance() -> plugin process killed
12:48:00.587  RefreshNuGetStaticStateTask           -> RaiseStartMSBuildRestoreTasks
                                                    -> HttpHandlerResourceV3.CredentialService = null
12:48:00.774  RestoreTask  dotnet-portable.csproj   -> must cold-start a NEW plugin
              [CredentialProvider]Sending response: 'GetAuthenticationCredentials'. Time elapsed in ms: 31
              Installed: 6 package(s) to ...\dotnet-portable.csproj
              Retrying 'FindPackagesByIdAsync' for source '...' / A task was canceled.
12:48:02.441  NuGet.targets(201,5): error : Object reference not set to an instance of an object.   (x10)
              Response status code does not indicate success: 401 (Unauthorized).                   (many)
```

Note the replacement plugin answers one request successfully in 31 ms and six packages install — then a request is cancelled and **every** subsequent request to the authenticated feeds is a 401 with no credential attempt at all.

### 1. Teardown is unconditional, but `RestoreTask`'s `finally` cannot know whether the build will restore again

```csharp
static PluginManager()
{
    // The plugin processes must not outlive the build that started them; tear them down at end of restore.
    StaticState.EndMSBuildRestoreTasks += ResetSharedInstance;
}
```

`ResetSharedInstance()` swaps in a fresh `Lazy<IPluginManager>` and disposes the previous one, "terminating any running plugin processes and their idle/keep-alive timers". It is raised from `RestoreTask.Execute`'s `finally`, so it fires after *each* restore, not after the *last* one — and as the table above shows, five more restores may follow in the same process.

`ResetSharedInstance` also installs the replacement `Lazy` **before** disposing the old instance, so a following restore's plugin startup can overlap the previous instance's disposal. Disposing a plugin cancels its outstanding requests (`plugin.Dispose()` → `Connection.Dispose()` → `MessageDispatcher` close → `OutboundRequestContext.Close()` → `TrySetCanceled()`), which is the only code path that produces an *immediate* cancellation — the observed `A task was canceled.` is not a timeout, since `ProtocolConstants.HandshakeTimeout` and `RequestTimeout` are both 30 s while the whole failing restore ran under 2 s.

### 2. The credential-service reset leaves a *sticky* failure

```csharp
static DefaultCredentialServiceUtility()
{
    StaticState.StartMSBuildRestoreTasks += ResetCredentialService;
}

internal static void ResetCredentialService()
{
    HttpHandlerResourceV3.CredentialService = null;
    DelegatingLogger = null;
}
```

Every consumer reads that static defensively — `HttpHandlerResourceV3.CredentialService?.Value` — and therefore silently passes **`null`**:

```csharp
messageHandler = new HttpSourceAuthenticationHandler(packageSource, clientHandler,
                                                     HttpHandlerResourceV3.CredentialService?.Value)
```

`HttpSourceAuthenticationHandler` then does:

```csharp
if (_credentialService == null)
{
    return response;      // 401 returned without ever attempting to acquire credentials
}
```

Because `HttpSource` and its handlers are **cached per source** (`HttpSourceResourceProvider._cache.GetOrAdd(...)`), a handler constructed inside the reset window keeps `_credentialService == null` for the remaining lifetime of the process. That explains why the 401s never recover and why retry-based triage does not help.

`SetupDefaultCredentialService` has the same shape of hazard:

```csharp
var providers = new AsyncLazy<IEnumerable<ICredentialProvider>>(
    async () => await GetCredentialProvidersAsync(DelegatingLogger));
```

the lambda reads the static `DelegatingLogger` at *deferred* execution time, and `ResetCredentialService` nulls it — the same "read the shared static twice" pattern behind NuGet/Home#15045.

### 3. Some restores raise neither event

`Microsoft.Build.NuGetSdkResolver` performs a real, authenticated restore during **evaluation**, against a generated project under `…\Temp\NuGetScratch\<guid>\<guid>.proj` — four of them in this build. These use `PluginManager.Instance` but raise **neither** `StartMSBuildRestoreTasks` **nor** `EndMSBuildRestoreTasks`, so their plugin can be disposed by an unrelated later `RestoreTask`'s `finally` that they never paired with, and they get no compensating credential-service reset.

More generally the two events are asymmetric — eleven resets are attached to `Start` (credential service, proxy cache, throttles, environment caches, `PluginLogger`, …), exactly one — plugin teardown — to `End`:

| event | raised by | when |
|---|---|---|
| `StartMSBuildRestoreTasks` | `RefreshNuGetStaticStateTask` | only via `Restore` → `_GenerateRestoreGraph` → `_RefreshNuGetStaticState` |
| `EndMSBuildRestoreTasks` | `RestoreTask.Execute`'s `finally` | unconditionally, on every `RestoreTask` |

### Contributing lifetime hazards in the same commit

- **`PluginFactory.Dispose()` skips in-flight creations.** It only disposes plugins whose creation task is already `RanToCompletion`, never clears `_plugins`, and never waits for process exit. A plugin still handshaking is left neither disposed nor reachable.
- **`PluginLogger` contradicts `PluginFactory`.** `PluginFactory.Dispose` deliberately stops disposing the shared logger, justified in a comment on the premise that `PluginLogger.DefaultInstance` is never recreated — but the same commit adds `StartMSBuildRestoreTasks += PluginLogger.ResetDefaultInstance`, which recreates *and disposes* it every restore. Inert unless `NUGET_PLUGIN_ENABLE_LOG=true`, and the likely source of the `Cannot write to a closed TextWriter` variant.
- **`HttpSourceResourceProvider.ResetThrottle` has no effect.** `HttpSource` captures its throttle at construction and is cached, so nulling the static cannot affect already-cached sources.
- `PluginProcess.CancelRead` gained a `catch (InvalidOperationException)` for "the plugin process has already exited" — suggesting this class of problem was anticipated but not fully covered.

## Impact

- Any authenticated feed becomes intermittently unusable in builds that restore more than once per process — which, via Arcade, is every .NET repo. Credential acquisition fails, restore falls back to anonymous, and reports `401 Unauthorized`.
- The failure is **sticky within a process** once a handler is cached without a credential service, so retries inside the same build do not recover.
- The surfaced diagnostic is frequently a bare `NullReferenceException` with no stack and no mention of plugins or authentication, making this very expensive to diagnose. Ours initially looked like an auth/infrastructure outage.
- It is nondeterministic — different legs fail on each run — so it also defeats retry-based triage.

Measured in `dotnet/dotnet` official builds, before and after the SDK bump that introduced the change (`11.0.100-preview.6.26359.118` → `11.0.100-preview.7.26381.103`):

| Builds | Legs logging this error |
|---|---|
| 6 consecutive builds before | **0** |
| 3 consecutive builds after | **every build, 2–6 legs each** |

(Some legs recover on retry, so the count of legs *logging* the error exceeds the count that ultimately fail.) No other change in that window is a plausible cause; the only other candidate was scoped to a repository that never exhibits the failure.

## Suggested fix

The intent (don't leak plugin processes into a reused MSBuild process) is right; the mechanism needs to preserve plugin reuse *within* a build. Options:

1. **Don't tear down on every `RestoreTask`.** Tear down once when the build is ending, or reference-count/defer, so plugins survive across the restores of a single build as they did before. `RestoreTask`'s `finally` has no way to know whether another restore is coming.
2. **Raise `EndMSBuildRestoreTasks` from the same scope that raises `StartMSBuildRestoreTasks`** (a target-level task, not `RestoreTask`'s `finally`), so the two are guaranteed to pair — and so restores that raise neither, such as the SDK resolver's, are not silently affected by another restore's teardown.
3. **Don't leave `HttpHandlerResourceV3.CredentialService` observably null.** Swap in the replacement atomically instead of nulling first, so no cached `HttpSource`/handler can capture a null credential service.
4. If teardown must stay where it is, at minimum **make disposal and re-creation mutually exclusive**, and have `PluginFactory.Dispose` account for in-flight creations rather than skipping them.
5. Regardless of the above, **plugin-teardown failures should not surface as `NullReferenceException`**. Disposal races should be swallowed or reported as a diagnosable authentication/plugin error naming the plugin.

JanProvaznik:
> [!IMPORTANT]
> **Do not merge without explicit agreement from the NuGet team.** Review belongs on the authoritative PR: **[NuGet/NuGet.Client#7630](https://github.com/NuGet/NuGet.Client/pull/7630)**. This mirror exists only to shorten the path into the product build if that is desirable — if normal codeflow is preferred, close it.

Mirrors NuGet/NuGet.Client#7630 into `src/nuget-client`. No other directory is touched.

## What it fixes

[#7507](https://github.com/NuGet/NuGet.Client/pull/7507) wired NuGet's process-wide static-state resets to **restore** boundaries, but every requirement they serve is scoped to a **build**. `RestoreTask`'s `finally` therefore tears down the credential-provider plugin while that restore's async work is still draining, and while later restores in the same build still need it.

Arcade issues two `Targets="Restore"` invocations from a single target, so both run in one MSBuild process — which is why this blocked `dotnet/dotnet` official builds:

```
.dotnet/sdk/11.0.100-preview.7.26381.103/NuGet.targets(201,5): error :
    Object reference not set to an instance of an object.

.dotnet/sdk/11.0.100-preview.7.26381.103/NuGet.targets(201,5): error :
    Problem starting the plugin 'CredentialProvider.Microsoft.dll'.
    Cannot write to a closed TextWriter. Object name: 'StreamWriter'.
```

The change collapses `StartMSBuildRestoreTasks` + `EndMSBuildRestoreTasks` into a single `BuildEnded` event raised once per build per node.

Fixes [NuGet/Home#15044](https://github.com/NuGet/Home/issues/15044), [NuGet/Home#15045](https://github.com/NuGet/Home/issues/15045), [NuGet/Home#15046](https://github.com/NuGet/Home/issues/15046).

## Validation in this repo

`dotnet-unified-build` (unsigned, dev branches), all 73 jobs:

| Build | Branch content | SDK doing the build's own restores | Result |
|---|---|---|---|
| [3048183](https://dev.azure.com/dnceng/internal/_build/results?buildId=3048183) | this change only, on main | `preview.7` — still has the regression | ✅ green after 2 selective retry rounds (12 → 6 → 3 legs lost, every one this bug) |
| [3048363](https://dev.azure.com/dnceng/internal/_build/results?buildId=3048363), [3048693](https://dev.azure.com/dnceng/internal/_build/results?buildId=3048693), [3048697](https://dev.azure.com/dnceng/internal/_build/results?buildId=3048697) | + built NuGet grafted onto the bootstrapped SDK | contains the fix | ✅ green, **first attempt** each |

[3049341](https://dev.azure.com/dnceng/internal/_build/results?buildId=3049341) re-bootstraps the VMR on the SDK that 3048183 produced: **53/53 vertical legs green, zero retries.** The one failing leg, `ValidateInstallers_Linux_arm64`, runs from a job template the re-bootstrap step does not cover, so it bootstrapped the unfixed `preview.7` SDK and hit the bug — within a single build, every leg on the fixed SDK passed and the only leg on the old SDK did not.

Full validation detail, including the standalone reproduction and the unit tests, is on [NuGet/NuGet.Client#7630](https://github.com/NuGet/NuGet.Client/pull/7630).


JanProvaznik:
unnecessary, this will flow normally

PR #8346, opened 2026-08-17
Avoid code coverage artifacts overwriting publish files

dsplaisted:
Fix issue that surfaced in dotnet/sdk#55754.  During publish, `Microsoft.CodeCoverage.targets` replaced application assemblies—especially `System.Text.Json` 10.0.0.11 and `System.Memory` 4.0.5.0—with older CodeCoverage-private copies (6.0.0.0 and 4.0.1.1).  This caused assembly loading failures in the net472 tests.  For example:

> System.TypeInitializationException : The type initializer for 'Microsoft.NET.Sdk.Publish.Tasks.HttpClientExtensions' threw an exception.
> ---- System.IO.FileLoadException : Could not load file or assembly 'System.Text.Json, Version=10.0.0.11, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)

It looks like previously a subsequent publish copy with _CopyResolvedFilesToPublishPreserveNewest would overwrite the older files from the code coverage package, but now the code coverage package ended up with newer timestamps on the older assemblies, leading to this issue.

## Description

`Microsoft.CodeCoverage.targets` copied the entire collector tree after `ComputeFilesToPublish`, allowing collector dependencies such as `System.Memory.dll` to overwrite application-owned publish files. This change assigns collector artifacts their publish-relative paths and removes any path already owned by `ResolvedFileToPublish` using path-like metadata matching, while continuing to publish unique collector artifacts.

A focused regression test executes the packaged target with an intentionally newer colliding collector file. It verifies that the application-owned file wins regardless of timestamps and that a unique nested coverage artifact is still copied.


akoeplinger:
> It looks like previously a subsequent publish copy with _CopyResolvedFilesToPublishPreserveNewest would overwrite the older files from the code coverage package, but now the code coverage package ended up with newer timestamps on the older assemblies, leading to this issue.

@omajid this sounds quite similar to https://github.com/NuGet/Home/issues/15050 which was triggered by the deterministic pack timestamp changes. I wonder if we should make that part opt-in instead.

omajid:
Yes, we seem to be running into too many unexpected regressions.

akoeplinger:
@omajid are you going to open a nuget-client PR?

dsplaisted:
@dotnet/dotnet-testing-admin Thoughts on this?  It looks like the issue was exposed by deterministic timestamps in NuGet packages, but there probably needs to be a fix in how the code coverage targets copy dependencies.

nohwnd:
@fhnaseer please have a look on this code coverage issue.

fhnaseer:
@nohwnd Can you make the changes in this pr directly to vstest?

omajid:
> @omajid are you going to open a nuget-client PR?

Sorry, I was swamped with other things.

PR #8542, opened 2026-08-21
Add net12 known framework references

marcpopMSFT:
Port of https://github.com/dotnet/sdk/pull/55880 to accelerate VMR flow while review and testing continue in the SDK repository.

## Summary

- Add the complete live `net12.0` known framework and runtime-pack metadata set.
- Freeze `net11.0` package versions and default runtime versions at the public preview 7 build, including version-sensitive RID lists.
- Add explicit net12 RID inheritance boundaries so future runtime flows do not change net11 metadata.
- Update the annual target-framework rollover guide with the package-version and RID invariants learned from net11.

The temporary net11 pins use `11.0.0-preview.7.26381.103` and should move to `11.0.$(VersionFeature110)` after .NET 11 GA.

Tracking: https://github.com/dotnet/sdk/issues/55851

## Validation

Not run locally; this is an exact two-commit port of the SDK PR for VMR validation.


src/sdk/src/Layout/redist/targets/GenerateBundledVersions.targets
```diff
@@ -779,6 +810,123 @@ Copyright (c) .NET Foundation. All rights reserved.
                               RuntimePackExcludedRuntimeIdentifiers="android%3Blinux-bionic"
                               />
 
+    <!-- .NET 11.0 -->
```
joeloff, line 841: I still wonder if we can simplify this using batching, just for the 1.0 through N-1 versions.

joeloff (approved):

src/sdk/src/Layout/redist/targets/GenerateBundledVersions.targets
marcpopMSFT: @joeloff what did you have in mind? The versions and supported rid list are all subtly different between versions so it's been hard to imagine a simplification that would be easy to do here.

joeloff: With MSBuild batching, if we had an item group that contained net12, net11, etc. we could loop over it to generate everything, instead of copying each time.

ViktorHofer (commented):
null

ViktorHofer (changes requested):
The changes in `repo-projects/Directory.Build.targets` look wrong to me. We shouldn't pass these additional global properties in to every repository. That pollutes the inputs of all inner repos while only sdk needs them.

Also, we shouldn't pass underscore properties in as global properties. I wouldn't pass any of those in and instead condition the logic in `GenerateBundledVersions.targets` logic on the existing global properties, i.e. `KnownFrameworkReferenceMicrosoftNETCoreAppCurrentTargetingPackVersion` and introduce the two additional global properties for the aspnetcore targeting and runtime pack.

Btw, all these properties enable the following infrastructure in arcade to make SB target the latest versions automatically: https://github.com/dotnet/arcade/blob/6e8b73df4173e497174229e0acbda23b30f7ed96/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets#L139

So the SDK might be able to not do anything special at all and just set `EnableKnownPackVersionUpdatesForCurrentTargetFramework=true` in redist.csproj.

marcpopMSFT:
🤖 Reworked this along the proposed path in 8642c1c7eda:

- Restored `repo-projects/Directory.Build.targets` to its original public known-pack property flow; no underscore-prefixed properties or ASP.NET-specific properties are passed to inner repos.
- Explicitly enabled current-framework known-pack updates for `redist.csproj`.
- `GenerateBundledVersions.targets` now uses the existing public NETCore current-framework override as the signal to replace frozen net11 values. ASP.NET versions come from the SDK's existing live `MicrosoftAspNetCoreApp*PackageVersion` properties, so additional VMR globals are unnecessary.
- The source-build `ProductMonikerRid` is added at the net11 boundary before net12 inherits the RID lists. This is still required because the generated bundled metadata is consumed by scenario-test projects after the SDK build; Arcade's in-memory item updates during redist evaluation do not rewrite that generated metadata.

This preserves the preview-7 fallback when no current-framework override is supplied.

marcpopMSFT:
🤖 Reworked the source-build fix in `32d3abc3f13` to use the SDK-local package properties directly:

- Product builds keep the frozen `11.0.0-preview.7.26381.103` values.
- `DotNetBuildSourceOnly=true` overrides the net11 NETCore and ASP.NET values with `MicrosoftNETCoreApp*PackageVersion` and `MicrosoftAspNetCoreApp*PackageVersion`.
- There is no dependency on `EnableKnownPackVersionUpdatesForCurrentTargetFramework` or the VMR-flowed `KnownFrameworkReference*` properties.
- The existing `ProductMonikerRid` additions for apphost, runtime, ASP.NET runtime, Crossgen2, and ILCompiler now occur at the net11 item boundaries, so net11 and net12 inherit them. Crossgen2/ILCompiler portable RID snapshots remain free of the non-portable source-build RID.
- Added TODOs to remove the temporary source-build version and net11 RID overrides when stable .NET 11 targeting packs are available in SBRP.

`repo-projects/Directory.Build.targets` remains restored to its pre-workaround state.

marcpopMSFT:
🤖 Applied Viktor's latest feedback in commit 893fae4074a. The existing latest-TFM ProductMonikerRid additions are restored unchanged, while the temporary .NET 11-specific source-build RID additions remain in place. This intentionally allows the source-built RID to appear twice in the latest TFM lists until the temporary .NET 11 workaround can be removed.

src/scenario-tests/src/Directory.Build.targets
```diff
@@ -28,6 +28,7 @@
 
     <ItemGroup>
       <TestEnvironmentVariable Include="RestoreConfigFile=$(TestRestoreConfigFile)" Condition="'$(TestRestoreConfigFile)' != ''" />
+      <TestEnvironmentVariable Include="RuntimeFrameworkVersion=$(TestFxVersion)" Condition="'$(TestFxVersion)' != ''" />
```
ViktorHofer, line 31: @marcpopMSFT what is this change about? I don't understand it.

ViktorHofer (approved):

src/scenario-tests/src/Directory.Build.targets
marcpopMSFT: Sent offline as well but that was the copilot fix for the scenario tests that were failing in the ubuntu and windows legs. By pinning the 11 version to p7, p7 needs to be installed for the app to run and we dont' have that in the env. Not sure if there's a better way to fix it.
[Pipelines - Run 20260821.35 logs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1563192&view=logs&j=879e79ff-d5e4-564a-9322-bee9d0a651ca&t=a8ad6d6f-3a63-551a-2266-a41dc5f07c4b)

Since you approved, I'm going to mark resolved and merge so I can test the resulting build and we can fix up if there's an issue.

PR #8546, opened 2026-08-21
Add Windows bootstrap validation pipeline

### LINKED ISSUE #1071: Add a windows stage2 pipeline that runs on schedule and on demand (comment trigger)
Motivation: We want to make sure that the we can build the VMR with a just live-built VMR (called stage2 build) so that we can detect issues earlier, before we do a SDK toolset re-bootstrap.

Building windows stage1 and stage2 combined takes close to 4 hours which is too long to enable in PRs. We should add a pipeline that runs on schedule (i.e. once per day) and on demand (comment trigger) that has two jobs:
- win-x64
- win-x64 BuildPass="stage2"

The second job should depend on the previous job and build with the produced SDK and packages.

mmitche:
Fixes #1071.\n\nAdds a daily and comment-triggered public pipeline that builds win-x64 twice. The second job depends on the first and bootstraps from its SDK and Arcade packages.

### DIFF
--- a/eng/msft-toolset-init.ps1
+++ b/eng/msft-toolset-init.ps1
@@ -105,7 +105,7 @@ function Initialize-MsftToolset([string]$repoRoot, [string]$customSdkDir, [strin
 
     Write-Host "Using bootstrap Arcade SDK version '$arcadeVersion'"
     Update-GlobalJsonMSBuildSdkVersion (Join-Path $repoRoot 'global.json') 'Microsoft.DotNet.Arcade.Sdk' $arcadeVersion
-    $extraProperties += "/p:ExtraRestoreSourcePath=$customPackagesDir"
+    $extraProperties += "/p:CustomPreviouslySourceBuiltPackagesPath=$customPackagesDir"
     $extraProperties += "/p:RestoreAdditionalProjectSources=$customPackagesDir"
   }
 
--- a/eng/msft-toolset-init.ps1
+++ b/eng/msft-toolset-init.ps1
@@ -1,0 +1,113 @@
+### Initializes a Microsoft build from an externally supplied SDK and package set.
+### This must be loaded before eng/common/tools.ps1 so the latter sees the updated global.json.
+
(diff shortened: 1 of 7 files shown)


eng/pipelines/templates/steps/bootstrap-toolset.yml
```diff
@@ -0,0 +1,46 @@
...
+  inputs:
+    artifactName: ${{ parameters.artifactName }}
+    itemPattern: assets/*/Sdk/**/dotnet-sdk-*-win-x64.zip
+    targetPath: $(Agent.TempDirectory)/bootstrap-sdk-download
+
+- task: DownloadPipelineArtifact@2
+  displayName: Download Bootstrap Packages
+  inputs:
+    artifactName: ${{ parameters.artifactName }}
+    itemPattern: packages/*/NonShipping/arcade/**
```
ViktorHofer, line 17: Don't we need the other shipping packages as well that accompany an SDK release? I.e. the ILLink packages?

eng/msft-toolset-init.ps1
```diff
@@ -0,0 +1,113 @@
...
+    Add-NuGetSourceToConfig (Join-Path $repoRoot 'NuGet.config') 'bootstrap-packages' $customPackagesDir
+
+    $arcadeVersion = Find-PackageVersion $customPackagesDir 'Microsoft.DotNet.Arcade.Sdk'
+    if (-not $arcadeVersion) {
+      throw "Microsoft.DotNet.Arcade.Sdk package not found under '$customPackagesDir'"
+    }
+
+    Write-Host "Using bootstrap Arcade SDK version '$arcadeVersion'"
+    Update-GlobalJsonMSBuildSdkVersion (Join-Path $repoRoot 'global.json') 'Microsoft.DotNet.Arcade.Sdk' $arcadeVersion
+    $extraProperties += "/p:CustomPreviouslySourceBuiltPackagesPath=$customPackagesDir"
```
mthalman, line 108: I wouldn't think this needs to be set given this is not a source-build scenario.

ViktorHofer:
Additional question, could we run this directly as part of the validation stage inside official builds?

PR #8567, opened 2026-08-22
Revert "Apply NuGet.Client close-on-exec fix"

agocke:
Reverts dotnet/dotnet#8490

this fix was incorrect because open is a variadic function and you can’t pinvoke it directly. The proper fix is being tested in nuget now


akoeplinger (approved):

akoeplinger:
Thanks. It looks like this actually caused test failures on osx-arm64.

PR #8599, opened 2026-08-24
Stop bootstrap Arcade repos from consuming shared components

mthalman:
## Problem

Bootstrap Arcade repositories such as `arcade` and `source-build-assets` should resolve their build dependencies from previously source-built (PSB) packages. However, the VMR orchestration also supplied them with current 1xx shared-components inputs through package-version properties, NuGet source mappings, and `AdditionalSourceBuiltNupkgCacheDir`.

When PSB and shared components are at different servicing levels, the shared-components package versions can override PSB versions and create a torn bootstrap state. Bootstrap repositories can also restore incidental dependencies from the shared-components feed even though they should not depend on those components.

## Solution

Use `BootstrapArcadeRepos` to exclude every bootstrap Arcade repository from shared-components inputs while preserving the existing behavior for all other repositories. Bootstrap repositories now receive package versions and restore inputs from PSB only; shared-components version properties, NuGet mappings, and the additional package cache are omitted from their builds.

### DIFF
--- a/repo-projects/Directory.Build.props
+++ b/repo-projects/Directory.Build.props
@@ -6,6 +6,8 @@
   <PropertyGroup>
     <RepositoryName>$(MSBuildProjectName)</RepositoryName>
     <UseBootstrapArcade Condition="$([System.String]::new(';$(BootstrapArcadeRepos);').Contains(';$(RepositoryName);'))">true</UseBootstrapArcade>
+    <!-- SBA must bootstrap from PSB rather than current 1xx shared components. -->
+    <_UseSharedComponentsInputs Condition="'$(RepositoryName)' != 'source-build-assets'">true</_UseSharedComponentsInputs>
 
     <BuildInParallel Condition="'$(BuildInParallel)' == ''">true</BuildInParallel>
     <RunEachTargetSeparately>false</RunEachTargetSeparately>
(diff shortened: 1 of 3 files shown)


ViktorHofer:
> source-build-assets is a bootstrap repository whose build dependencies should come from previously source-built (PSB) packages.

Isn't Arcade in the same bucket?

mthalman:
> > source-build-assets is a bootstrap repository whose build dependencies should come from previously source-built (PSB) packages.
> 
> Isn't Arcade in the same bucket?

Fixed in https://github.com/dotnet/dotnet/pull/8599/commits/4b0a884487bc947e82ed8e7b3b4e12b118382fe2

repo-projects/Directory.Build.props
```diff
@@ -6,6 +6,8 @@
   <PropertyGroup>
     <RepositoryName>$(MSBuildProjectName)</RepositoryName>
     <UseBootstrapArcade Condition="$([System.String]::new(';$(BootstrapArcadeRepos);').Contains(';$(RepositoryName);'))">true</UseBootstrapArcade>
+    <!-- Bootstrap Arcade repos must bootstrap from PSB rather than current 1xx shared components. -->
+    <_UseSharedComponentsInputs Condition="!$([System.String]::new(';$(BootstrapArcadeRepos);').Contains(';$(RepositoryName);'))">true</_UseSharedComponentsInputs>
```
ViktorHofer: Why compute the value again?

mthalman: Fixed in c9065a5

ViktorHofer (approved):

akoeplinger:
should we backport this?

mthalman:
> should we backport this?

I don't think there's a need to. There's already a different solution that exists in 10.0.4xx that's not quite as elegant but it still works.

Issue #7677, opened 2026-07-11
CI leg for community platform to avoid regressions


am11:
Community platforms use cross-building through officially supported platforms to perform VMR end-to-end builds—e.g., freebsd-x64 cross-built on linux-x64. The most significant difference is that the dotnet/runtime build uses the `--bootstrap` flag for these platforms. This builds host tools in a first pass, then uses those live-built tools to fulfill requirements that typically come from nuget.org (see https://github.com/dotnet/dotnet/blob/0c4423f49f/src/runtime/docs/workflow/building/coreclr/cross-building.md#building-coreclr-with-bootstrapping). 

This type of platform is not covered by the CI and runs into regressions from time to time: https://github.com/dotnet/dotnet/pull/398#issuecomment-4927043227, https://github.com/dotnet/dotnet/issues/7596 etc. Once this "stage 0" SDK is built, we can perform a native build on the host machine using that stage 0 SDK. This is how the FreeBSD folks have been shipping for several years (cc @Thefrank).

The platforms which fall under this category are:

* freebsd-x64
* freebsd-arm64
* linux-loongarch64
* linux-riscv64
* illumos-x64
* solaris-x64
* haiku-x64
* openbsd-x64

Other platforms joining them in the near future are linux-s390x and linux-ppc64le (currently they are mono-only and coreclr work is in progress).

So we can see it is roughly a 50-50 split when it comes to officially-supported vs. community-supported platforms.

This is the generic build command for all of these platforms:
https://github.com/filipnavara/dotnet-riscv/blob/8abd76591d67cc23e77c6e01bf3a9b3cac207e90/.github/workflows/build.yml#L77-L80

It would be good to have one CI leg to exercise this use case using the same command, which takes 60-80 mins (well within the range of other CI builds). This is to prevent regressions earlier in the cycle. I recommend using freebsd-arm64 because:

1. The FreeBSD OS is the oldest and most mature in this category (the port started in 2015, though its arm64 architecture was added later).
2. It covers a "different OS" and "different arch" from the host OS (linux-x64); several conditions have had to be tweaked for them in the past.

MCR's docker tag for this one is `azurelinux-3.0-net11.0-cross-freebsd-14-arm64` (soon it will be in azurelinux-4.0 as well: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1674).

jkotas:
[...]

PR #8618, opened 2026-08-24
Integrate runtime bootstrap into the main build

mmitche:
## Summary

Reduce runtime build time by replacing the separate bootstrap build and broad `artifacts/bin` / `artifacts/obj` deletion with shared producers in one MSBuild invocation.

- `-bootstrap` enables the integrated graph with `EnableBootstrap=true`.
- Normal runtime, library, host, and compiler prerequisites build once. Their completed outputs form a copied, input-hashed snapshot at `artifacts/bootstrap/<rid>/`.
- The remaining product build and isolated shipping-tool publication run after that boundary; dependent packages are built after both finish.
- Deferred tools, packages, and library tests are selected from the existing subset graph, preserving requested scope and platform conditions.
- Explicit `Subset=bootstrap` and subsequent `--use-bootstrap` builds retain their legacy behavior.
- Shared layout manifests handle configuration changes, added/removed inputs, missing files, and portable/distro RID aliases.

Tracking: [DNCENG #12337](https://dev.azure.com/dnceng/internal/_workitems/edit/12337)

## How bootstrapping worked before

Here, **runtime bootstrapping** means building the shipping runtime tools/components using live runtime inputs from the current source tree. It is not the source-build stage-two operation that rebuilds the entire SDK using a previously built SDK.

Ordinary runtime developer builds can publish tools against last-known-good (LKG) assets from the pinned build SDK. A bootstrap build instead needs the current runtime's reference/runtime packs, native hosts, and, when applicable, NativeAOT SDK files. This matters both for shipping a coherent set of bits and for targets whose packs are not available from the usual feeds.

Before this PR, `eng/build.ps1` and `eng/build.sh` implemented `-bootstrap` / `--bootstrap` as two separate build invocations:

```text
build wrapper
  |
  +-- invocation 1: Subset=bootstrap
  |     build bootstrap producers
  |     copy selected outputs into artifacts/bootstrap/<rid>/
  |
  +-- delete artifacts/bin and artifacts/obj
  |     (Unix preserves obj/source-built-upstream-cache)
  |     keep artifacts/bootstrap/<rid>/
  |
  +-- invocation 2: requested product subsets, UseBootstrap=true
        rebuild product outputs
        publish shipping tools against the retained bootstrap layout
        package/sign/publish the requested product
```

The first invocation selected `host.native+libs.sfx+libs.pretest`, plus either `clr.nativeaotlibs+clr.nativeaotruntime+libs.native` for NativeAOT components or `clr.runtime+clr.corelib` for the non-NativeAOT/single-file path. `SetupBootstrapLayout` copied the reference pack, runtime pack, native hosts, applicable AOT SDK files, and generated RID graphs into the retained layout.

The separate bootstrap subset also selected bootstrap-specific composition settings. In particular, it set `SwapNativeForIL`, but CoreCLR IL replacement actually runs only when `UseNativeAotCoreLib` is not true. The NativeAOT SDK is a separate output from the NativeAOT producers; building only `clr.runtime` is not sufficient.

The cleanup separated the bootstrap and product configurations, but threw away reusable compilation outputs and intermediates. The product invocation then repeated overlapping native/library/host work. Simply removing the cleanup would not establish correct restore ordering, prevent conflicting writers, or preserve the required bootstrap layout.

## Desired state

**Keep live-input bootstrap semantics, but pay for shared product producers once instead of building them, deleting them, and building them again.**

The intended dependency ordering is:

```text
one top-level runtime MSBuild invocation
  |
  +-- shared product producers
  |     native runtime/JITs, libraries, hosts, build prerequisites
  |     and host-runnable ILC where needed
  |
  +-- copy a stable bootstrap input snapshot
  |
  +-------------------------------+-------------------------------+
  | remaining product work        | bootstrap shipping consumers  |
  | normal artifacts directory    | separate bin/obj directory    |
  |                               | restore, then publish tools   |
  +-------------------------------+-------------------------------+
                                  |
                        join before dependent packages
                        and bootstrap-dependent library tests
```

The requirements are producer reuse, correct evaluation/restore ordering, independent output ownership, and preservation of requested subsets and supported platform modes. This is **not** a goal of compiling every project exactly once: isolated consumer dependencies can still be built in their own context, and genuinely different configurations remain different MSBuild project instances.

This also does not eliminate the initial build SDK or the in-build host tools. For example, `ILCompiler_inbuild` is built from current compiler sources but runs using the available host/LKG packs; it then compiles the shipping NativeAOT tools using live framework/AOT inputs. That distinction breaks the bootstrap cycle.

### 1. Make bootstrap a graph choice, not another wrapper invocation

The wrappers now pass `EnableBootstrap=true` instead of invoking the separate bootstrap build and deleting product artifacts. `Build.proj` selects `main-build.proj` and `bootstrap.proj` as parallel traversal references. Without `EnableBootstrap`, it keeps the ordinary project traversal.

`EnableBootstrap` means "orchestrate the integrated build"; `UseBootstrap` means "consume an existing bootstrap layout." Keeping those meanings separate avoids applying consumer settings to all product producers.

### 2. Select prerequisites and deferred work from the existing subset graph

`Subsets.props` adds the necessary bootstrap prerequisites to the requested subsets and marks projects with their roles:

| Metadata | Role |
| --- | --- |
| `BootstrapProducer` | Shared product inputs needed before the snapshot |
| `BootstrapPrerequisite` | Build integration/ILCompiler task outputs also needed in the isolated consumer tree |
| `BootstrapConsumer` | Live-input publication of ILC, Crossgen2, ILAsm, or cDAC, as selected by the subsets |
| `BootstrapPackage` / `BootstrapCdacPackage` | Packages that must wait for published outputs |
| `BootstrapTest` | Library tests whose restore/build needs the bootstrap layout |

The deferred lists are captured **after** configuration metadata has been assigned, and removed from the initial traversal only for integrated builds. This avoids an independent hard-coded tool/package matrix: a host-only build does not suddenly publish SDK tools, and a build without `packs` does not acquire tool packages merely because bootstrap is enabled.

### 3. Share the producer execution and make its completion a real boundary

Both traversal paths request `bootstrap-producers.proj`. It builds the selected producer items in their existing order/metadata, then creates the layout as an `AfterTargets="Build"` step. The explicit before-build hooks in the two traversals establish that boundary before either downstream path proceeds.

Producer references use matching global properties. `BootstrapBuildLane` and `BootstrapBuildArtifactsDir` are stripped when crossing back into shared producers, using the metadata appropriate to both restore and traversal execution. This matters because MSBuild identifies a project by its path **and global properties**, not its filename alone. Different lane properties could otherwise create independent executions writing the same outputs. The measured binlogs show later references reusing completed producer work.

### 4. Snapshot inputs without deleting the product output tree

`bootstrap-layout.targets` collects files after producers finish and writes a manifest containing each source path, destination path, and SHA-256 of the source content. A changed file set/content or a missing destination invalidates the layout. Rebuilding removes only `artifacts/bootstrap/<rid>/`, copies its inputs, and writes `.complete` after copying succeeds.

This is an **input-hashed copied snapshot**, not a content-addressed directory or a permanently immutable cache. It remains stable for its consumers during that build and is refreshed in place by a subsequent bootstrap build when needed. The manifest shares the layout's RID scope so switching Release -> Debug -> Release cannot incorrectly accept another configuration's layout.

The layout also aliases distro and portable RID directories when necessary. It uses generated RID graphs when source-build adds RIDs, and checked-in graphs otherwise, rather than assuming portable builds generate graph files.

### 5. Isolate consumers, but reuse the native inputs and host compiler

`bootstrap.props` centralizes three distinct locations:

| Location | Ownership |
| --- | --- |
| Normal `artifacts/bin` and `artifacts/obj` | Product producers, native/JIT inputs, and host-runnable compiler |
| `artifacts/bootstrap/<rid>/` | Copied framework/AOT/host/RID inputs consumed by bootstrap publication |
| `artifacts/bootstrap-build/<rid>/<configuration>/` | Isolated consumer intermediates and published outputs |

The isolated consumers still reference the normal product CoreCLR inputs; they do not rebuild the entire native producer tree under a second artifacts root. Relevant copies avoid hardlinks so isolation does not inadvertently share writable files. The native script changes consistently honor `ArtifactsDir` overrides across MSBuild and the invoked scripts.

`bootstrap.proj` restores/builds the isolated prerequisite projects, then restores the selected consumers, then invokes their **default targets**. The timing is important: generated NativeAOT imports and live pack metadata must exist before consumer restore/evaluation. Default targets are important because the publish projects use `Publish;PublishItemsOutputGroup`, not merely `Build`.

### 6. Rejoin for packaging and preserve the existing entry points

The root `BuildBootstrapPackages` target waits for both traversal paths. Crossgen2 packaging references the isolated publish context using the same consumer properties; its smoke test still uses the main product's IL CoreLib. NativeAOT packaging reads the appropriate compiler/build-integration outputs. The cDAC package deliberately combines the isolated published reader with the normal product's DBI rather than rebuilding the native debug component just for packaging.

Library-test restore/build is similarly deferred and receives `UseBootstrap=true`; `Test` and `VSTest` are forwarded. Explicit `Subset=bootstrap` still imports the layout targets in the ordinary traversal, retains its legacy producer/composition settings, and does not publish shipping consumers. The canonical layout path remains usable by subsequent `--use-bootstrap` builds.

### Reviewer focus

The important invariants are: matching producer project identities; no consumer restore before its prerequisites/layout exist; no unintended writes into the other traversal's outputs; complete package/test deferral with subset/configuration metadata preserved; and correct snapshot invalidation. The inline review comments walk through these points at the relevant diff hunks. Performance and validation evidence, including the remaining FreeBSD coverage gap, are recorded below.

## Performance

Local Windows x64 Release measurements from fresh output directories, using the same SDK and native compiler:

```text
build.cmd -subset clr+libs+host -configuration Release -arch x64 -bootstrap -bl
```

| Revision | Wall time |
| --- | ---: |
| Current-main baseline `d1920aa3fa8` | 26m22s |
| Integrated build, committed revision `a28bdbe0fca` | 20m00s |
| Improvement in this paired local measurement | **6m22s / 24.2%** |

These are runtime-build measurements, not an end-to-end VMR CI timing claim. Binlogs confirm that shared producer executions are reused by the second traversal rather than rebuilt.

### Full unified-build CI comparison

Before: [run 1605346](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1605346), tested revision `c27c4f20f79e90be8f5d78ac8d98da02b70f74ed`. After: [run 1606307](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1606307), tested merge `cfa37a52e1a5167c099f08257ce237d873c06a0e` with PR head `a28bdbe0fca`. Both ran on September 21, using pipeline 278 revision 31, `buildScope=auto`, with all nine jobs passing on their first attempt.

The before tree differs from the after run's main parent (`baa2aa16da0`) only by three license-exclusion lines: product source (`src/`) and build infrastructure (`eng/`) tree identities match exactly. The Windows x64 jobs used the same image, `windows.vs2026.amd64.open` version `2026.0908.185114`, but different agents. This is a historical before/after observation, not a repeated controlled A/B test; machine load, caches, and external inputs may differ.

**Full job duration** is job start to finish, including checkout/setup, building, validation, and artifact publishing, but excluding time before the job starts:

| Full CI job | Before | After | Observed change |
| --- | ---: | ---: | ---: |
| Windows x64 | 2h01m03s | 1h58m08s | 2m55s faster |
| Windows x86 | 1h23m19s | 1h14m27s | 8m52s faster |
| Ubuntu build/tests x64 | 1h29m06s | 1h20m54s | 8m12s faster |
| CentOS online source-build | 53m08s | 47m57s | 5m11s faster |
| CentOS offline stage-two | 50m53s | 49m02s | 1m51s faster |
| Android arm64 short stack | 36m38s | 36m42s | 0m05s slower |
| Browser WASM short stack | 1h09m54s | 1h09m11s | 0m44s faster |
| iOS Simulator arm64 short stack | 1h31m01s | 1h13m38s | 17m23s faster |
| VMR change validation | 3m03s | 2m58s | 0m06s faster |

**Full VMR Build-step duration** isolates the pipeline's `Build` task from job setup and post-build work:

| Build step | Before | After | Observed change |
| --- | ---: | ---: | ---: |
| Windows x64 | 1h50m18s | 1h46m57s | 3m21s faster |
| Windows x86 | 1h19m21s | 1h10m41s | 8m40s faster |
| Ubuntu build/tests x64 | 1h18m01s | 1h10m24s | 7m38s faster |
| CentOS online source-build | 40m58s | 36m12s | 4m46s faster |
| CentOS offline stage-two | 39m54s | 37m40s | 2m13s faster |

Within Windows x64, the runtime repository build fell from **1h18m24s to 1h11m44s**, a **6m40s / 8.5%** reduction. The complete VMR Build step improved by **3m21s**, and the complete job by **2m55s / 2.4%**. Other repository build spans were longer in the after run; not all runtime savings translated into end-to-end savings. A second nearby baseline, run 1605226, puts the Windows x64 full-job reduction at only 46 seconds, illustrating run-to-run variability.

**Entire pipeline, queue to final completion:** **2h10m12s before -> 2h46m59s after (36m47s longer)**. The iOS job started 34m54s after pipeline start in the baseline, versus 1h29m54s afterward: approximately 55 minutes later. Its own job execution was shorter, but the later start made it the after-run's final finisher. Pipeline start-to-finish was 2h06m55s -> 2h43m44s. These total durations include scheduling delays and should not be presented as build-performance regressions attributable to this PR.

The mobile/WASM short-stack legs do not enable the runtime bootstrap optimization; their timing changes, especially iOS, are not claimed as benefits of this change. Full-job and build-step tables above are elapsed durations, not a sum of parallel repository work. No CI end-to-end 24.2% claim is made; that figure applies only to the paired local runtime measurement above.

## Local validation

- Clean build of the exact committed revision: zero warnings and errors.
- 41 graph/layout regression assertions, including subset selection, legacy ordering, configuration switches, content changes with preserved timestamps, file additions/removals, missing-file recovery, and RID-graph selection.
- NativeAOT and `UseNativeAotForComponents=false` single-file builds completed; published ILC, Crossgen2, and ILAsm launch successfully.
- Explicit legacy bootstrap builds completed in NativeAOT and CoreCLR modes; the CoreCLR-mode snapshot contains byte-identical IL CoreLib.
- A real missing snapshot apphost was restored successfully.
- NativeAOT/Crossgen2 packages and the cDAC reader/DBI package payloads were inspected. Final-commit ILC and Crossgen2 package payload hashes match the live published executables; packaged Crossgen2 also successfully compiled CoreLib.
- An unchanged-source rebuild completed in approximately six minutes. Existing native `_version.h` invalidation still causes relinking and a legitimate snapshot refresh; the same native dirtiness was observed on unchanged main.

## CI validation

[Unified-build run 1606307](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1606307) passed **all nine jobs on their first attempt** for PR head `a28bdbe0fca` (tested merge revision `cfa37a52e1a5167c099f08257ce237d873c06a0e`):

- Windows x64 and Windows x86.
- CentOS Stream 10 online source-build using the Microsoft SDK.
- CentOS Stream 10 offline stage-two build using the current source-built SDK.
- Ubuntu build tests.
- Android arm64, iOS Simulator arm64, and Browser WASM short-stack builds.
- VMR change validation.

## Remaining validation

Dedicated Linux-to-FreeBSD-arm64 validation remains outstanding; the default CI matrix is unchanged (see dotnet/dotnet#7677).

Broader local MSI generation was blocked by `wix.exe` not being found. The identical failure was reproduced on unchanged main; targeted live-tool package generation succeeded independently.

> [!NOTE]
> This pull request description was updated with GitHub Copilot.


mmitche:
@akoeplinger Thoughts here? My thought was that the bootstrapping throws away a bunch of artifacts that are identical between bootstrapping and the main build. By removing the block delete we could probably cut a bunhc of time off the builds.

JulieLeeMSFT:
@mmitche, @dotnet/dnr-codeflow, and @dotnet/roslyn-infrastructure, the build failed due to missing ILCompiler import and Roslyn VSIX target.

jjonescz:
Can you link the failing build? All links from this PR to azdo CI give me "build not found".

am11:
Runtime repo, at some point, switched from live-live builds to use LKG tools which broke community platforms since their packages aren't published on nuget.org. For that we implemented multi-stage build and it took me couple of months to figure out all the quirks and people had to pause the port work.. @jkoritzinsky, reworked "multi-stage build" to "bootstrap" and later used it in VMR for official Microsoft builds as well. In principle, it should be smooth and this PR should be an improvement over that.

In practice, during .NET 11 development, preview 1-4 were fine as far as community platforms are concerned, preview 5 broke because some bug sneaked in, preview 6 was fine, preview 7 broke again (so as of today, we don't have the latest preview). This is because official builds have a "luxury" to fallback to LKG so it papers over such bug and only ones which are affected are community platforms. There are <ins>**nine**</ins> community platforms at the moment, so the pain ripples. We should _at least_ add one community platform (like freebsd-arm64 would be my pick; different-than-host OS as well as arch) in VMR CI (be it in full matrix if PR legs is too much; which it shouldn't as it takes 60-70mins in GH actions..) to avoid such breakages with this type of change.

RE: https://github.com/dotnet/dotnet/issues/7677

src/runtime/eng/build.ps1
```diff
@@ -414,54 +414,7 @@ $arguments += " /clp:ForceNoAlign"
...
-    Write-Error "Bootstrap build failed. Stopping build."
-    exit 1
-  }
-
-  # Remove artifacts from the bootstrap build so the product build is a "clean" build.
-  Write-Host "Cleaning up artifacts from bootstrap build..."
-  Remove-Item -Recurse "$PSScriptRoot/../artifacts/bin"
-  Remove-Item -Recurse "$PSScriptRoot/../artifacts/obj"
-  $arguments += " /p:UseBootstrap=true"
+  $arguments += " /p:EnableBootstrap=true"
```
mmitche, line 417: **1/14: Replace the two-build wrapper, not bootstrap semantics.** Previously this block ran `Subset=bootstrap`, retained `artifacts/bootstrap/<rid>`, deleted product `bin`/`obj`, and launched the requested product build with `UseBootstrap=true`. Now it only enables graph orchestration. The shell wrapper makes the equivalent change. `EnableBootstrap` means "produce and coordinate"; `UseBootstrap` remains "consume the layout." The deleted wrapper work is what made shared compilation outputs unavailable to the second build.

_GitHub Copilot-generated explanatory annotation._

src/runtime/eng/Subsets.props
```diff
@@ -188,6 +190,8 @@
     <_subset>$(_subset.Replace('+host+', '+$(DefaultHostSubsets)+'))</_subset>
     <_subset>$(_subset.Replace('+packs+', '+$(DefaultPacksSubsets)+'))</_subset>
     <_subset>$(_subset.Replace('+bootstrap+', '+bootstrap+$(BootstrapSubsets)+'))</_subset>
+    <_subset Condition="'$(EnableBootstrap)' == 'true' or '$(BootstrapMainBuild)' == 'true'">$(_subset)+$(BootstrapSubsets)+</_subset>
```
mmitche, line 193: **2/14: Add prerequisites without switching the whole product into the legacy bootstrap subset.** Both integrated traversals expand the same requested subsets with the required bootstrap producers. This supplies missing inputs even for a narrow request. NativeAOT needs its libraries/runtime plus the host compiler/JIT prerequisites; the single-file path needs CoreCLR and CoreLib instead. This does not globally force `SwapNativeForIL` onto the normal product graph; explicit `Subset=bootstrap` retains its separate legacy composition behavior.

_GitHub Copilot-generated explanatory annotation._

src/runtime/eng/Subsets.props
```diff
@@ -824,4 +856,15 @@
     </ProjectToBuild>
   </ItemGroup>
 
+  <ItemGroup>
+    <BootstrapConsumer Include="@(ProjectToBuild->WithMetadataValue('BootstrapConsumer', 'true'))" />
```
mmitche, line 860: **3/14: Partition the already-selected projects, preserving their metadata.** These lists are formed after per-category and cross-subset configuration properties have been attached. Integrated builds then remove the selected consumers/packages/tests from the early traversal, so initial restore cannot reach them before the live layout exists. Ordinary builds leave them in place. This is also the scope guard: `host.native` does not imply publishing ILC, and omitting `packs` does not silently add tool packages. Platform conditions remain at each project's original declaration.

_GitHub Copilot-generated explanatory annotation._

src/runtime/eng/bootstrap-producers.proj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.Build.Traversal">
+
+  <Import Project="bootstrap.props" />
+
+  <ItemGroup>
+    <ProjectReference Include="$(RepoTasksDir)tasks.proj"
+                      BuildInParallel="false" />
+    <ProjectReference Include="@(ProjectToBuild->WithMetadataValue('BootstrapProducer', 'true'))" />
```
mmitche, line 8: **4/14: One shared producer boundary feeds both downstream traversals.** This selects the existing producer items rather than launching another complete runtime build. Their ordering and metadata are retained. The imported layout target runs after these producers complete. Both `main-build.proj` and `bootstrap.proj` explicitly request this boundary before proceeding; the main traversal subsequently references the product projects again, but matching MSBuild project identities reuse completed targets instead of recompiling them. The measured binlogs confirm that reuse for the shared producer traversal, native runtime, and shared libraries.

_GitHub Copilot-generated explanatory annotation._

src/runtime/eng/bootstrap.proj
```diff
@@ -0,0 +1,49 @@
+<Project Sdk="Microsoft.Build.Traversal">
+
+  <Import Project="bootstrap.props" />
+
+  <ItemGroup>
+    <ProjectReference Include="$(RepositoryEngineeringDir)bootstrap-producers.proj"
+                      AdditionalProperties="EnableBootstrap=false;BootstrapMainBuild=true"
+                      GlobalPropertiesToRemove="BootstrapBuildLane;BootstrapBuildArtifactsDir"
+                      UndefineProperties="BootstrapBuildLane;BootstrapBuildArtifactsDir"
```
mmitche, line 9: **5/14: Global-property normalization is essential to producer reuse.** MSBuild caches execution by project path plus global properties. Letting the consumer lane's marker or isolated artifacts root flow into the shared producer reference could create another project instance writing the same product outputs. Both removal metadata forms are intentional: `GlobalPropertiesToRemove` serves the project-reference/restore path, while `UndefineProperties` is honored by the traversal's MSBuild task. The explicit MSBuild call below performs the corresponding `RemoveProperties` operation.

_GitHub Copilot-generated explanatory annotation._

mmitche (commented):
Reviewer walkthrough of the bootstrap redesign. These are explanatory annotations, not change requests. Read the new description sections in order: how bootstrapping worked before, desired state, and how the implementation reaches that state. The inline notes call out the graph boundaries, property identities, artifact ownership, restore timing, compatibility, and regression coverage. No implementation changes are part of this review.

> [!NOTE]
> This explanation and its inline annotations were generated with GitHub Copilot.

src/runtime/eng/bootstrap-layout.targets
```diff
@@ -0,0 +1,93 @@
...
+    </ItemGroup>
+    <Delete Files="$(BootstrapReadyFile)" Condition="'@(MissingBootstrapFile)' != ''" />
+  </Target>
+
+  <Target Name="CreateBootstrapLayout"
+          AfterTargets="Build"
+          DependsOnTargets="_ValidateBootstrapLayout"
+          Inputs="$(BootstrapInputManifest)"
+          Outputs="$(BootstrapReadyFile)">
+    <RemoveDir Directories="$(BootstrapLayoutDir)" />
```
mmitche, line 86: **6/14: This is a scoped snapshot refresh, not the old product cleanup.** Only the bootstrap layout is removed, and only when its input manifest changes or a destination is missing. The manifest above hashes source contents and records the file set/destinations, so preserved timestamps and removed files are handled; `.complete` is written after successful copies. The layout and manifest share a RID scope to handle Release -> Debug -> Release correctly. This is not a content-addressed cache or a hash check of existing destination contents. RID aliases and graph selection above preserve distro/portable consumption.

_GitHub Copilot-generated explanatory annotation._

src/runtime/eng/bootstrap.props
```diff
@@ -0,0 +1,18 @@
...
+    <BootstrapRidGraphDir Condition="'$(BootstrapRidGraphDir)' == ''">$(BootstrapLayoutDir)ridgraph</BootstrapRidGraphDir>
+    <BootstrapPropertiesToRemove>EnableBootstrap;BootstrapMainBuild;BootstrapBuildLane;BootstrapBuildArtifactsDir;CrossgenSmokeTestCoreLibPath;CdacReaderArtifactsPath;CoreCLRCrossILCompilerDir</BootstrapPropertiesToRemove>
+    <_BootstrapLayoutProperties>BootstrapRefPackDir=$(BootstrapRefPackDir);BootstrapRuntimePackDir=$(BootstrapRuntimePackDir);BootstrapAotSdkDir=$(BootstrapAotSdkDir);BootstrapHostDir=$(BootstrapHostDir);BootstrapRidGraphDir=$(BootstrapRidGraphDir)</_BootstrapLayoutProperties>
+    <_BootstrapConsumerProperties>UseBootstrap=true;ArtifactsDir=$(BootstrapBuildArtifactsDir);CoreCLROverridePath=$(_BootstrapProductCoreCLRArtifactsPath);LibrariesOverridePath=$(BootstrapBuildArtifactsDir);LibrariesAllConfigurationsOverridePath=$(BootstrapBuildArtifactsDir);HostArtifactsPath=$(BootstrapConsumerHostArtifactsPath);$(_BootstrapLayoutProperties);CreateHardLinksForCopyFilesToOutputDirectory=false;CreateHardLinksForCopyLocalIfPossible=false;CreateHardLinksForPublishFilesIfPossible=false</_BootstrapConsumerProperties>
```
mmitche, line 15: **7/14: Separate writable outputs from reusable inputs.** `ArtifactsDir` moves consumer `bin`/`obj` into `bootstrap-build/<rid>/<configuration>`, while `CoreCLROverridePath` deliberately points back to the normal product CoreCLR tree for native/JIT inputs, build integration, and the host-runnable ILC. Framework/AOT/host/RID inputs come from the copied bootstrap layout. These overrides are therefore not interchangeable. Disabling relevant hardlink copies avoids sharing writable output files between contexts. Small consumer prerequisites/dependencies may still be built separately; the optimization is reuse of the large shared product producers, not a claim that every project executes once.

_GitHub Copilot-generated explanatory annotation._

src/runtime/eng/bootstrap.proj
```diff
@@ -0,0 +1,49 @@
...
+             Properties="$(_BootstrapConsumerProperties)"
+             RemoveProperties="$(BootstrapPropertiesToRemove)"
+             BuildInParallel="false"
+             StopOnFirstFailure="true" />
+    <MSBuild Projects="@(BootstrapConsumerPrerequisite)"
+             Properties="$(_BootstrapConsumerProperties);ExcludeRestorePackageImports=false"
+             RemoveProperties="$(BootstrapPropertiesToRemove)"
+             BuildInParallel="false"
+             StopOnFirstFailure="true" />
+    <MSBuild Projects="@(BootstrapConsumer)"
```
mmitche, line 36: **8/14: Restore must be late, and publication must use the default targets.** The shared producer/layout boundary has completed before this target. Isolated prerequisite outputs are restored/built first; only then are the actual consumers restored and executed. This avoids evaluating live NativeAOT imports or pack metadata before they exist on a clean machine. The second consumer MSBuild call intentionally omits `Targets`: the ILC/Crossgen2 publish projects default to `Publish;PublishItemsOutputGroup`, so substituting `Targets="Build"` would not produce the shipping executables. Consumer-only globals are also kept out of unrelated producer identities.

_GitHub Copilot-generated explanatory annotation._

src/runtime/Build.proj
```diff
@@ -1,48 +1,85 @@
...
+                      UndefineProperties="BootstrapBuildLane;BootstrapBuildArtifactsDir"
+                      BuildInParallel="true" />
+    <ProjectReference Include="$(RepositoryEngineeringDir)bootstrap.proj"
+                      AdditionalProperties="BootstrapBuildLane=true;BootstrapBuildArtifactsDir=$(BootstrapBuildArtifactsDir)"
+                      BuildInParallel="true"
+                      Pack="false"
+                      Publish="false" />
+  </ItemGroup>
+
+  <Target Name="BuildBootstrapPackages"
```
mmitche: **9/14: Package after the two downstream paths have joined.** The root traversal has waited for both remaining product work and isolated tool publication before this build hook runs. Packages are taken from the subset-selected lists, restored here, and invoked through their default targets. The cDAC case intentionally uses two owners: `mscordaccore_universal` comes from isolated publication, while `mscordbi_universal` stays in the normal native product tree. The new `CdacReaderArtifactsPath` override supports that combination without another debug-component build. Crossgen2's smoke test similarly keeps its input IL CoreLib in the main product tree.

_GitHub Copilot-generated explanatory annotation._

src/runtime/src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Crossgen2.props
```diff
@@ -23,7 +25,10 @@
       Include="$(CrossgenPublishProject)"
       OutputItemType="_RawCrossgenPublishFiles"
       ReferenceOutputAssembly="false"
-      Private="false" />
+      Private="false">
+      <AdditionalProperties Condition="'$(BootstrapBuildArtifactsDir)' != ''">$(_BootstrapConsumerProperties)</AdditionalProperties>
```
mmitche, line 29: **10/14: Packaging must reference the same isolated publish context.** Crossgen2 packaging still follows a project reference to collect publication outputs. When integrated bootstrap is active, this reference receives the centralized consumer properties and drops orchestration-only properties, so it addresses the intended publish context rather than reverting to normal output paths. The condition is important: ordinary/non-integrated packaging must not receive an empty or bootstrap-specific `ArtifactsDir` override. `Private=false` is an existing main-branch fix retained here, not a new bootstrap optimization.

_GitHub Copilot-generated explanatory annotation._

src/runtime/Build.proj
```diff
@@ -1,48 +1,85 @@
...
+             StopOnFirstFailure="true" />
+    <MSBuild Projects="@(BootstrapCdacPackageProject)"
+             Targets="Restore"
+             Properties="CdacReaderArtifactsPath=$(BootstrapConsumerCoreCLRArtifactsPath)" />
+    <MSBuild Projects="@(BootstrapCdacPackageProject)"
+             Properties="CdacReaderArtifactsPath=$(BootstrapConsumerCoreCLRArtifactsPath);ExcludeRestorePackageImports=false" />
+  </Target>
 
-  <Target Name="SetupBootstrapLayout"
+  <Target Name="BuildBootstrapTests"
```
mmitche, line 47: **11/14: Preserve bootstrap-dependent library-test behavior.** The old wrapper passed `UseBootstrap=true` to the whole second invocation, which also affected library tests. Narrowing that setting to shipping-tool consumers alone would lose this behavior, especially where target packs cannot be downloaded. The selected library-test traversal is therefore restored/built after the bootstrap inputs exist, with `UseBootstrap=true`; the adjacent hooks also forward `Test` and `VSTest`. This is a compatibility boundary, not an extra unconditional test run.

_GitHub Copilot-generated explanatory annotation._

src/runtime/Build.proj
```diff
@@ -1,48 +1,85 @@
...
+    <MSBuild Projects="@(BootstrapTestProject)"
+             Targets="VSTest"
+             Properties="UseBootstrap=true;ExcludeRestorePackageImports=false"
+             RemoveProperties="$(BootstrapPropertiesToRemove)"
+             StopOnFirstFailure="true" />
+  </Target>
+
+  <Import Project="$(RepositoryEngineeringDir)SubsetValidation.targets" />
+  <Import Project="$(RepositoryEngineeringDir)restore\optimizationData.targets" Condition="'$(DotNetBuildSourceOnly)' != 'true'" />
+  <Import Project="$(RepositoryEngineeringDir)bootstrap-layout.targets" Condition="'$(Subset)' == 'bootstrap' and '$(_OrchestrateBootstrap)' != 'true'" />
```
mmitche, line 83: **12/14: Keep the explicit two-command interface available.** A user can still build `Subset=bootstrap` and later use `--use-bootstrap`. This import creates the canonical layout on that ordinary traversal without enabling shipping-tool publication. The legacy subset's producer settings remain intact. In CoreCLR mode its IL swap is real, but `SwapNativeForIL=true` alone is not proof of an IL replacement: `externals.csproj` skips the CoreCLR override when `UseNativeAotCoreLib=true`. The integrated path instead snapshots the selected normal product producers.

_GitHub Copilot-generated explanatory annotation._

src/runtime/src/coreclr/runtime.proj
```diff
@@ -122,7 +122,7 @@
     <Message Text="Executing $(_CoreClrBuildCommand)" Importance="High" />
     <Exec Command="$(_CoreClrBuildCommand)"
           IgnoreStandardErrorWarningFormat="true"
-          EnvironmentVariables="EMSDK_PATH=$(EMSDK_PATH)"
+          EnvironmentVariables="EMSDK_PATH=$(EMSDK_PATH);__RootBinDir=$(ArtifactsDir)"
```
mmitche, line 125: **13/14: Keep native script output paths aligned with MSBuild.** Changing `ArtifactsDir` in MSBuild is not sufficient if a called native script reconstructs `<repo>/artifacts` internally. This passes the selected root across the process boundary; the corresponding `.cmd`/`.sh` changes honor it and preserve the default when none is supplied. The corehost and native-libraries changes follow the same pattern and preserve existing environment entries such as WASI SDK paths. This is path-routing infrastructure, not a second native producer build in the integrated graph.

_GitHub Copilot-generated explanatory annotation._

src/runtime/eng/testing/bootstrap-tests.ps1
```diff
@@ -0,0 +1,200 @@
...
+  Assert-True (Test-Path -LiteralPath (Join-Path $layout 'ref\extra.dll')) 'Added inputs must be copied.'
+  Remove-Item -LiteralPath $extraSource
+  Invoke-Layout
+  Assert-True (-not (Test-Path -LiteralPath (Join-Path $layout 'ref\extra.dll'))) 'Removed inputs must not remain in the snapshot.'
+
+  $debugRef = Join-Path $inputRoot 'debug-ref'
+  New-Item -ItemType Directory -Path $debugRef | Out-Null
+  Set-Content -LiteralPath (Join-Path $debugRef 'reference.dll') -Value 'debug'
+  Invoke-Evaluation $fixtureProject @('Configuration=Debug', "MicrosoftNetCoreAppRefPackDir=$debugRef") @('-t:CreateBootstrapLayout', '-getProperty:BootstrapReadyFile,BootstrapLayoutDir') | Out-Null
+  Assert-True ((Get-Content -LiteralPath (Join-Path $layout 'ref\reference.dll') -Raw).Trim() -eq 'debug') 'Switching configurations must update the snapshot.'
```
mmitche, line 175: **14/14: Test the invariants rather than only a successful build.** The script evaluates real project/subset metadata and exercises the actual layout targets with small fixture files. This section catches a subtle stamp bug: because the layout path is shared across configurations, returning to Release after Debug must not accept Debug's layout just because Release's old manifest is unchanged. Other assertions cover file-content changes with preserved timestamps, additions/removals, missing destinations, RID aliases/graphs, deferred restore selection, and ordinary/legacy paths. These checks complement, rather than replace, the clean product and cross-platform CI runs.

_GitHub Copilot-generated explanatory annotation._

mmitche:
I still need to take a close look over this, but I think it's ready for review.

jkoritzinsky:
I don't think the powershell tests are worth keeping around, especially as windows-only tests.

src/runtime/eng/bootstrap-layout.targets
jkoritzinsky, line 7: Can we do this without a stamp/complete file? We are trying to remove the various cases where we have this pattern in the runtime repo today.

src/runtime/eng/bootstrap.proj
```diff
@@ -0,0 +1,49 @@
...
+
+  <Target Name="_BuildBootstrapProducersBeforeBootstrap" BeforeTargets="Build">
+    <MSBuild Projects="$(RepositoryEngineeringDir)bootstrap-producers.proj"
+             Properties="EnableBootstrap=false;BootstrapMainBuild=true"
+             RemoveProperties="BootstrapBuildLane;BootstrapBuildArtifactsDir"
+             BuildInParallel="false"
+             StopOnFirstFailure="true" />
+  </Target>
+
+  <Target Name="BuildBootstrapConsumers" AfterTargets="Build" Condition="'@(BootstrapConsumer)' != ''">
```
jkoritzinsky, line 21: Instead of building the consumers here, can we rely on the fact that the runtime build currently sets `BuildInParallel` to `false` for the projects that make up each subset?

src/runtime/eng/main-build.proj
```diff
@@ -0,0 +1,19 @@
...
+                      BuildInParallel="false" />
+  </ItemGroup>
+
+  <Target Name="_BuildBootstrapProducersBeforeMainBuild"
+          BeforeTargets="Build">
+    <MSBuild Projects="$(RepositoryEngineeringDir)bootstrap-producers.proj"
+             Properties="EnableBootstrap=false;BootstrapMainBuild=true"
+             BuildInParallel="false"
+             StopOnFirstFailure="true" />
+  </Target>
```
jkoritzinsky, line 17: Why are we building `bootstrap-producers.proj` here twice?

src/runtime/eng/Subsets.props
```diff
@@ -824,4 +856,15 @@
...
+    <BootstrapDependentPackageProject Include="@(ProjectToBuild->WithMetadataValue('BootstrapDependentPackage', 'true'))" />
+    <BootstrapCdacPackageProject Include="@(ProjectToBuild->WithMetadataValue('BootstrapCdacPackage', 'true'))" />
+    <BootstrapTestProject Include="@(ProjectToBuild->WithMetadataValue('BootstrapTest', 'true'))" />
+  </ItemGroup>
```
jkoritzinsky, line 864: All of these different `Bootstrap*` metadata properties add a lot of maintenance burden. Why do we need all these different ones? What benefit to we get from having multiple different metadata properties (and item groups) here?

src/runtime/src/installer/pkg/projects/Microsoft.DotNet.Cdac.Transport/Microsoft.DotNet.Cdac.Transport.pkgproj
jkoritzinsky, line 1: We shouldn't need any changes in this file as we shouldn't build the cdacreader in the bootstrap phase (so nothing should move).

src/runtime/src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Crossgen2.sfxproj
jkoritzinsky, line 1: Same question in this file, why do we need changes here?

jkoritzinsky (commented):
This seems to add a lot of complexity to the build. Is there a way we can make this less intrusive?

One of the benefits of the existing design is that it really doesn't impact the performance of the non-bootstrapped build at all and doesn't add a ton of maintenance burden.

PR #8656, opened 2026-08-25
Baseline ARM64 DirectWriteForwarder BA2007

wtgodbe:
Adds the ARM64-only shipping `DirectWriteForwarder.dll` BA2007 record from [internal unified build 3056983](https://dev.azure.com/dnceng/internal/_build/results?buildId=3056983&view=results).

The x86/x64 records are left unchanged. Cross-checking the build's Windows x86, x64, and ARM64 hydrated baselines showed this was the only ARM64-only record.

### DIFF
--- a/.config/guardian/.gdnbaselines
+++ b/.config/guardian/.gdnbaselines
@@ -8,7 +8,7 @@
     "default": {
       "name": "default",
       "createdDate": "2026-08-12 03:04:30Z",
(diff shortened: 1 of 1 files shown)


akoeplinger (approved):

akoeplinger:
Not tested in public, don't need to waste CI resources.

PR #8657, opened 2026-08-25
[release/11.0.1xx] Baseline ARM64 DirectWriteForwarder BA2007

wtgodbe:
Adds the ARM64-only shipping `DirectWriteForwarder.dll` BA2007 record from [internal unified build 3053551](https://dev.azure.com/dnceng/internal/_build/results?buildId=3053551&view=results).

The latest build did not emit this record because incremental BinSkim did not select the shipping copy. Build 3053551 on this branch did select it and failed with the exact RC2 signature used here. Existing x86/x64 records remain unchanged.


wtgodbe:
Closing after cross-checking the Windows x86, x64, and ARM64 hydrated baselines from build 3056675. Every ARM64 record already shares a primary signature with an existing baseline entry, so this branch has no ARM64-only record to add.

PR #8658, opened 2026-08-25
Install microsoft.esrpclient into Agent.TempDirectory

akoeplinger:
Avoids it being picked up by guardian tools scanning artifact outputs in Build.StagingDirectory


ViktorHofer (approved):

akoeplinger:
we only need it in 1xx since DAC signing is only done there

PR #8661, opened 2026-08-25
Add auto-generated BinSkim remediations

wtgodbe:
Test out auto-generated BinSkim remediations for aspnetcore, diagnostics, and SDK


fernandoesteban2408-ux (approved):

jkotas:
FWIW, we have on and off discussion about the protection that this actually provides for .NET and whether it makes sense to .NET binaries.

Does this change actually make the binshim happy on the binaries that it is trying to fix?

jkotas:
Also, the requirement to specify SpectreMitigation is being removed, so we should not need it going forward. It comes with performance overhead.

wtgodbe:
> Does this change actually make the binshim happy on the binaries that it is trying to fix?

This was just testing whether the auto-generated remediations work, which they seem to.

jkotas:
> which they seem to.

I doubt that this change would be a full fix for the issue that binskim is complaining about.

wtgodbe:
> I doubt that this change would be a full fix for the issue that binskim is complaining about.

Why do you say that?

jkotas:
It would not address the issue for anything that is NativeAOT compiled.

PR #8704, opened 2026-08-26
Improve scenario test process diagnostics

dsplaisted:
Source-only validation failed here: https://dev.azure.com/dnceng/internal/_build/results?buildId=3057640&view=results

The apparent cause was a restore hang when running `dotnet new webapi` as part of the scenario tests.  These changes add additional logging and split the restore out as a separate operation from creating the template.  This should hopefully help pinpoint the failure better if in recurs.


mthalman (approved):

akoeplinger:
@dsplaisted looks like there are conflicts in newer branches, please port manually (VMR doesn't do inter-branch merging)

PR #8745, opened 2026-08-27
Use complog for source indexer stage1

jjonescz:
Validation: https://dev.azure.com/dnceng/internal/_build/results?buildId=3058906&view=results

### DIFF
--- a/eng/pipelines/templates/steps/source-index-stage1.yml
+++ b/eng/pipelines/templates/steps/source-index-stage1.yml
@@ -1,4 +1,4 @@
-### Steps to process multiple per-repo Build.binlog files into SLN format and upload
+### Steps to process multiple per-repo Build.binlog files into CompLog format and upload
 ### as source index stage 1 data for https://github.com/dotnet/source-indexer.
 ###
 ### Unlike eng/common/core-templates/steps/source-index-stage1-publish.yml which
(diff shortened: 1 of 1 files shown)


jaredpar (approved):

tannergooding:
What's the set of complogs we're collecting with this?

It mostly just makes a difference in what we can do index merging around, but I think the key distinct builds (at least for dotnet/runtime) we may want to consider are:
* Windows/Linux
* Debug/Release
* NAOT/JIT
* x64/Arm64

Right now I believe we're collecting `Release NAOT Linux` and so the Windows and JIT specific sources are missing. I think most debug vs release is just a matter of hyperlinks/ifdefs (not files) and same for x64 vs arm64.

We don't have to go and collect them all, can do a limited subset, or even grab the others in a later PR; but having some of them would let us start using the merged index feature.

jjonescz:
> What's the set of complogs we're collecting with this?

Currently it's the exact same as we did with BinLogToSln.

I think we can try adding more in a follow up. First I'd like to try that everything works end-to-end. Already discovered two issues so far (https://github.com/dotnet/source-indexer/pull/304, https://github.com/dotnet/source-indexer/pull/305).

tannergooding (approved):

PR #8759, opened 2026-08-27
Fix component authoring

joeloff:
The registry key component for writing specific version information should not be fixed. This makes it a shared component that isn't always removed, so the registry contains invalid information.

Example, when you install 10.0.9 and 10.0.10, the 10.0.10 entry remains. This can cause problems for ISVs that rely on the keys to detect specific version, potentially skipping over installs. It can also trigger scanners to report older versions that aren't actually installed.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix5/product/registrykeys.wxs
+++ b/src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix5/product/registrykeys.wxs
@@ -4,7 +4,7 @@
 
   <Fragment>
     <ComponentGroup Id="AuthoredRegistryKeys">
-      <Component Id="SetupRegistry_x86" Directory="TARGETDIR" Bitness="always32" Guid="f1dc3d97-93c5-4f3b-baa3-24e6a84a2e15" >
+      <Component Id="SetupRegistry_x86" Directory="TARGETDIR" Bitness="always32" >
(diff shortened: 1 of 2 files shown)


src/arcade/src/Microsoft.DotNet.Build.Tasks.Installers/build/wix5/product/registrykeys.wxs
```diff
@@ -4,7 +4,7 @@
 
   <Fragment>
     <ComponentGroup Id="AuthoredRegistryKeys">
-      <Component Id="SetupRegistry_x86" Directory="TARGETDIR" Bitness="always32" Guid="f1dc3d97-93c5-4f3b-baa3-24e6a84a2e15" >
+      <Component Id="SetupRegistry_x86" Directory="TARGETDIR" Bitness="always32" >
```
PranavSenthilnathan, line 7: AI is saying that automatic GUID generation will also generate the same GUID every time because it uses the registry name as the seed and doesn't consider the value. So, it looks like this change will work once (when the GUID changes from the existing fixed GUID to the new auto generated one) but won't work after. Is that correct and/or intentional?

joeloff: The keypath is off the registry value, so it generates new values each time because the version changes. It generates stable GUIDs as long as you have stable keypaths, that's why it's useful for shared components, but the keypath will change each time we bump the version. So two consecutive builds of say 10.0.14 will generate the same GUID, but 10.0.14 and 10.0.15 will be different.

Here's an example. 2 components, same registry key, different values

image

image

PranavSenthilnathan (approved):

PR #8858, opened 2026-08-31
Fix MTP handshakes in source-build test pipelines

mthalman:
## Problem

**Error or behavior:** Source-build SDK diff and license-scan tests exited with code 5 before running tests following the MTP migration in https://github.com/dotnet/dotnet/pull/8060.

**Root cause:** `dotnet test` forwarded the MSBuild `-flp` argument to the MTP xUnit application, which rejected it as an unknown option.

## Description

**Fix:** Remove `-flp` from both pipeline invocations while retaining binlog and console logging. Local validation confirmed successful MTP handshakes and filtered test discovery.

### DIFF
--- a/eng/pipelines/templates/jobs/sdk-diff-tests.yml
+++ b/eng/pipelines/templates/jobs/sdk-diff-tests.yml
@@ -147,7 +147,6 @@ jobs:
       --filter-trait "Category=SdkContent"
       -c Release
       -bl:$(Build.SourcesDirectory)/artifacts/log/Debug/BuildTests_$(date +"%m%d%H%M%S").binlog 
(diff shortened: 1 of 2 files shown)


ViktorHofer (approved):
Thanks for fixing. Isn't it weird that `dotnet test` supports -clp but not -flp?

akoeplinger:
AzDO build passed but didn't report to GitHub, merging

akoeplinger:
>  Isn't it weird that dotnet test supports -clp but not -flp?

sounds weird to me too. I assume there's a filter of flags which are passed to msbuild somewhere?

akoeplinger:
Presumably https://github.com/dotnet/sdk/blob/37e1660e335d4536fde6e1b9311806394048ef11/src/Cli/dotnet/LoggerUtility.cs#L144-L151

PR #8312, opened 2026-08-14
[main] Initialize .NET 12 branding

mmitche:
Initializes .NET 12 branding to alpha.1. This PR and the major-release Maestro PR may merge in either order; branding first is preferred.

PR #8902, opened 2026-09-01
Fix build failures when building non-portable TargetRid that is different from the SDK's non-portable rid.

tmds:
The compilation fails when when usings a non-portable SDK (e.g. rhel.10-x64) to build a different non-portable RID (e.g. centos.10-x64).

The root cause is that ILCompiler's SingleEntry.targets is not able to determine the host package and target OS for the unknown that is being built RID.

This change passes that information from arcade to SingleEntry.targets.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets
@@ -99,28 +99,32 @@
   </ItemGroup>
 
   <!--
-    When .NET gets built from source, make the SDK aware there are bootstrap packages
-    for Microsoft.NETCore.App.Runtime.<rid> (including the NativeAOT-labeled runtime pack),
-    Microsoft.NETCore.App.Crossgen2.<rid>, and the ILCompiler packs
-    (runtime.<rid>.Microsoft.DotNet.ILCompiler / Microsoft.NETCore.App.Runtime.NativeAOT.<rid>).
-    The on-disk previous SDK does not list the build's non-portable RID in these RID sets, so
-    append it in-memory here. This allows publishing for the non-portable RID (e.g. self-contained,
-    ReadyToRun, and AOT) to validate and restore the matching source-built packs.
+    When .NET gets built from source, the SDK's known RID sets may not include the RIDs used during the build.
+    Append them here so that self-contained publishing (ReadyToRun, NativeAOT) can locate the matching source-built packs.
 
-    Use $(TargetRid) (the build's non-portable RID, e.g. centos.10-x64), not
-    $(NETCoreSdkRuntimeIdentifier). In a stage 1 source-build leg the previous SDK is the Microsoft
-    portable SDK, whose $(NETCoreSdkRuntimeIdentifier) is the portable RID (e.g. linux-x64) and would
-    not unblock publishing for the build's non-portable RID.
+    $(NETCoreSdkRuntimeIdentifier) is added to KnownRuntimePack and KnownCrossgen2Pack because the runtime *_inbuild tool projects (crossgen2_inbuild, ILCompiler_inbuild)
+    are published self-contained for the host using RuntimeIdentifier=$(NETCoreSdkRuntimeIdentifier).
+
+    Both $(TargetRid) and $(NETCoreSdkRuntimeIdentifier) are added to KnownILCompilerPack because source-build produces two ILCompiler NuGet packages:
+    one for the target RID ($(TargetRid)), and one for the host SDK RID ($(NETCoreSdkRuntimeIdentifier)).
+    Both need to be known for pack resolution to succeed.
+
+    IlcHostRuntimeIdentifier/IlcTargetOS are set to $(NETCoreSdkRuntimeIdentifier)/$(TargetOS).
+    This is needed because the logic in SingleEntry.targets does not work when using a non-portable SDK to build an SDK with a different non-portable RID.
   -->
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
+    <IlcHostRuntimeIdentifier Condition="'$(IlcHostRuntimeIdentifier)' == ''">$(NETCoreSdkRuntimeIdentifier)</IlcHostRuntimeIdentifier>
+    <IlcTargetOS Condition="'$(IlcTargetOS)' == ''">$(TargetOS)</IlcTargetOS>
+  </PropertyGroup>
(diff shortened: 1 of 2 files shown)


tmds:
> The compilation fails when when usings a non-portable SDK (e.g. rhel.10-x64) to build a different non-portable RID (e.g. centos.10-x64).

This regressed in https://github.com/dotnet/dotnet/pull/7300.

@mthalman @jkotas @akoeplinger ptal.

The changes include comments that describe why what rid gets used where.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets
```diff
@@ -99,28 +99,32 @@
...
+    are published self-contained for the host using RuntimeIdentifier=$(NETCoreSdkRuntimeIdentifier).
+
+    Both $(TargetRid) and $(NETCoreSdkRuntimeIdentifier) are added to KnownILCompilerPack because source-build produces two ILCompiler NuGet packages:
+    one for the target RID ($(TargetRid)), and one for the host SDK RID ($(NETCoreSdkRuntimeIdentifier)).
+    Both need to be known for pack resolution to succeed.
+
+    IlcHostRuntimeIdentifier/IlcTargetOS are set to $(NETCoreSdkRuntimeIdentifier)/$(TargetOS).
+    This is needed because the logic in SingleEntry.targets does not work when using a non-portable SDK to build an SDK with a different non-portable RID.
   -->
+  <PropertyGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
```
am11, line 115: Does it hold for DotNetBuildSourceOnly!=false as well?

tmds: Yes. I matched the Condition to the one that was already on the related/following ItemGroup.

jkoritzinsky (dismissed):

src/runtime/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
```diff
@@ -1,6 +1,16 @@
...
 
   <PropertyGroup Condition="'$(RuntimeIdentifier)' != '' and '$(NETCoreSdkPortableRuntimeIdentifier)' != ''">
+    <!--
+      Based on RuntimeIdentifier we decide on the host package and target OS.
+      - Normally, the RuntimeIdentifier is a portable rid.
+        Then we use the host package for NETCoreSdkPortableRuntimeIdentifier, and determine the target OS based on the NETCoreSdkPortableRuntimeIdentifier.
+      - A special case is when the SDK is non-portable and RuntimeIdentifier matches its non-portable NETCoreSdkRuntimeIdentifier.
+        Then we use the host package for that non-portable rid, and determine the target OS based on NETCoreSdkPortableRuntimeIdentifier.
+      - When .NET itself gets built, the RuntimeIdentifier may be a non-portable rid that does not match the SDKs non-portable rid.
+        In that case, the IlcHostRuntimeIdentifier and IlcTargetOS may be used to override the above logic.
```
akoeplinger, line 11: should we document the new override properties somewhere?

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets
akoeplinger: FYI this will conflict with the changes in https://github.com/dotnet/dotnet/pull/8312/changes#diff-16689dbb954f2efc0cb792ae9d79568a49695cf8ed7476507fd5e4bec2c5375eR115

am11:
> /azp run dotnet-unified-build-full

It would be useful to have at least one community platform coverage in this pipeline. e.g. freebsd-arm64 covers both "different OS" and "different arch" w.r.t host to avoid regressions from such changes in VMR (saves cycles).

am11:
I've triggered it here https://github.com/filipnavara/dotnet-riscv/actions/runs/33807980599, need to check two things: 1) it is green 2) download artifacts and check (with `file`) if crossgen2, ilc and other AOT published binaries are valid riscv64 ELF.

BTW, preview7 slipped through some "only affects community platform" because "official platform once again got away with LKG package fallback.." and failed the preview release https://github.com/filipnavara/dotnet-riscv/actions/runs/31601669575. (not the first time this pain is shared: https://github.com/dotnet/dotnet/issues/7677)

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets
tmds: OK. I'd like this to be backported to .NET 11, so may be we want to merge this before https://github.com/dotnet/dotnet/pull/8312 so we only have to deal with the merge conflicts once? Otherwise, we'll deal with the merge conflicts here and then again on the backport, which is also fine.

src/runtime/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
tmds: I tried to pick proper names and add sufficient documentation in the XML comments. I think that may be sufficient.

am11:
> 1) it is green 2) download artifacts and check (with `file`) if crossgen2, ilc and other AOT published binaries are valid riscv64 ELF.

Both checked, LGTM! 👍

akoeplinger:
@tmds we needed to get the net12 PR in first, please fix the conflicts here, thanks :)

tmds:
The .NET 12 PR (https://github.com/dotnet/dotnet/pull/8312) moved additional logic from `src/sdk/Directory.Build.targets` into `src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets`.

To minimize changes for .NET 11, I've updated this PR to target the .NET 11 branch (release/11.0.1xx) with the exact changes that were already reviewed, approved and validated.

I will make a separate PR for `main` based on the changes made in https://github.com/dotnet/dotnet/pull/8312.

@akoeplinger can this be merged into the .NET 11 branch? Is `release/11.0.1xx` the proper target?

akoeplinger (approved):

PR #9176, opened 2026-09-09
Fix build failures when building non-portable TargetRid that is different from the SDK's non-portable rid.

tmds:
This is https://github.com/dotnet/dotnet/pull/8902 applied to main, taking into account some changes made in https://github.com/dotnet/dotnet/pull/8312.

@jkoritzinsky @akoeplinger @am11 ptal.

### DIFF
--- a/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets
+++ b/src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets
@@ -99,38 +99,40 @@
   </ItemGroup>
 
   <!--
-    When .NET gets built from source, make the SDK aware there are bootstrap packages
-    for Microsoft.NETCore.App.Runtime.<rid> (including the NativeAOT-labeled runtime pack),
-    Microsoft.NETCore.App.Crossgen2.<rid>, and the ILCompiler packs
-    (runtime.<rid>.Microsoft.DotNet.ILCompiler / Microsoft.NETCore.App.Runtime.NativeAOT.<rid>).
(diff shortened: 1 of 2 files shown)


src/runtime/src/coreclr/nativeaot/BuildIntegration/Microsoft.DotNet.ILCompiler.SingleEntry.targets
```diff
@@ -1,6 +1,16 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup Condition="'$(RuntimeIdentifier)' != '' and '$(NETCoreSdkPortableRuntimeIdentifier)' != ''">
+    <!--
```
tmds, line 4: These changes make this file identical to that on the .NET 11 branch.

src/arcade/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetingPacks.BeforeCommonTargets.targets
```diff
@@ -99,38 +99,40 @@
   </ItemGroup>
 
   <!--
```
tmds, line 101: I've changed this to unconditionally add both rids and explain why they might not be there yet. I think this is the simplest way to go about it.

I don't think we should bother with filtering out duplicates or empty ones.

am11:
Kicked linux-riscv64 build against the PR branch: https://github.com/filipnavara/dotnet-riscv/actions/runs/34333712887.

Update: all good 👍

jkoritzinsky (approved):

akoeplinger (approved):

PR #9184, opened 2026-09-09
Fix TaskAnalyzer dependencies for source-build for msbuild

AR-May:
## Summary

- include `Microsoft.Build.TaskAuthoring.Analyzer` in MSBuild's source-build solution filter
- track its Roslyn dependencies as previously source-built inputs
- flow N-1 Roslyn versions into MSBuild through analyzer-specific package properties
- preserve the released Roslyn pins for non-source-build configurations


VolPlita (approved):

src/msbuild/eng/Version.Details.xml
```diff
@@ -104,6 +104,20 @@
...
+         Update these manual dependencies with the N-1 SDK. -->
+    <Dependency Name="Microsoft.CodeAnalysis.Analyzers" Version="5.11.0-1.26420.103" Label="Manual">
+      <Uri>https://github.com/dotnet/dotnet</Uri>
+      <Sha>1d599674e31ad86ca9e6e1dad7055b531e4d58e4</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.CodeAnalysis.CSharp" Version="5.11.0-1.26420.103" Label="Manual">
+      <Uri>https://github.com/dotnet/dotnet</Uri>
+      <Sha>1d599674e31ad86ca9e6e1dad7055b531e4d58e4</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="5.11.0-1.26420.103" Label="Manual">
```
ViktorHofer: `5.11.0-1.26420.103`  is quite a recent build which means that our analyzer assembly can only be loaded with a compiler >= 5.11. Is that necessary?

src/msbuild/eng/Version.Details.xml
```diff
@@ -104,6 +104,20 @@
       <Sha>
       </Sha>
     </Dependency>
+    <!-- Track the N-1 Roslyn packages used to build TaskAnalyzer in source-only builds.
+         Update these manual dependencies with the N-1 SDK. -->
+    <Dependency Name="Microsoft.CodeAnalysis.Analyzers" Version="5.11.0-1.26420.103" Label="Manual">
```
ViktorHofer: I would remove the `Label="Manual"`. We don't read those labels and Maestro doesn't consider them. AspNetCore used them for readability but we don't need them here.

ViktorHofer (commented):
Looks good aside from the one version question about the minimum required compiler.

src/msbuild/eng/Version.Details.xml
VolPlita: Updated to source-built Roslyn 4.14 (`4.14.0-3.25210.103`). TaskAnalyzer now disables central transitive pinning for this override, so the Roslyn closure resolves `System.Collections.Immutable` and `System.Reflection.Metadata` 9.0 instead of 10.0. I verified the produced analyzer loads, reports diagnostics, and runs its suppressor under the .NET 9 Roslyn 4.14 compiler with no CS8032 or CS9057 warnings.

VolPlita (approved):

repo-projects/msbuild.proj
```diff
@@ -22,10 +22,9 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(DotNetBuildSourceOnly)' == 'true'">
-    <!-- TaskAnalyzer's released Roslyn packages are prebuilts. Use the N-1 SDK Roslyn because
-         live Roslyn is newer than the executing compiler and causes CS9057. -->
-    <ExtraPackageVersionPropsPackageInfo Include="Analyzer_MicrosoftCodeAnalysisCSharpVersion" Version="%24(MicrosoftCodeAnalysisCSharpPreviousVersion)" />
-    <ExtraPackageVersionPropsPackageInfo Include="Analyzer_MicrosoftCodeAnalysisCSharpWorkspacesVersion" Version="%24(MicrosoftCodeAnalysisCSharpWorkspacesPreviousVersion)" />
+    <!-- Use source-built Roslyn 4.14 so TaskAnalyzer remains loadable by older compilers. -->
+    <ExtraPackageVersionPropsPackageInfo Include="Analyzer_MicrosoftCodeAnalysisCSharpVersion" Version="4.14.0-3.25210.103" />
+    <ExtraPackageVersionPropsPackageInfo Include="Analyzer_MicrosoftCodeAnalysisCSharpWorkspacesVersion" Version="4.14.0-3.25210.103" />
```
ViktorHofer: Don't hardcode the versions here. These are already available inside msbuild via the eng/Version.Details.props file. The VMR shouldn't have any hardcoded versions per repo.

src/msbuild/eng/dependabot/Directory.Packages.props
```diff
@@ -73,7 +73,7 @@
   </ItemGroup>
 
   <!-- TaskAnalyzer uses released Roslyn packages by default.
-       Source-only builds use the N-1 SDK Roslyn to avoid CS9057. -->
+       Source-only builds use source-built Roslyn 4.14 for compiler compatibility. -->
```
ViktorHofer: I'm confused. Why only source-build? The shipped packages are produced from the non-source-build jobs. A consumer would still not be able to load the package with an older compiler version then.

src/msbuild/src/TaskAnalyzer/TaskAnalyzer.csproj
```diff
@@ -11,6 +11,7 @@
     <!-- Don't ship yet, just make it available to partner repositories. -->
     <IsShipping>false</IsShipping>
     <IsTestProject>false</IsTestProject>
+    <CentralPackageTransitivePinningEnabled Condition="'$(Analyzer_MicrosoftCodeAnalysisCSharpVersion)' != ''">false</CentralPackageTransitivePinningEnabled>
```
ViktorHofer: Why was this necessary?

src/msbuild/eng/dependabot/Directory.Packages.props
VolPlita: The Roslyn override is source-build-only because normal shipping builds already use the older released Roslyn 4.12 packages. But, central transitive pinning upgrades Roslyn’s supporting dependencies to the repository-wide versions, which can make the resulting analyzer incompatible with older compiler hosts. We therefore need to disable central transitive pinning for TaskAnalyzer in all build modes, allowing Roslyn 4.12 and the source-built Roslyn 4.14 to use their respective compatible dependency closures.

repo-projects/msbuild.proj
VolPlita: Fixed. The VMR mapping now uses the MicrosoftCodeAnalysisCSharpVersion and MicrosoftCodeAnalysisCSharpWorkspacesVersion properties from MSBuild eng/Version.Details.props instead of hard-coded versions.

src/msbuild/eng/dependabot/Directory.Packages.props
```diff
@@ -72,11 +72,17 @@
     <PackageVersion Update="PolySharp" Condition="'$(PolySharpVersion)' != ''" Version="$(PolySharpVersion)" />
   </ItemGroup>
 
-  <!-- Roslyn analyzer authoring packages (used by TaskAnalyzer) -->
+  <!-- TaskAnalyzer uses released Roslyn packages by default.
```
ViktorHofer: Move this whole block out of the dependabot Directory.Packages.props file. We dont want dependabot to update these dependencies as they enforce a minimum supported compiler version. We only want them to get updated manually.

src/msbuild/src/TaskAnalyzer/TaskAnalyzer.csproj
VolPlita: Central transitive pinning upgraded System.Collections.Immutable and System.Reflection.Metadata to 10.0, which caused every analyzer instance to fail under the Roslyn 4.14 host with CS8032. Disabling it for TaskAnalyzer resolves Roslyn 4.14 compatible 9.0 dependencies. I verified analyzer diagnostics and the suppressor under the .NET 9 Roslyn 4.14 compiler with no CS8032 or CS9057 warnings.

src/msbuild/eng/dependabot/Directory.Packages.props
VolPlita: Moved the complete TaskAnalyzer package-version block to the manually maintained src/msbuild/Directory.Packages.props file. Dependabot will no longer update these compiler-baseline dependencies.

src/msbuild/eng/dependabot/Directory.Packages.props
VolPlita: Aligned both modes on Roslyn 4.14: normal builds now use released 4.14.0, while source-build uses the matching source-built package version. The TaskAnalyzer transitive-pinning exception is also now unconditional, so shipped packages retain the compatible 9.0 dependency closure.

src/msbuild/eng/Version.Details.xml
VolPlita: Thanks for clarifying. I had incorrectly treated an older source-built 4.14 package as an input we should pin in the current VMR graph. I restored the intended flow: normal MSFT builds use the manually pinned released 4.14.0 packages, while source-build maps the analyzer-specific properties to the automatically generated *PreviousVersion PSB properties (currently 5.x). The historical 4.14 source-built dependency entries were removed.

VolPlita:
Updated the dependency metadata to match the normal build: Roslyn packages are 4.14.0 from dotnet/roslyn commit 8edf7bcd4f1594c3d68a6a567469f41dbd33dd1b, and Microsoft.CodeAnalysis.Analyzers is 3.11.0 from dotnet/roslyn-analyzers commit f2384e61845314d0e8ad74cc0fe67d1bd4e9f102. Source-build still gets the PSB versions through the VMR PreviousVersion properties.

src/msbuild/Directory.Packages.props
```diff
@@ -18,6 +18,15 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.Collections" Version="$(MicrosoftCodeAnalysisCollectionsVersion)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Contracts" Version="$(MicrosoftCodeAnalysisContractsVersion)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.PooledObjects" Version="$(MicrosoftCodeAnalysisPooledObjectsVersion)" />
+
+    <!-- TaskAnalyzer dependencies define its minimum supported compiler version. Update manually. -->
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.14.0" />
+    <PackageVersion Update="Microsoft.CodeAnalysis.CSharp" Condition="'$(Analyzer_MicrosoftCodeAnalysisCSharpVersion)' != ''" Version="$(Analyzer_MicrosoftCodeAnalysisCSharpVersion)" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
```
ViktorHofer: You are introducing the new properties in `Version.Details.props`, i.e. `MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion` but they aren't used anywhere. Instead of hardcoding the version yet again here, use the property instead.

Same for the other two.

ViktorHofer (approved):

PR #9229, opened 2026-09-10
Add MSBUILDNODEHANDSHAKESALT for node reuse

ViktorHofer:
Add MSBUILDNODEHANDSHAKESALT to EnvironmentVariables.


JanProvaznik (approved):
note that if we want to have node reuse in the build we should have the reused nodes timeout fast or explicitly kill them. otherwise they'll accumulate and slow down the overall build I'd expect

ViktorHofer:
What do you mean by "them"? The default idle time for server / reusable worker nodes is 15min, right? Should we pass an additional env var in to control that?

JanProvaznik:
reused nodes, yes

JanProvaznik (approved):

